PR #3, opened 2020-11-25
Allow only one of `build` and `run` scripts to be present.

RagnarGrootKoerkamp:
As discussed.

Also clarify the `current working directory` situation, hopefully accurately.

@niemela @eldering


niemela:
This is still not quite right. The current text seems to imply that if either `build` and `run` exists then `run` must exist.

The intended change is good, but I'll close this PR.

PR #4, opened 2020-11-25
Add @EXPECTED_RESULTS@ tag.

RagnarGrootKoerkamp:
As discussed.

Suggestions on wording are welcome. I think the directory name should be one of:
- `mixed_results`
- `mixed_verdicts`
- `multiple_results` (although only one would still be OK)
- `multiple_verdicts`

### DIFF
--- a/spec/problem_format.md
+++ b/spec/problem_format.md
@@ -401,7 +401,7 @@ subdirectories of `submissions/`. The possible subdirectories are:
 | `wrong_answer`         | Wrong answer for some test file, but is not too slow and does not crash for any test file                                          |                                          |
 | `time_limit_exceeded` | Too slow for some test file. May also give wrong answer but not crash for any test file.                                           |                                          |
 | `run_time_error`      | Crashes for some test file                                                                                                         |                                          |
-| `mixed_result` | The submission contains `@EXPECTED_RESULTS@:` with expected verdicts, as explained below. | Use when the result is not deterministic. |
+| `mixed_result` | The submission contains `@EXPECTED_VERDICTS@:` with expected verdicts, as explained below. | Use when the result is not deterministic. |
 
 ```todo
 `mixed` vs `multiple` and `result` vs `verdict`.
@@ -427,30 +427,27 @@ filenames. It is mandatory to provide at least one accepted solution.
 Submissions must read input data from standard input, and write output
 to standard output.
 
-### EXPECTED_RESULTS
+### EXPECTED_VERDICTS
 
 Submissions in the `submissions/mixed_result` directory must contain an
-`@EXPECTED_RESULTS@: ` tag (including the final space) exactly once in one
+`@EXPECTED_VERDICTS@: ` tag (including the final space) exactly once in one
 of their source files. This must be followed (on the same line) by a non-empty
 comma separated subset of:
 
-- `ACCEPTED`
-- `WRONG_ANSWER`
-- `RUN_TIME_ERROR` 
-- `TIME_LIMIT_EXCEEDED` 
-- <s class="problemarchive"><s>`PARTIALLY_ACCEPTED` (for scoring problems only).</s></s>
+- `AC`
+- `WA`
+- `TLE`
+- `RTE`
 
 Additional white space is ignored and the tags are not case sensitive.
 
 This tag implies that the submission may get any of the listed verdicts as
-final verdict. Single testcases may still result in a verdict not in the list.
-In particular `ACCEPTED` does not have to be listed unless the submission is
-allowed to pass *all* testcases.
+final verdict.
 
 If `@EXPECTED_RESULTS@: ` is found in a submission in any of the other
 directories, a warning is raised when the directory verdict is not listed. Note
-that this does not have other other effect, and in particular only the
-directory verdict is an allowed.
+that this does not have any other effect, and in particular only the
+verdict matching the directory is allowed.
 
 
 ## Input Validators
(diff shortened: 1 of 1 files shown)


simonlindholm:
Could be worth coming up with a syntax that naturally extends to multiple test groups. We have an unofficial tool https://github.com/nordicolympiad/testdata_tools/pull/13/files that parses `@EXPECTED_GRADES@ AC AC TLE TLE` as "expect groups 1 and 2 to pass, 3 and 4 to TLE"; the similar syntax with different semantics is a bit confusing. (Though it probably won't lead to errors in practice since they are used in different contexts.)

`@ALLOWED_VERDICTS@` is another possible bikeshed color, not sure if a good one.

eldering:
> `@ALLOWED_VERDICTS@` is another possible bikeshed color, not sure if a good one.

Unless there's a good reason to change, I'd prefer to stick to this name as it is already in use in DOMjudge and also by some problem setters.

RagnarGrootKoerkamp:
We should use short hand AC/WA/RTE/TLE

RagnarGrootKoerkamp:
drop mixed result

spec/problem_format.md
```diff
@@ -423,6 +427,29 @@ filenames. It is mandatory to provide at least one accepted solution.
...
+- `WA`
+- `TLE`
+- `RTE`
+
+Additional white space is ignored and the tags are not case sensitive.
+
+This tag implies that the submission may get any of the listed verdicts as
+final verdict.
+
+If `@EXPECTED_RESULTS@: ` is found in a submission in any of the other
```
eldering, line 447: This doesn't match the `@EXPECTED_VERDICTS@` above.

jsannemo:
@RagnarGrootKoerkamp I think we agreed on calling the mixed-bag-directory `rejected`, but I would still like to rescue the EXPECTED_RESULTS for those submissions!

Mind bumping this PR (and fixing @eldering 's comment) and perhaps @niemela can take a look?

RagnarGrootKoerkamp:
So the semantics would be:

- `@EXPECTED_RESULTS@` is only allowed in `rejected/`. (Or anywhere, for backwards compatibility?)
- `rejected` submissions optionally include `@EXPECTED_RESULTS@`. If not, any non-AC verdict is ok.

What about submissions that are WA or AC based on randomness? Should they be in accepted or rejected?

niemela:
> What about submissions that are WA or AC based on randomness?

When are they useful?

> Should they be in accepted or rejected?

Feels like that should be `is_submission` or `ignore` or something?

jsannemo:
Replying to your original question, I agree that those are the semantics we want. I'd say to only allow it in `rejected/`? Problems allowing multiple results that are in one of the other subdirectories (except TLE that also allows WA) already break the spec, so it's not really breaking backwards compatibility to break them.

RagnarGrootKoerkamp:
Regarding having `ACCEPTED` as part of the `EXPECTED_RESULTS`: I have 23 submissions in the last 3 years of BAPC that do this, spread over 12 problems.

Reasons for doing this:
- We have problems with guaranteed random input. Sometimes you have a solution that happens to pass all 100 testcases with say 50% probability (and WAs on the other half). This is unfortunate, but does not prevent us from wanting to write and test such a solution.
- We have submissions that RTE/TLE/AC depending on whether pypy or cpython is used, because of recursion limit issues.
- We have non-intented 'bruteforce' submissions that are only included to test correctness of the answers, that may sometimes happen to pass the tests anyway. These should not be used to set the timelimit (e.g. using 2* slowest submission), so are moved to another directory.

thorehusfeldt:
I am late to this conversation; and much less experienced than many of you. 

(I have, however, given this topic some thought, and used my own script at [analyzetestgroups.py](https://github.com/Kodsport/testdata_tools/blob/master/analyzetestgroups.py) quite a bit. It focusses on `@EXPECTED_GRADES@` for problems with test groups, and I’ve found it incredibly useful in particular in co-developing with less experienced setters. Still, there are many annoying issues with this.)

Here’s a very different approach. It 
1. avoids cluttering source code with what is ultimately problem development information
2. allows rich specification of expected behaviour
3. is easily accessed and parsed by tools
4. allows changes to test group structure and expected submission behaviour during development (such as test group 2 being removed)
5. is decoupled from other specifications

The idea is that the `submissions` directory can contain `expected_verdicts.yaml` that enumerates (some) submissions and specifies behaviour consistent with the layout of `data/*`, much like `generators.yaml`. 

I want to be able to specify the _allowed verdicts_ per submission and per named test group. (An allowed verdict can be a list, such as `TLE, RTE`.) The tree structure of `data` allows inheriting verdicts downwards, so I can quickly specify “everything should get `AC`”, but I could also specify “should get AC except for the huge instances in `data/huge`, where it should `TLE` or `RTE`, which depends on the recursion limit.

If one right, this allows arbitrarily fine-grained specification (say, specifying that _this_ submission is guaranteed to fail on sample 3, but gets AC for the rest of the sample group.)

This would work (and be useful), independently of “we use test groups internally to organise test cases logically during development for pass/fail problems” or “we use test groups for graded problems with scoring and expose them to the solver.”

Here is an example, using fictional (and not complete thought-through) syntax:

```yaml
submissions:
  accepted/th.py:
    sample: AC
    data: AC
  partially_accepted/quadratic_time.py:
    sample: AC
    data:
        huge: TLE, RTE
        *: AC
  partially_accepted/simple_graph.py:
    sample: AC
      '3': WA
    data:
        non-simple: WA
        with-loops: WA
        *: AC
```

niemela:
Will close this. @thorehusfeldt is working on an updated suggestion.

PR #10, opened 2021-01-28
absolute -> relative link for problem.yaml example

ghamerly:
I think this fixes the link (absolute to relative).


RagnarGrootKoerkamp:
Looks like it does!

PR #12, opened 2021-04-30
PvP problem support

default-username-852:
This pull request defines a PvP problem and how it is to be defined.

This was developed by 8 students as part of a course at KTH which has finished. If someone wants to keep working on this, feel free to take what we have done and keep on building on it!


jsannemo:
@niemela I'm assuming this is obsolete and can be closed?

niemela:
> @niemela I'm assuming this is obsolete and can be closed?

Sadly, yes. (I would love to add something like this).

PR #14, opened 2022-09-03
Drop libraries from spec

niemela:
The libraries setting has very rarely (never?) been used, and is not implemented in most (all?) tools supporting the problem package format.


eldering (approved):
In DOMjudg we've never used this, so LGTM.

nickygerritsen (approved):

Issue #18, opened 2022-09-07
Introduce version field in problem.yaml


niemela:
Introduce something like `problem_format_version` in `problem.yaml`.

This would be a required going forward, but tools could (and likely will) treat missing `problem_format_version` as a "compatibility mode".

(See also discussions in #15)

jsannemo:
I like `problem_format_version`. `problem_spec_version` works too, but I think either is better than `format_version` or `spec_version` which is IMO too brief.

niemela:
Regarding the value and semantics I suggest:

* A version string on the format `<yyyy>-<mm>` (like what we have in the [Contest API](https://github.com/icpc/ccs-specs).
* The version must be changed whenever there is a non-backwards-compatible change
* The version should not be changed for "bugfixes", i.e. when the intended meaning does not change, only the explanation or examples.
* The version does not have to be changed for added features, but could/should be for significant amounts of added features.

I.e. pretty weak guarantees. Only "if you recognize the version then all the things you know will mean what they think you mean, but there could be things you don't know". So, you can't just look at the version and know if you will be able to use the package, but you *can* look and version to know if you will be able to tell whether you can use the package.

I think only the last item is controversial? Thoughts?

niemela:
Yeah, that's a very good point. It was kinda what I was getting at with the "significant amounts of added features" but that might be too vague/weak...

In fact we should bake the "subset" thing into this as well. Then the version string would be on the form `<yyyy>-<mm>[/<subset>]`. So e.g. `2022-09/icpc` would mean the icpc subset of the `2022-09` version of the format.

jsannemo:
I guess. The main use case I can imagine for that is to force problemtools to explicitly disregard everything not in the `icpc` subset (or even complain about the presence of such aspects in the problem), to ensure that problems used at the WF only use features all judges support. (I considered a lot of other use cases, but they seemed contrived, confusing or unlikely to be useful in practive -- but maybe you have thought of some).

I'm not 100% convinced that use case alone is worth it to be included in the versioning string rather than having a separate tool or just a verifyproblem flag for ICPC WF judges when testing their problems.

I also don't really care since I would never use the `/icpc` subset, so feel free to ignore my opinion.

jsannemo:
I'm fine with a judge accidentally including more features than bargained for; it's unlikely to be a problem in practice, unlike changing the meaning or removing functionality old problems may be using. However, I'd prefer bumping the version when new functionality is added.

[...]

Issue #15, opened 2022-09-06
Defaulting to Python 3


jsannemo:
Python 2 has now been EOL:ed for quite some time. Looking at e.g. NCPC 2020 Python 3 is >10x more popular than Python 2, and I expect this has only increased since.

I think it's time for the default Python version in the spec to be Python 3.

This is backwards-incompatible, but easily fixed for old problems: add a shebang on Python 2 code if the validator, submission etc crashes (typically they will have a `print blah` call that causes a syntax error).

Thoughts? e.g. @niemela @ghamerly @simonlindholm

simonlindholm:
Agreed. (And I'm usually *very* reluctant to make backward incompatible changes.)

Here's a script for automatically adding shebangs to Python 2 files that are missing them (or are using `#!/usr/usr/env python` to mean Python 2 which I don't know which behavior we want out of):
```py
import sys

dry_run = False
files = []
for arg in sys.argv[1:]:
    if arg.startswith("-"):
        if arg == "--dry-run":
            dry_run = True
        else:
            print("bad arg:", arg)
            sys.exit(1)
    else:
        files.append(arg)

for filename in files:
    if not filename.endswith(".py"):
        continue
    with open(filename) as f:
        source = f.read()
    first_line = source.split("\n")[0]
    if "python2" in first_line or "python3" in first_line:
        continue
    if dry_run:
        print("would convert:", filename)
    else:
        print("converting:", filename)
        if first_line.startswith("#!/"):
            source = "\n".join(source.split("\n")[1:])
        source = "#!/usr/bin/env python2\n" + source
        with open(filename, "w") as f:
            f.write(source)
```
Usage: `find -type f -name '*.py' -print0 | xargs -0 python3 add_python2_shebangs.py --dry-run`

niemela:
I think this would be a *very* good change.

Maybe we could introduce a "version" to the spec just before this. The idea being that version has to be increased for any non-backwards-compatible change. Then you could at least detect problems that are not using the latest version and in best case automatically translating them, works case manually doing so?

Thoughts?

On small but annoying issues is, what should the version field be called? I think that the obvious choice ("version") would be misinterpreted as the version of the problem.

jsannemo:
The verbose `problem_format_version` maybe?

That would also help with the other backwards incompatible (but probably not in practice) change (with reject/accept score).

deboer-tim:
Or `spec_version` to be shorter?

[...]

PR #19, opened 2022-09-07
Change default Python version to Python 3

jsannemo:
As discussed in https://github.com/Kattis/problem-package-format/issues/15.

This PR does not include any wording on versioning of the spec or compatibility; that will come as part of https://github.com/Kattis/problem-package-format/issues/18 instead.


jsannemo:
Replaced with new PR

Issue #17, opened 2022-09-06
Allowing run_time_error solutions to either WA, or TLE


jsannemo:
I thought this had been the case at some point. It's something I was bitten by recently in Coding Cup, having a solution that was incorrect (in a crashy way), but which sometimes manifested as a WA and sometimes as a TLE.

TLE also allows WA since we only want to verify that /some/ test case times it out, but the above case doesn't seem to be clearly mappable in the spec right now?

@niemela @simonlindholm

niemela:
If we allow submissions that claim to be RTE to either WA or TLE, then what are we really testing? Would having a "rejected" directory, with the requirement being that it is not accepted (i.e. it can RTE, WC, TLE, ...) solve this need?

jsannemo:
Yes, that's fine too.

I guess the reason why I'm fine with letting RTE get TLE and WA is that:
1. In C++, an RTE can often result in either
2. I can't ever remembered writing a solution where I want to *test* for RTE, unlike WA and TLE - solutions only get placed there because the happen to crash on a case (except they sometimes don't /always/ do...

In the end I don't care very much, either solves my problem, so pick whatever you prefer :)

jsannemo:
@RagnarGrootKoerkamp what does BAPC tools call this?

RagnarGrootKoerkamp:
I'm using the Domjudge way to handle this, using `@EXPECTED_RESULTS@:`, see [here](https://github.com/RagnarGrootKoerkamp/BAPCtools/blob/master/doc/implementation_notes.md).

jsannemo:
Are they placed in an arbitrary folder matching one of the expected verdicts?

+1 for `rejected` anyway

RagnarGrootKoerkamp:
Yes indeed, any of the matching folders works.

I think I have an assert somewhere that the folder it is in must be one of the listed verdicts.

`submissions/rejected` sounds convenient indeed

thorehusfeldt:
In case anybody is still here, I have a draft implementation that allows you to specify this in `<problemdir>/submissions/expected_grades.yaml' like this:

```
time_limit_exceeded/recursive.py:
  verdict:
    - RTE
    - TLE
```
or, if you like it terse:
```
time_limit_exceeded/recursive.py: ["RTE", "TLE"]
```

This can be specified down to individual testgroups, so you could also do, hypothetically:
```
time_limit_exceeded/recursive.py:
  subgroups:
    sample: AC
    secret:
      group1: AC
[...]

PR #20, opened 2022-09-07
Add rejected submission subdirectory

jsannemo:
As discussed in https://github.com/Kattis/problem-package-format/issues/17.

I added some normative guidance on the folder for users (since I think this spec is not use primarily by implementers, but by problem creators).


RagnarGrootKoerkamp (approved):

jsannemo:
@niemela PTAL?

eldering (approved):

Issue #16, opened 2022-09-06
Changing the meaning of accept_score and reject_score


jsannemo:
Scoring problems today allow the keys `accept_score` and `reject_score` for a test group. These set the default score of a test case for which no scoring validator exists.

When a scoring validator exists, they are instead treated as a fallback in case the validator, for some reason, doesn't output a score for a given test case. The number of times this makes sense should be small (we have never encountered it on PO).

I propose that accept_score and reject_score are instead treated as a multiplier to whatever the validator outputs, and that an accept score of `1` and a reject score of `0` is defined to be the default for a scoring problem if the output validator doesn't output anything. This is essentially compatible with the behaviour today, except for the case where a validator only sometimes outputs a score.

The benefit of this change is that for scoring problem where different groups have different maximum scores, a validator can be used to denote what percentage of that group's score should be given. This is used every now and then in olympiad problems, where for example constructing a valid solution vs determining whether a solution exists or not can give 100% vs 50% of the score. Today, this requires passing input validator flags per-group (not implemented in Kattis) or writing a custom grader. This seems like a simpler way.

As discussed with @niemela and @simonlindholm

simonlindholm:
Possibly contentious question: should we round the result after multiplying?

niemela:
I think this would be a good change.

@ghamerly @austrin Do you have any thoughts?

@jsannemo Do you want to make a PR?

ghamerly:
I have more limited (but not no) experience with using these flags, so I don't have a strong opinion. But I like the idea, and I like being able to balance the fraction of total score while avoiding special arguments to the scoring validator.

Just to be sure I understand, if a scoring problem does not have a custom scoring validator, then the score for a group would be `accept_score` (if it were accepted). (I.e. "essentially compatible", as you say.)

My only constructive comment is that the names (`accept_score` and `reject_score`) are no longer good names for this concept. They should really be `accept_score_multiplier` or something similarly meaningful.

jsannemo:
You understand correctly.

W.r.t naming that depends on whether we want both concepts to coexist or not. I think "accept_score" might still be fine: view it as the scoring validator outputting the multiplier rather than the other way around, with the default multiplier being "1" if no score is output (this is consistent with outputting no score for WA rather than zero, since we want 1*reject_score then).

thorehusfeldt:
I’d like to work on this.

niemela:
[...]

PR #21, opened 2022-09-07
Change accept_score and reject_score to be multipliers

jsannemo:
As discussed in https://github.com/Kattis/problem-package-format/issues/16.


spec/problem_package_format.md
```diff
@@ -348,8 +348,8 @@ group will be used. The format of `testdata.yaml` is as follows:
...
-| reject\_score                               | String                                                | 0            | Default score for rejected input files. May only be specified for scoring problems.                                                                                                                                                                                                                               |
+| accept\_score                               | String                                                | 1            | If an output validator outputs a score for an accepted input file, it is multiplied by the `accept_score` of the group. Otherwise, the score for the input file is the `accept_score` of the gruop. May only be specified for scoring problems.                                                                                                                                                                                                                               |
```
eldering, line 351:
```suggestion
| accept\_score                               | String                                                | 1            | If an output validator outputs a score for an accepted input file, it is multiplied by the `accept_score` of the group. Otherwise, the score for the input file is the `accept_score` of the group. May only be specified for scoring problems.                                                                                                                                                                                                                               |
```

jsannemo: Thanks @eldering , fixed

Issue #7, opened 2020-12-08
Specification for interactive problems


RagnarGrootKoerkamp:
Doesn't have to be exhaustive, but should mention the basics:

- `.interaction` files for samples and its syntax
- Samples *do not* have a corresponding `.in` file. For BAPCtools I was assuming this, but it seems that at least @simonlindholm thinks otherwise. I agree that not putting a `.in` file makes sense.
- Testdata for the samples can be provided in e.g. `attachments/1.in`.
- `validation: custom interactive` for the `problem.yaml`
- `\begin{Interaction} ... \end{Interaction}` for problem statement

simonlindholm:
> Samples do not have a corresponding .in file. For BAPCtools I was assuming this, but it seems that at least @simonlindholm thinks otherwise. I agree that not putting a .in file makes sense.

Example of a problem without .in: https://github.com/Kattis/problemtools/tree/develop/examples/guess/data/sample

jsannemo:
Reviving this a bit...

`validation: custom interactive` is in the spec now.

Let's document `.interaction`.

Surely `.in` and `.ans` should be *allowed* in the spec. PO surfaces sample results in the UI, and showing how you did on the sample case presented in the statement could be valuable.

RagnarGrootKoerkamp:
So, how about this:?

The `.interaction` is what is shown in the statement.

If a `.in` is present, that means submissions should be run on the sample, and the `.in` is fed to the interactor.
The `.in` should never be shown to contestants, since the interactor input is 'private'.

If a `.in` is present, an empty `.ans` must be present as well.
(Or alternatively, we forbid all `.ans` for interactive problems.)

jsannemo:
Yes, I agree.

Why must .ans be empty? I understand it most often is, but I'm sure there has been at least one instance where we've used it in PO because there was some natural separation between input to the interactor and answer to compare with?

RagnarGrootKoerkamp:
Ah yes you're right. The .ans could contain some hard-to-compute information that can help the interactor.

In that case `.ans` must always be present.

jsannemo:
@RagnarGrootKoerkamp will you make a PR for this?

RagnarGrootKoerkamp:
I think I'm too busy at the moment to do that; sorry

niemela:
I consider this done after merging #22.

@jsannemo @RagnarGrootKoerkamp do you disagree?

PR #22, opened 2022-09-07
Add wording on .interaction files

jsannemo:
As discussed in https://github.com/Kattis/problem-package-format/issues/7

@niemela @RagnarGrootKoerkamp

### DIFF
--- a/spec/problem_package_format.md
+++ b/spec/problem_package_format.md
@@ -305,6 +305,16 @@ information illustrating the related test file.
 
 Input, answer, description, hint and image files are matched by the base
 name.
+  
+### Interactive Problems
+
+For interactive problems, any sample test cases must provide an interaction protocol with the extension `.interaction` for each sample demonstrating the communication between the submission and the output validator, meant to be displayed in the problem statement.
+An interaction protocol consists of a series of lines starting with `>` and `<`.
+Lines starting with `>` signify an output from the submission to the output validator, while `<` signify an output from the output validator to the submission.
+                                                                                                 
+A sample test case may have only an `.interaction file without a corresponding `.in` and `.ans` file.
+However, if either of a `.in` or a `.ans` file is present the other one must also be present.
+Unlike `.in` and `.ans` files for non-interactive problem, interactive `.in` and `.ans` files are not meant to be displayed in the problem statement. 
 
 ### Test Data Groups
 
--- a/spec/problem_package_format.md
+++ b/spec/problem_package_format.md
@@ -312,7 +312,7 @@ For interactive problems, any sample test cases must provide an interaction prot
 An interaction protocol consists of a series of lines starting with `>` and `<`.
 Lines starting with `>` signify an output from the submission to the output validator, while `<` signify an input from the output validator to the submission.
                                                                                                  
-A sample test case may have only an `.interaction` file without a corresponding `.in` and `.ans` file.
+A sample test case may have just an `.interaction` file without a corresponding `.in` and `.ans` file.
 However, if either of a `.in` or a `.ans` file is present the other one must also be present.
 Unlike `.in` and `.ans` files for non-interactive problem, interactive `.in` and `.ans` files are not meant to be displayed in the problem statement.
 If you want to provide files related to interactive problems (such as testing tools or input files) you can use [attachments](#attachments).
--- a/spec/problem_package_format.md
+++ b/spec/problem_package_format.md
@@ -310,7 +310,7 @@ name.
 
 For interactive problems, any sample test cases must provide an interaction protocol with the extension `.interaction` for each sample demonstrating the communication between the submission and the output validator, meant to be displayed in the problem statement.
 An interaction protocol consists of a series of lines starting with `>` and `<`.
-Lines starting with `>` signify an output from the submission to the output validator, while `<` signify an output from the output validator to the submission.
+Lines starting with `>` signify an output from the submission to the output validator, while `<` signify an input from the output validator to the submission.
                                                                                                  
 A sample test case may have only an `.interaction` file without a corresponding `.in` and `.ans` file.
 However, if either of a `.in` or a `.ans` file is present the other one must also be present.
--- a/spec/problem_package_format.md
+++ b/spec/problem_package_format.md
@@ -312,7 +312,7 @@ For interactive problems, any sample test cases must provide an interaction prot
 An interaction protocol consists of a series of lines starting with `>` and `<`.
 Lines starting with `>` signify an output from the submission to the output validator, while `<` signify an output from the output validator to the submission.
                                                                                                  
-A sample test case may have only an `.interaction file without a corresponding `.in` and `.ans` file.
+A sample test case may have only an `.interaction` file without a corresponding `.in` and `.ans` file.
 However, if either of a `.in` or a `.ans` file is present the other one must also be present.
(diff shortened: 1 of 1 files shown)


spec/problem_package_format.md
```diff
@@ -305,6 +305,16 @@ information illustrating the related test file.
...
 Input, answer, description, hint and image files are matched by the base
 name.
+  
+### Interactive Problems
+
+For interactive problems, any sample test cases must provide an interaction protocol with the extension `.interaction` for each sample demonstrating the communication between the submission and the output validator, meant to be displayed in the problem statement.
+An interaction protocol consists of a series of lines starting with `>` and `<`.
+Lines starting with `>` signify an output from the submission to the output validator, while `<` signify an output from the output validator to the submission.
+                                                                                                 
+A sample test case may have only an `.interaction file without a corresponding `.in` and `.ans` file.
```
simonlindholm, line 315:
```suggestion
A sample test case may have only an `.interaction` file without a corresponding `.in` and `.ans` file.
```

spec/problem_package_format.md
```diff
@@ -305,6 +305,16 @@ information illustrating the related test file.
 
 Input, answer, description, hint and image files are matched by the base
 name.
+  
+### Interactive Problems
+
+For interactive problems, any sample test cases must provide an interaction protocol with the extension `.interaction` for each sample demonstrating the communication between the submission and the output validator, meant to be displayed in the problem statement.
+An interaction protocol consists of a series of lines starting with `>` and `<`.
+Lines starting with `>` signify an output from the submission to the output validator, while `<` signify an output from the output validator to the submission.
```
RagnarGrootKoerkamp, line 313:
```suggestion
Lines starting with `>` signify an output from the submission to the output validator, while `<` signify an input from the output validator to the submission.
```

spec/problem_package_format.md
```diff
@@ -305,6 +305,16 @@ information illustrating the related test file.
...
+                                                                                                 
+A sample test case may have only an `.interaction file without a corresponding `.in` and `.ans` file.
+However, if either of a `.in` or a `.ans` file is present the other one must also be present.
+Unlike `.in` and `.ans` files for non-interactive problem, interactive `.in` and `.ans` files are not meant to be displayed in the problem statement.
```
RagnarGrootKoerkamp, line 317:
```suggestion
Unlike `.in` and `.ans` files for non-interactive problem, interactive `.in` and `.ans` files are not meant to be displayed in the problem statement or a possible `samples.zip` file. 
If sample inputs are needed (e.g. for a local testing tool), provide them as `attachments/1.{in,ans}`.
```

jsannemo: `samples.zip` are not specified anywhere, but is a judge-specific implementation detail (I'm not even sure it's in the CCS spec).

I'm not sure it's necessary to specify here that people can use attachments for adding tests. I'm assuming a problem author is already familiar with a way to provide files (whether through attachments, problem statement links or distributing it to contestant machines) since they are providing the testing tool, so presumably they understand that they can provide the test cases the same way.

@niemela wdyt?

niemela: > (I'm not even sure it's in the CCS spec)

It is not.

> I'm not sure it's necessary to specify here that people can use attachments for adding tests.

It's not **necessary**, but it's correct and helpful (a.k.a. good advice), so why not mention it?

> I'm assuming a problem author is already familiar with ...

The learning curve to creating problem packages is already higher than what we would prefer. Why not make it simpler when we can?

jsannemo: I pushed a different suggested wording, wdyt? @RagnarGrootKoerkamp @niemela 

> The learning curve to creating problem packages is already higher than what we would prefer. Why not make it simpler when we can?

My point was nothing like that at all. It was that specifically suggesting adding testcases as attachments seemed superfluous since a problemsetter would anyway have to figure out a way to provide the referenced testing tool (thus my rewording as a more general pointer to attachments).

niemela: > My point was nothing like that at all. It was that specifically suggesting adding testcases as attachments seemed superfluous since a problemsetter would anyway have to figure out a way to provide the referenced testing tool (thus my rewording as a more general pointer to attachments).

Yeah, that's a really good point. 👍

RagnarGrootKoerkamp: @jsannemo yes LGTM

spec/problem_package_format.md
```diff
@@ -305,6 +305,17 @@ information illustrating the related test file.
 
 Input, answer, description, hint and image files are matched by the base
 name.
+  
+### Interactive Problems
+
+For interactive problems, any sample test cases must provide an interaction protocol with the extension `.interaction` for each sample demonstrating the communication between the submission and the output validator, meant to be displayed in the problem statement.
+An interaction protocol consists of a series of lines starting with `>` and `<`.
+Lines starting with `>` signify an output from the submission to the output validator, while `<` signify an input from the output validator to the submission.
```
eldering, line 313: This assumes that interaction is always fully line-based.

spec/problem_package_format.md
```diff
@@ -305,6 +305,17 @@ information illustrating the related test file.
...
+                                                                                                 
+A sample test case may have only an `.interaction` file without a corresponding `.in` and `.ans` file.
```
eldering, line 315: Nitpick, just to prevent reading this as "may only have" as I did.
```suggestion
A sample test case may have just an `.interaction` file without a corresponding `.in` and `.ans` file.
```

spec/problem_package_format.md
jsannemo: Yes, in the sense that it doesn't allow representing non-line-ending problems.

I don't think this is a problem in practice (at least I've never seen such a problem). I guess it's technically possible to add some marker for "this interaction didn't end in a newline", but that feels unnecessary to spec (for a visual-only feature) until such a problem arise, IMO.

spec/problem_package_format.md
jsannemo: SGTM, updated/

niemela (approved):

spec/problem_package_format.md
```diff
@@ -305,6 +305,17 @@ information illustrating the related test file.
...
+  
+### Interactive Problems
+
+For interactive problems, any sample test cases must provide an interaction protocol with the extension `.interaction` for each sample demonstrating the communication between the submission and the output validator, meant to be displayed in the problem statement.
+An interaction protocol consists of a series of lines starting with `>` and `<`.
+Lines starting with `>` signify an output from the submission to the output validator, while `<` signify an input from the output validator to the submission.
+                                                                                                 
+A sample test case may have just an `.interaction` file without a corresponding `.in` and `.ans` file.
+However, if either of a `.in` or a `.ans` file is present the other one must also be present.
+Unlike `.in` and `.ans` files for non-interactive problem, interactive `.in` and `.ans` files are not meant to be displayed in the problem statement.
```
eldering, line 317: So then what is the purpose/use of `.in` and `.ans` files in this case?

eldering, line 317:
```suggestion
Unlike `.in` and `.ans` files for non-interactive problem, interactive `.in` and `.ans` files are not meant to be displayed in the problem statement, but just used as input files for the output validator.
```

niemela: `.in` and `.ans` files are specified elsewhere (at the very top), and the use here is the same. The only exception is that `.in` and `.ans` should not actually be displayed in the problem statement where there is an `.interaction`.

PR #44, opened 2023-07-21
First stab at global constants

### LINKED ISSUE #34: Allow global constants in the package
E.g. reusing numbers in the statement and a validator.

Suggested by @jsannemo.

evouga:
@jsannemo 

fixes #34


RagnarGrootKoerkamp:
Maybe you already considered this, but could we not write a `constants.h` (and `constants.tex` etc) that can be included by whatever needs it. That way cpp and latex files can still keep their usual valid syntax, and be nice to editors.

The formatting Syntax also looks a bit convoluted. Something python-like `{n:format-specifier}` would be nicer imo.

(Generally I think the display formatting is only relevant for latex? A simpler solution would be introducing a separate string variable for this, although slightly ugly, especially when different languages use different formatting.)

RagnarGrootKoerkamp:
We had some discussion:
- Building a 'header' file per language doesn't work for ie markdown statements

evouga:
I've made changes per our discussion.

spec/problem_package_format.md
```diff
@@ -216,6 +217,36 @@ languages.
...
+These format flags must only appear in constant expressions whose values are integer or floating-point numbers.
+
+#### Examples
+For the constant `max_n` with value `5000000`:
+| Constant Sequence | Replacement |
+|-------------------|-------------|
+| `@@max_n@@` | `5000000` |
+| `@@max_n@s"\,"@@` | `5\,000\,000` |
+| `@@max_n@s,w4@@` | `500,0000` |
+| `@@max_n@e@@` | `5 \times 10^6` |
```
eldering, line 247: This example uses both the old `@@` tag and underscores.

spec/problem_package_format.md
```diff
@@ -139,6 +139,7 @@ stated. Any unknown keys should be treated as an error.
...
 | <s class="kattis"> <s>languages</s>                          </s> | <s class="kattis"> <s>String or sequence of strings</s> </s> | <s class="kattis"> <s>all</s>                                              </s> | <s class="kattis"> <s>Set of languages or "all".</s>                                                                                                                                                                                                                                                                                                              </s> |
+| constants                                 | map of strings |  | Global constant values used by the problem. See definition below. |
```
eldering, line 142: Shouldn't this be something like "map of strings to int|float|string"?

spec/problem_package_format.md
evouga: removed the stale examples

RagnarGrootKoerkamp (approved):
LGTM; will also need updating the cue spec

eldering (approved):

spec/2023-07-draft/problem_package_format.md
```diff
@@ -218,6 +219,26 @@ languages.
 
 </div>
 
+### constants
```
jsannemo, line 222:
```suggestion
### Constants
```

spec/2023-07-draft/problem_package_format.md
```diff
@@ -218,6 +219,26 @@ languages.
...
+  - included code
+  - example submissions
+  - `testdata.yaml`
+    
+Constant sequences are **not** replaced in test data files or in `problem.yaml` itself.
+
+The `name` in every constant sequence must be a valid constant name, and `format` must consist of exactly one of the following characters: `e`, `s`, or `c`. If a `format` is not specified, the constant sequence is replaced verbatim by the corresponding value. Format flags have the following effects:
+| Flag                      | Meaning |
+|---------------------------|---------|
+| `e`                       | Formats an integer or floating-point value in Latex using scientific notation (as `a \cdot 10^{b}`). For integer values, `a` is an integer and `b` is the number of trailing zeros in the value. For floating-point numbers, the mantissa and exponent follow the usual IEEE 754 conventions. |
```
jsannemo, line 236:
```suggestion
| `e`                       | Formats an integer or floating-point value in LaTeX using scientific notation (as `a \cdot 10^{b}`). For integer values, `a` is an integer and `b` is the number of trailing zeros in the value. For floating-point numbers, the mantissa and exponent follow the usual IEEE 754 conventions. |
```

spec/problem_package_format.md
```diff
@@ -216,6 +217,34 @@ languages.
 
 </div>
 
+### constants
+
+A map of names to values. Names must not contain whitespace, and neither names nor values shall contain the character `@`. *Constant sequences* have the form `@@name@@` or `@@name@format@@`, where `name` is a string not containing whitespace or the character `@` and `format` is specified below. All instances of constant sequences in the following files will be replaced by the value of the corresponding constant, formatted as described below:
+  - problem statements
+  - input and output validators
+  - included code
+  - `testdata.yaml`
```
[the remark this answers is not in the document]
evouga: There is already an "example submissions" in the list? I used this name because it's the current heading for judge submissions in the spec, but I agree it's unclear. We want to be clear that only the *judge* submissions will get replaced (not contestant submissions).

RagnarGrootKoerkamp (approved):

pehrsoderman (approved):

RagnarGrootKoerkamp:
Are we ready to merge this? I think we all feel this one may be slightly more controversial/risky than other changes we did, but I'd saw we can merge for now, implement it, and see how it goes.

niemela:
> Are we ready to merge this? I think we all feel this one may be slightly more controversial/risky than other changes we did, but I'd say we can merge for now, implement it, and see how it goes.

Technically we still call this version `2023-07-draft`, until we remove `-draft` we can be a bit more "careless". So, with that in mind, I think it's better to merge this, it will make it easier for people to read the draft and maybe come up with a great argument against it.

Could you rebase so we can merge it?

Tagl (approved):
Looks good, but in the future I would like to explore how to flexibly support different languages (localization).

niemela:
I have created #118, with the exact same contents as this PR, and will merge it now.

Issue #30, opened 2023-07-21
Create CUE specification for testdata.yaml and problem.yaml


thorehusfeldt:
Here are some ideas:

```cue
// Defines valid contents of testdata.yaml
package problemformat

import "strings"
import "strconv"

#filename: =~ "^[a-zA-Z0-9][a-zA-Z0-9_.-]*[a-zA-Z0-9]$"
#path: =~ "[a-zA-Z0-9_.-/]*"

#testdata_settings: {
	on_reject: *"break" | "continue"
	grading: *"default" | "custom"
	grader_flags:  *"" | string
    if grading == "default" { grader_flags? : #default_grader_flags }
	input_alidator_flags: *"" | string
	output_validator_flags: *"" |string
	accept_score: *"1" | #score
	reject_score: *"0" | #score
	range: *"-inf +inf" | string
    // Verify that the scores and ranges make sense:
    _range_syntax: strings.Split(range, " ") & [#score, #score] // space-separated scores  
    _lo: strconv.ParseFloat(strings.Split(range, " ")[0], 64)   // parses to float (including '-inf')
    _hi: strconv.ParseFloat(strings.Split(range, " ")[1], 64)
    _wa: strconv.ParseFloat(reject_score, 64)
    _ac: strconv.ParseFloat(accept_score, 64)
    _order: true & _lo <= _wa  & _wa <= _ac & _ac <= _hi
}
// matches "1", "06", "21", ".4", "-1.2", but not 'inf'
#score: =~ "^-?([0-9]+|[0-9]*.[0-9]+)$" 

// Default grader
// --------------

#default_grader_flags: this={
    string  
    _as_struct: { for w in strings.Split(this, " ")  { (w): _ } }  // convert to struct and ...
    _as_struct: #valid_default_grader_fields                       // ... validate its fields
    }

// Default grader flags (converted to fields of a CUE struct for validation)
#valid_default_grader_fields: {
    #verdict_aggregation_mode?   // at most one verdict aggregation mode
    #score_aggregation_mode?     // at most one score aggregation mode
    ignore_sample?: _            // two more optional flags
    accept_if_any_accepted?: _
}

#verdict_aggregation_mode: {first_error: _ } | *{worst_error: _  } | {always_accept: _ }
#score_aggregation_mode:  {min: _ } | {max: _ } | *{sum: _ } | {avg: _ }
```

and here’s a specification of `problem.yaml`

```cue
#problem_settings: {
[...]

PR #45, opened 2023-07-22
Schemas for problem.yaml, testdata.yaml

thorehusfeldt:
Add formal specifications for
* `problem.yaml` in [spec/schemas/problem.cue](https://github.com/Kattis/problem-package-format/compare/master...thorehusfeldt:problem-package-format:master#diff-bb0250814460bc90e45e4bfd930df57fa5037546e079cc344ea49e393f6b6856)  
*  `testdata.yaml` in [spec/schemas/testdata_settings.cue](https://github.com/Kattis/problem-package-format/compare/master...thorehusfeldt:problem-package-format:master#diff-37015d52e0429546558e36cfde5c9020e2bc2070f2afb2fb0ac318157fc2ce28)

Notable changes after working group meeting 21 Jul 2023 (Lund):
1. made `""` the default for `validator_flags` (consistent with other flags)
2. made `*"all"` the default for `languages`, note that `string` is _not_ an option (but `[...string]` is).
3. the type for `time_multiplier` and `time_safety_margin` is now “nonnnegative number”
4. the `range` keys are checked for internal consistency
5. the `grading`/`scoring` key is removed

Things to discuss:
1. <s>`input_validator_flags` may change to `input_validators`.</s>
2. should there be a fixed regex for numbers? (Because we may want to think about `005.15` or `1e-15`.)
3. Consider changing the values of `validation` to `*"default" | "custom" | "interactive" | "score" | "interactive score"` (shorter, but also less lenient)


RagnarGrootKoerkamp:
discussion notes:

```
problem.yaml:
- validator_flags

PROPOSAL:

testdata.yaml
- input_validator_flags: <flags>
- input_validator_flags:
  <name>: <flags>
  syntax.ctd: --skip
- output_validator_flags

NOTES:
- deprecate validator_flags from problem.yaml
- only allow a single output validator
- input_validator_flags: <flags> means flags for all validators
- input_validator_flags: <dict> specifies flags for some validators, but all are run anyway
- output_validator_flags: <flags> specifies flags for the single output validator
- output_validator_flags: <dict> is *not allowed*

What about testdata.yaml inheritance?
- Update the spec to explicitly say: if testdata.yaml is not present in a directory=testgroup, inherit it from the closest ancestor
- ICPC spec: only testdata.yaml in data/, with only allowed key being output_validator_flags
```

thorehusfeldt:
This pull request addresses issue #30

spec/schemas/problem.cue
```diff
@@ -0,0 +1,49 @@
+import "strings"
+
+#problem_settings: {
+    name: string | close({ [#language_code]: string })
+    problem_format_version?: *"legacy" | "draft" | =~ "\d\d\d\d-\d\d(-draft)?"
```
jsannemo, line 5: Pessimistic assumption that the spec won't be updated in e.g. the year 10323.

spec/schemas/problem.cue
```diff
@@ -17,9 +17,9 @@ import "strings"
     //} 
     limits?: #limits
     validation?: *"default" | "custom" | {
-        multipass: *false | true
-        interaction: *false | true
-        scoring: *false | true
+        multipass?: *false | true
+        interaction?: *false | true
```
niemela, line 21:
```suggestion
        interactive?: *false | true
```

spec/schemas/problem.cue
```diff
@@ -17,9 +17,9 @@ import "strings"
...
+        scoring?: *false | true
```
niemela, line 22: I wonder how often there is a problem with `type = scoring` and a validator that is *not*.  @jsannemo do you know of any such examples?

niemela (changes requested):

spec/schemas/problem.cue
simonlindholm: Most of our problems that use output validators are of that kind. EGOI carnivalgenerals, bikesvscars and sopsug, for instance.

niemela: Ah, well then, never mind...

spec/schemas/problem.cue
thorehusfeldt: Changed to ` =~"[0-9-]+(-draft)?"`

niemela: Not very important, but that just seems less precise.

I'll rather make an update when needed (in 7976 years).

thorehusfeldt:
(I think all of @niemela ’s change requests were addressed and resolved, yet the “Changes requested” warning remains red. Maybe only FN can approve this? I’m asking him to re-review.)

spec/schemas/problem_package.cue
```diff
@@ -0,0 +1,56 @@
...
+
+	license?: *"unknown" | "public domain" | "cc0" | "cc by" | "cc by-sa" | "educational" | "permission"
+	if license != "public domain" {rights_owner?: string}
+
+	// TODO (CUE version > 0.5.0)
+	// if license =~ "unknown|public domain" {
+	//    numexists(>=1, rights_owner, author, source)
+	//} 
+	limits?:     #limits
+	validation?: *"default" | "custom" | {["multipass" | "interactive" | "scoring"]: *false | true}
```
niemela, line 19: Isn't this a little bit strange, now when all the keys are just bools?

What I mean is that we say `custom` (which is obviously "equivalent" to `custom: true`), but we do say `multipass: true`.

spec/schemas/problem_package.cue
```diff
@@ -0,0 +1,56 @@
+import "strings"
+
+#problem_settings: {
+	name:                    string | close({[#language_code]: string})
+	problem_format_version?: *"legacy" | "draft" | =~"[0-9-]+(-draft)?"
```
niemela, line 5: I think `[0-9-]+(-draft)?` is less good than `\d\d\d\d-\d\d(-draft)?`. The latter more precisely specifies what we actually expect.

Also, I think we (including me) are overthinking this. We don't *have* properly support any future name now, we can change this if we realize in the future that we want to allow more. But right *now* we are thinking that all version names will be `legacy`, `draft`, or `\d\d\d\d-\d\d(-draft)?`, so I think we should say that.

niemela (approved):
I'm fine with this as is, but see comments...

spec/schemas/problem_package.cue
thorehusfeldt: I have no strong opinion on this and welcome alternatives. 

## what we have

Here are some things you can do with the above:
```yaml
validation: "default" # probably seldom used, because it's default
validation: "custom"
validation:
    multiplass: true
validation:
    interactive: true
    scoring: true
    multipass: true
```
This makes sense to me. 

## alternative 1
However, an alternative is:
```cue
custom-validation: { 
    custom : true
     [ "multipass" | "interactive" | "scoring"]: *false | true  
```
Then the above would become:
```yaml
# no way to explicitly say "validation: default"
costum-validation:
    custom: true
validation:
    custom: true
    multiplass: true
custom-validation:
    custom: true
    interactive: true
    scoring: true
    multipass: true
```
This is  very explicit but feels verbose

## alternative 2
However, an alternative is:
```cue
use-default-output-validator?: *true | false
if use-default-output-validator == false:
    custom-validation?: {  [ "multipass" | "interactive" | "scoring"]: *false | true  
```
Then the above would become:
```yaml
use-default-output-validator: false
use-default-output-validator: true

custom-validation:
    multipass: true

use-default-output-validator: true # redundant, but ok to specify
custom-custom-validation:
    interactive: true
    scoring: true
    multipass: true
```

I have no strong opinion about these.

spec/schemas/problem_package.cue
```diff
@@ -1,56 +1,54 @@
-import "strings"
-
-#problem_settings: {
+#problem_settings_icpc: {
 	name:                    string | close({[#language_code]: string})
-	problem_format_version?: *"legacy" | "draft" | =~"[0-9-]+(-draft)?"
-	type?:                   *"pass-fail" | "scoring"
+	problem_format_version?: *"legacy" | "draft" | =~"^[0-9]{4}-[0-2][0-9](-draft)?$"
```
thorehusfeldt, line 3: If @jsannemo insists, this can become `^[0-9]+-` etc.

spec/schemas/problem_package.cue
```diff
@@ -1,56 +1,54 @@
...
-	//    numexists(>=1, rights_owner, author, source)
-	//} 
-	limits?:     #limits
-	validation?: *"default" | "custom" | {["multipass" | "interactive" | "scoring"]: *false | true}
-	keywords?:   string | [...string]
-	uuid?:       string
-	languages?:  *"all" | [...string]
+	if license != _|_ && license != "public domain" {
+		rights_owner?: string
+		// if license =~ "unknown" {  numexists(>=1, rights_owner, author, source) }
```
thorehusfeldt, line 11: This would be nice to specify, but isn’t worth it in CUE 0.5.0 because `numexists` is not implemented (and the workarounds very verbose.)

spec/schemas/problem_package.cue
```diff
@@ -1,56 +1,54 @@
...
+			time_limit_to_tle: *1.5 | float
+		}
+		time_limit?:      number & >0
+		time_resolution?: *1.0 | float
+		[#other_limits]:  int
+	}
+	validation?: *"default" | "custom" | {
+		"interactive": *false | true
+		... // other keys are allowed but ignored
+	}
```
thorehusfeldt, line 25: From the perspective of the ICPC subset, this is not optimal. A cleaner (but more verbose) version is:

```cue
custom-validation?: {
  custom: false | true
  interactive?: *false | true
  ...
}
```
but it’s not clear to me what the defaults or optionality should be.

spec/schemas/problem_package.cue
niemela: I think "what we have" is pretty good. Minor clarification request though. I think all of these are legal, right?:
```yaml
validation: default # probably seldom used, because it's default
validation: custom
validation:
    multipass: true
validation:multipass: true
validation:
    interactive: true
    scoring: true
    multipass: true
```

I don't like "alternative 1" as much:
- allowing explicit `validation: default` is nice (but not very important)
- `custom: true` is implied by any other value being true, but if I understand correctly you're not required to always set `custom: true` (nor are you forbidden from doing it).
- You're not actually suggesting that they key should be called `custom-validation`, right?

I don't understand what your "alternative 2" suggestion is. Why are there 3 different keys (`use-default-output-validator`, `custom-validation`, `custom-custom-validation`)?

spec/schemas/problem_package.cue
```diff
@@ -1,56 +1,54 @@
...
-		"validation_memory" |
-		"validation_output",
-	]: "int"
+#testdata_settings_icpc: output_validator_flags: *"" | string
+
+// Full specification extends the ICPC subset:
+
+#problem_settings: {
+	#problem_settings_icpc
+	type?:       *"pass-fail" | "scoring"
```
thorehusfeldt, line 40: Note that the key `type` is _forbidden_ in the `icpc` subset if we do it like this.

spec/schemas/problem_package.cue
```diff
@@ -1,56 +1,54 @@
...
 		score?: number
-	max_score:              number
-	aggregation:            "sum" | "min"
-	verdict:                "first_error" | "accept_if_any_accepted"
-	input_validator_flags:  *"" | string | {[string]: string}
-	output_validator_flags: *"" | string
-}
+	max_score:             number
+	aggregation:           "sum" | "min"
+	verdict:               "first_error" | "accept_if_any_accepted"
```
thorehusfeldt, line 52: `verdict` is not a good name. Even `mode` is better. For symmetry, consider `score-aggregation` and `verdict-aggregation`.

spec/schemas/problem_package.cue
```diff
@@ -1,56 +1,54 @@
...
+	type?:       *"pass-fail" | "scoring"
+	validation?: close({["multipass" | "interactive" | "scoring"]: *false | true})
+	keywords?:   string | [...string]
+	languages?:  *"all" | [...string]
 }
 
 #testdata_settings: {
+	#testdata_settings_icpc
 	grading?:
 		score?: number
```
thorehusfeldt, line 49: This is now a one-key optional field, which seems structurally unnecessary.

RagnarGrootKoerkamp:
If we only allow bools (and not `multipass: 2`), I think passing a list (set) of flags is simple:
```
validation: default
validation: custom
validation: interactive
validation: multipass
validation: [scoring, interactive, multipass]
```

side question: how does `validation: scoring` interact with `type: scoring`? Does one imply the other? Is one of the two redundant?

spec/schemas/problem_package.cue
niemela: Super duper nitpick; why `...[0-2]...`? There are only 12 months? I would understand either `...[0-1]...`, or `...[0-9]...`.

(And I think I would prefer `^[0-9]{4}-[0-9]{2}(-draft)?$`.)

spec/schemas/problem_package.cue
niemela: I think it's fine. The ICPC subset *does* live in the reality that it is a subset of a greater whole.

spec/schemas/problem_package.cue
RagnarGrootKoerkamp: `validation:multipass: true` this is not valid yaml sadly

Tagl:
> side question: how does `validation: scoring` interact with `type: scoring`? Does one imply the other? Is one of the two redundant?

Scoring in validation implies the output validator also returns a custom score for each testcase, which can then be used by the score aggregation method. The default score is 1. Scoring in type states that the problem is not pass-fail.

thorehusfeldt:
Let’s make “what should `validation` be a separate issue.

`validation: scoring` means “there is a cusom output validator and it produces `score.txt`.

OTOH, `type:scoring` works fine with the default output validator (in fact, that setup accounts for most of the problems I know), but means “the default _grader_ should aggregate scores”. (The scores can be `1` for every `AC`, and please sum them. That’s the current default if you say nothing else.)

But `validation: scoring` makes no sense without `type: scoring`.
——

I’d be happy to reduce some of this redundancy and conceptual overlap.

spec/schemas/problem_package.cue
niemela: Could you elaborate on why `verdict` is not good?

I don't really like `mode`, because as you are alluding both `aggregation`, and `verdict` describes different modes of operation. I kinda like `score-aggregation` and `verdict-aggregation` t.b.h.. @jsannemo is it too long?

(Maybe `score-agg` and `verdict-agg`? I'm only 80% joking).

spec/schemas/problem_package.cue
niemela: I don't understand, can you explain?

spec/schemas/problem_package.cue
thorehusfeldt: A `verdict` to me should be something like `AC` or `WA`, not “how is the verdict computed”, which should be `verdict-policy` or something like that.

But it’s a rare enough key that I don’t want to make a fuss about this. (It’s possible that it will _never_ appear in any yaml ever, because of the sensible defaults.)

spec/schemas/problem_package.cue
thorehusfeldt: We now have
```yaml
grading:
    score: 45
max_score: 54
```
If `grading` can never have any other subkey, we might as well say
```yaml
score: 45
max_score: 54
```

spec/schemas/problem_package.cue
niemela: > `validation:multipass: true` is not valid yaml sadly

Really? 😞 The random YAML validator I found online accepted it.

spec/schemas/problem_package.cue
niemela: > If `grading` can never have any other subkey, we might as well say

*If*, sure, but it *can* have `max_score`, `aggregation`, and `verdict`. I feel I'm missing something here?

spec/schemas/problem_package.cue
thorehusfeldt: For [reasons], I cannot specify a schema extension that removes an earlier restriction on a key. 

Whatever conforms to `-full` must also conform to `-icpc`, otherwise the latter is not extension. (Extensions can allow _more keys_ but cannot _undo restrictions on “inherited” keys_.)

So what one _can_ do is add new keys. So `-full` can have a `type: "pass-fail" | "scoring"` (or better, a `scoring: *false | true` field) that `-icpc` lacks.

All of this would go away if a problem could say “I want to be understood as an ICPC-compliant problem.”. For instance by specifying that in the versions string or adding a `subversions` field or a `icpc: boolean` field. There was no consensus about those proposals.

spec/schemas/problem_package.cue
thorehusfeldt: > But if it's rare then it should be fine to make it longer?

Wow. Actually a great argument.

spec/schemas/problem_package.cue
thorehusfeldt: Is
```yaml
validation: { multipass: true }
```
valid?

RagnarGrootKoerkamp: yes, that works

spec/schemas/problem_package.cue
RagnarGrootKoerkamp: > It’s possible that it will never appear in any yaml ever, because of the sensible defaults

is it rare enough to drop?

niemela: > is it rare enough to drop?

No, I don't think so. It is in fact used some.

spec/schemas/problem_package.cue
eldering: Let's not let the restrictions of CUE prevent us from doing what we think is best.

I'd opt (as I understand with Fredrik) for making `type: pass-fail` an optional key/value in the ICPC subset. I'm fine that in the CUE specification this then allows any other value as well, or some other minor difference from what we intend.

thorehusfeldt:
Let me make another pitch for allowing `problem.yaml` to explicitly signal “I am an ICPC-conformal problem.”

_How_ this is signalled is not important,  but for instance `version: "2023-07-icpc-draft"` is one way. If we have that then we can always allow the optional `type` field in `problem.yaml`, but also restrict its values like this:

```cue
type?: *"pass-fail" | "scoring"
if version =~ "icpc": { type?: "pass-fail" }
```

Thus, a `problem.yaml` for an ICPC-problem _can_ specify its version, _can_ in that version string include `icpc`, and _can_ be explicit about `type: "icpc"`. But tooling _can_ infer the default value of the optional `type` field as `pass-fail`. 

The only thing that is forbidden is an explicit `type: scoring` when `version` matches `icpc`. Exactly what we want.

Other problems can be `pass-fail` _without_ also promising `icpc`. On the other hand, `icpc` would forbid many of the `-full` fields (such as some `validation` keys, `languages`, etc.)

To me, this exactly what we want. Explicit, clear, readable, no overhead. Easier to write tooling, easier to communicate among humans.

spec/schemas/problem_package.cue
jsannemo: I expect it to be used ~0 times (the defaults are chosen to be compliant with the normal IOI scoring), so feel free making it a UUID or something other absurd thing.

niemela: Ok, I think an UUID would be worse than necessary, but we could absolutely do `score-aggregation` and `verdict-aggregation` then?

spec/schemas/problem_package.cue
RagnarGrootKoerkamp: The point is that the only allowed key in `grading:` is `score:`, so we may as well drop the nesting and directly put `score:` into the surrounding scope, which indeed makes sense to me.

niemela: Exactly the current structure according to the spec is:

```yaml
grading:
  score: 1
  max_score: 100
  aggregation: sum
  verdict: first_error
input_validator_flags: none
output_validator_flags: none
```

So removing `grading` because `score` is the only thing in there is just wrong. OTOH, we *could* remove `grading` because there are only a total of 6 keys, do we really need two levels? If so, we would get:

```yaml
score: 1
max_score: 100
aggregation: sum
verdict: first_error
input_validator_flags: none
output_validator_flags: none
```

I honestly don't know which one I prefer...

jsannemo:
I like that all grading things are in one subkey. It also makes it only a
single key that "really" should be ignored by the ICPC subset (both input
and output validator flags makes sense to keep IMO).

niemela:
> I like that all grading things are in one subkey. It also makes it only a single key that "really" should be ignored by the ICPC subset (both input and output validator flags makes sense to keep IMO).

Yes, those are (at least) two good points.

thorehusfeldt:
Something else that happened is that `examples/problem_settings` contains a half a dozen `problem.yaml` examples, see [https://github.com/thorehusfeldt/problem-package-format/tree/master/examples/problem_settings](https://github.com/thorehusfeldt/problem-package-format/tree/master/examples/problem_settings). These are automatically checked against the schema by running

```sh
bash test.sh
```

in [https://github.com/thorehusfeldt/problem-package-format/blob/master/spec/schemas/test.sh](https://github.com/thorehusfeldt/problem-package-format/blob/master/spec/schemas/test.sh), provided you have `cue` installed.

thorehusfeldt:
Allow me to draw you attention to `examples/problem_icpc.yaml`, which contains a bunch of icpc-valid `problem.yaml`:

```yaml
# Minimal
name: Hello World!
---
# Typical
name: Hello World!
license: cc by-sa
author: Alice McAuthorson
---
# Can set type, but it's redundant
name: Hello World!
type: pass-fail
---
# Use all available fields
name: 
  en: Maximally Specified ICPC Problem
  de: Maksimal spezifiziertes ICPC-Problem
problem_format_version: 2023-07
type: pass-fail
author: Maxine Maximus
source: MCPC 2023
source_url: http://maximal-contest.2023.com
license: cc by-sa
rights_owner: Momma Maximus
limits:
  time_multiplier: 
    ac_to_time_limit: 3.0
    time_limit_to_tle: 1.4
  time_limit: 1
  time_resolution: .5
  memory: 1024
  validation_time: 21
validation:
  interactive: true
keywords: [greedy, interactive, needs big integers]
constants:
  max_n: 1000
  bases: ACTG
```
A similar file (for the full spec) is in `examples/problem_full.yaml`. This gives you a good overview of where the syntax is by now. (All of these validate, see `schema/test.sh`.)

You can add more examples if you find that instructive. I’ll make a long file of rejected examples _real soon now_.

thorehusfeldt:
# Ready to merge

I think this PR is ready to be merged; I can’t really improve this much.

What’s new in this branch:

## YAML examples

Three commented YAML files are in https://github.com/thorehusfeldt/problem-package-format/tree/master/examples:

1. [`problem_icpc.yaml` ](https://github.com/thorehusfeldt/problem-package-format/blob/master/examples/problem_full.yaml) contains a handful of example `problem.yaml` for the ICPC subset
2. [`problem_full.yaml`](https://github.com/thorehusfeldt/problem-package-format/blob/master/examples/problem_icpc.yaml) contains a handful of example `problem.yaml` for the full specification (but not in the ICPC subset)
3. [problem_invalid.yaml](https://github.com/thorehusfeldt/problem-package-format/blob/master/examples/problem_invalid.yaml)  contains commented examples of invalid `problem.yaml`.


## CUE schema

The main contribution is a schema in https://github.com/thorehusfeldt/problem-package-format/blob/master/spec/schemas/problem_package.cue using the [CUE language](https://cuelang.org). 

It’s a concise and human-readable way to solidify our specification. Moreover, with cue installed, you can run

```bash
cue vet problem.yaml problem_package.cue
```
to “vet” (=validate) a `problem.yaml` file against the schema. 

You can restrict to the ICPC subset like this:
```bash
cue vet problem.yaml problem_package.cue -d "#icpc"
```

You can also validate testdata settings:
```bash
cue vet testdata.yaml problem_package.cue -d "#testdata_settings"
```

## Testing

The bash script [test.sh](https://github.com/thorehusfeldt/problem-package-format/blob/master/spec/schemas/test.sh) tests that indeed the examples behave as they should against the CUE specification.

jsannemo:
"# Typical" should probably include the problem format version (since it's otherwise a legacy problem.yaml).

spec/schemas/test.sh
```diff
@@ -0,0 +1,55 @@
...
+	echo "WARNING: Not all full examples validated"
+fi
+cue vet ../../examples/problem_icpc.yaml problem_package.cue  -d "#icpc"
+if [ $? -eq 0 ]; then
+	echo "OK: ICPC examples accepted"
+else
+	echo "WARNING: Not all ICPC examples validated against #icpc"
+fi
+
+output_dir="tmp"
```
RagnarGrootKoerkamp, line 14: You could use `mktemp -d` instead to get an actual temporary directory, and delete it again at the end of the script

spec/schemas/test.sh
```diff
@@ -0,0 +1,55 @@
...
+# is wrong or the example should be moved to examples/problem_icpc.yaml)
+
+counter=0
+awk -v RS='---' -v output_dir="$output_dir" '
+	{
+		file_counter++;
+		output_file = output_dir "/full/problem_" file_counter ".yaml"
+		print $0 > output_file;
+
+	}' ../../examples/problem_full.yaml
```
RagnarGrootKoerkamp, line 28: hrmm; so I take it `cue` can't reject the individual parts of a multipart yaml file? unfortunate but somewhat inevitable.

spec/schemas/test.sh
```diff
@@ -0,0 +1,55 @@
+cue vet ../../examples/problem_full.yaml problem_package.cue
```
RagnarGrootKoerkamp, line 1: We could add github CI to run this in some docker container with cue installed on every PR/commit to master.

examples/problem_full.yaml
```diff
@@ -0,0 +1,83 @@
+# All of these validate against the schema
```
RagnarGrootKoerkamp, line 1: We could replace the example `problem.yaml` files with these files, ideally by including them using a liquid template or so. I can figure that out as a follow up.

RagnarGrootKoerkamp (approved):

spec/schemas/test.sh
thorehusfeldt: I actually found it _more_ useful to have those files around for debugging. Ideally this would be flag. No strong opinion.

spec/schemas/test.sh
thorehusfeldt: Right now, the `cue vet` CLI bails at first error. It’s not part of the CUE spec, but an implementation detail of tooling.

examples/problem_full.yaml
thorehusfeldt: If you want to see where I want the documentation to go, check out https://problem-package-format-copy.readthedocs.io/en/latest/. Ideally, all directives like

```
.. code-block :: yaml
   bla
```

should be extracted by `test.sh` and validated against the schema. But this uses RTD, not markown; it could open up for

```
.. code-block :: yaml
    :schema: #icpc
   bla
```

I think of this as nice-to-have, low-priority.

RagnarGrootKoerkamp: side note: I saw that readthedocs can also support markdown with some plugins. That would be nice, if it can produce the same structured text as the `rst` sources

RagnarGrootKoerkamp: Oh, also just found this:
https://ccs-specs.icpc.io/
build as
https://github.com/icpc/ccs-specs/blob/gh-pages/dev-notes.md

thorehusfeldt: By the way, 
```rst
.. literalinclude:: path/to/problem.yaml
  :language: yaml
```
works out of the box with RST.

I’ll make a complete, small example problem in my fork whose various files are included in the documentation just so show this off. (Then CI could `verifyproblem` the problem and `cue` all the `yaml`. The sky is the limit.)

thorehusfeldt: OK, here you go. At [https://problem-package-format-copy.readthedocs.io/en/latest/core.html#problem-settings](https://problem-package-format-copy.readthedocs.io/en/latest/core.html#problem-settings) I include the file `increment/problem.yaml`, which is an example (toy but complete) problem in https://github.com/thorehusfeldt/problem_package_format_copy/tree/main/problems/increment . Theoretically, `increment` can be validated by `verifyproblem`. All of its files can be included in the documentation.

We should also have a scoring problem under `problems` and use that to illustrate the more advanced part of the specification.

niemela:
@thorehusfeldt How correct are this schemas still (considering all the changes that have happened since the PR was created)?

I think we all agree that having schemas is an obvious good. So my question is only should we merge this now, either because it's perfectly in line with the spec as it stands right now, or because it's close enough and we can fix it in a follow up PR?

thorehusfeldt:
The schemas in my own for are up-to-date, as far as I’ve been able to follow.

[https://github.com/thorehusfeldt/problem_package_format_copy/tree/main/support](https://github.com/thorehusfeldt/problem_package_format_copy/tree/main/support)

Moreover, the script at [https://github.com/thorehusfeldt/problem_package_format_copy/blob/main/test.sh](https://github.com/thorehusfeldt/problem_package_format_copy/blob/main/test.sh) runs every yaml file in my fork (in particular all the yaml files appearing in the documentation) against those schemas, to the documentation and the schema specification are _in agreement_. 

At BAPCtools, we’ve taking this to one more level; the `generators.yaml` framework not only as a validating CUE schema, but
1. the script that checks internal validity is run by Github’s  continuous integration framework, so validity is ensured on commit
2. we have translated the CUE schema also into a JSON schema
3. the JSON schema (and a full test suite) is published at the schema repository schemastore.org, so for every user on the planet that uses a compliant  editor (such as VS Code), there is now out-of-the-box syntax checking, autocompletion, and hints for writing problem package generators without any configuration necessary. It feels like complete science-fiction. (In VIM and Emacs you have to set this up, but it takes a few minutes.)

We can easily do something similar for `problem.yaml` and `testdata.yaml`.

niemela:
> The schemas in my own fork are up-to-date, as far as I’ve been able to follow.

Are you saying that this PR is, not not?

thorehusfeldt:
Requires some updates. I’ll close this PR and make a new one on a feature branch.

PR #49, opened 2023-07-22
timing changes

evouga:
Per our discussion @niemela @eldering

### DIFF
--- a/spec/2023-07-draft/problem_package_format.md
+++ b/spec/2023-07-draft/problem_package_format.md
@@ -173,8 +173,8 @@ the "typical system default" is what is needed.
 
 | Key                  | Comments             | Default        |
 |----------------------|----------------------|----------------|
-| accepted_to_limit    | optional float       | 5.0            |
-| limit_to_exceeded    | optional float       | 1.4            |
+| accepted_to_limit    | optional float       | 2.0            |
+| limit_to_exceeded    | optional float       | 1.5            |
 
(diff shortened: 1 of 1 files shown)


pehrsoderman (approved):

niemela (approved):

simonlindholm:
Not totally happy about this, it makes it more verbose to specify time_multiplier, which is the most common thing to want to do given the current workflow.

I wonder if there should be a helpful note recommending specifying time multiplier instead of time limit.

eldering:
> Not totally happy about this, it makes it more verbose to specify time_multiplier, which is the most common thing to want to do given the current workflow.
> 
> I wonder if there should be a helpful note recommending specifying time multiplier instead of time limit.

Note that there are still always time multipliers (possibly the default ones) that will be used to check that an explicitly chosen time limit satisfies the constraints.

eldering (approved):

simonlindholm:
> Note that there are still always time multipliers (possibly the default ones) that will be used to check that an explicitly chosen time limit satisfies the constraints.

Yep, my only worry is if people try to use the property to set a higher-than-default time limit, and the judge then turns out to be slower than the local machine, but not slow enough to fail the time multiplier checks. The warning note looks good.

> Tweak time limit multiplier defaults
>
> These defaults should be such that they provide some sensible
safety margins, but on the other hand are not too high so that
one would have to modify them almost for any problem and then
lead to people getting used to always tweaking them.

FWIW my experience has been that 5x is fine, and I wouldn't be surprised if the changed default leads to specifying 5 explicitly just as often as we currently specify 2 explicitly. Or that we sometimes become stricter in our time limits than we are currently out of laziness, which I'm not super happy about either. I don't think there's any default that would make custom time multipliers significantly less common, there's such a large variation in whether accepted solutions include Python solutions or C++ ones, fast solutions or slow ones, etc.

Issue #36, opened 2023-07-21
Add support for 2-pass (or multipass) problems


niemela:

jsannemo:
**Decision**: yes. PR made.

PR #50, opened 2023-07-22
Add multipass problems.

jsannemo:


pehrsoderman:
Two questions here:
* How are time limits on multipass problems calculated?
*  Should there be any guarantees regarding state for the program (for example, should files written to disk be preserved, which seems like a very natural use case)?

jsannemo:
1. Max of runtime over all passes.

jsannemo:
2. It's not necessary I think. The validator is responsible for writing the output, so it can append any data to be passed to the next stage. This simplifies things a lot too, since it doesn't change the judging flow too much.

pehrsoderman:
Okay, you may have to give me a use case for how this is intended to be used then. I thought it was intended for a two stage problem, where the program is first given some kind of initialization data to process (for example a map or a dictionary of words), and in a second stage it will be required to use the built up state to quickly answer queries regarding the data in the first stage.

As described, I am not quite sure how it differs from an interactive problem, other than in measuring the time for each interaction.

jsannemo:
https://noi17.kattis.com/contests/vfoqp8/problems/avian

jsannemo:
To elaborate now that we're back from lunch: the submission in a multipass problems must be stateless between passes. This is a property orthogonal to interactivity.

eldering:
> To elaborate now that we're back from lunch: the submission in a multipass problems must be stateless between passes. This is a property orthogonal to interactivity.

That sounds sensible. So I guess it should then be a requirement on the CCS that the submissions is rerun in a clean environment where any previously written files are not available anymore, right?

spec/2023-07-draft/output_validators.md
```diff
@@ -128,6 +127,16 @@ Note that a validator may choose to ignore the feedback directory
 entirely. In particular, the contest control system must not assume that
 the validator program creates any files there at all.
 
+<s class="kattis">
+### Multi-pass validation
+
+A multi-pass validator can be used for problems that should run the submission multiple times sequentially, using a new input set generated by the previous invocation of the submission.
```
eldering, line 133: `generated` sounds like the output of the previous run of the submission is literally used as input, but below it becomes clear that it's the output of the validator.

spec/2023-07-draft/output_validators.md
```diff
@@ -151,3 +160,4 @@ Almost all test cases failed, are you even trying to solve the problem?
 A valididator program is allowed to write any kind of debug information
 to its standard error pipe. This information may be displayed to the
 user upon invocation of the validator.
+
```
eldering, line 163: Nitpick: why add a line here?

jsannemo:
> > To elaborate now that we're back from lunch: the submission in a multipass problems must be stateless between passes. This is a property orthogonal to interactivity.
> 
> That sounds sensible. So I guess it should then be a requirement on the CCS that the submissions is rerun in a clean environment where any previously written files are not available anymore, right?

Yes

niemela:
Closes #36

spec/2023-07-draft/output_validators.md
jsannemo: Fixed

spec/2023-07-draft/output_validators.md
jsannemo: Reworded

PR #69, opened 2023-07-22
.args and .files support

evouga:
@niemela

### DIFF
--- a/spec/2023-07-draft/problem_package_format.md
+++ b/spec/2023-07-draft/problem_package_format.md
@@ -236,8 +236,23 @@ statement and sample test data are provided in the directory `attachments/`.
 The test data are provided in subdirectories of `data/`. The sample data in
 `data/sample/` and the secret data in `data/secret/`.
 
-All input and answer files have the filename extension `.in` and `.ans`
-respectively.
+All files and folders associated to a single test case have the same base name with varying extensions. The table below summarizes the supported test data:
+
+| Extension                       | Described In         | Summmary                               |
+|---------------------------------|----------------------|----------------------------------------|
+| `.in`                           | Input                | Input piped to standard input          |
+| `.ans`                          |                      | Expected (AC) answer                   |
+| `.hint`                         | Annotations          | Hint for solving the test case         |
+| `.desc`                         | Annotations          | Purpose of the test                    |
+| `.png`, `.jpg`, `.jpeg`, `.svg` | Annotations          | Illustration of the test case          |
+| `.interaction`                  | Interactive Problems | Interaction protocol sample            |
+| `.args`                         | Input                | Input passed as command-line arguments |
+| `.files`                        | Input                | Input available via file I/O           | 
+
+### Input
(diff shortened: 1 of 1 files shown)


niemela (approved):

evouga:
I didn't say anything about visibility of the (possibly changed) .files to the output validator; presumably that will be handled under the issue on working directories.

evouga:
Per the group discussion this feature has been nixed for now.

niemela:
> Per the group discussion this feature has been nixed for now.

Awaiting more feedback.

spec/2023-07-draft/problem_package_format.md
```diff
@@ -236,8 +236,23 @@ statement and sample test data are provided in the directory `attachments/`.
...
+| `.png`, `.jpg`, `.jpeg`, `.svg` | Annotations          | Illustration of the test case          |
+| `.interaction`                  | Interactive Problems | Interaction protocol sample            |
+| `.args`                         | Input                | Input passed as command-line arguments |
+| `.files`                        | Input                | Input available via file I/O           | 
+
+### Input
+Each test case can supply input via standard input, command-line arguments, and/or the file system. These options are not exclusive. For a test case with base name `test`, the file `test.in` (if it exists) is piped to the submission as standard input. The submission's filename is passed as the first command-line argument, followed by the whitespace-separated tokens in `test.args` (if this file exists).
+
+`test.files`, if it exists, must be a folder, and contains privileged data files available to the submission via file I/O. All files in this folder are copied into the submission's working directory after compiling but before executing the submission (possibly overwriting the compiled submission file or included data in the case of name conflicts).
```
pehrsoderman, line 255: It is probably a good idea to specify how ownership and file mask should be handled.

pehrsoderman (approved):

spec/2023-07-draft/problem_package_format.md
jsannemo, line 255: > into the submission's working directory

This is a bit ugly... As the text notes,
> possibly overwriting the compiled submission file
is somewhat iffy. At least in my judge I make several assumptions about the submission directory being immutable, so that it can be reused between tasks and still assume that the submission will be runnable.

This is not entirely non-ugly either, but I think I would prefer if we put the files in some known, reserved folder name relative to the CWD. That would also allow a very natural follow-up in making that folder writable, and having it shared with the output validator.

I think it also would make very much sense for the input files to be available to a custom output validator somehow too.

jsannemo (commented):
I'm slightly skeptical to `.args` and where the `.files` should be placed, but after thinking a bunch I do like the concept of per-case included files now.

spec/2023-07-draft/problem_package_format.md
jsannemo: Don't be too Linux-centric here eihter.

niemela:
There are a lot of conflicts. I will create a new PR with the conflicts resolved.

spec/2023-07-draft/problem_package_format.md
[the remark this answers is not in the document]
niemela: How about adding some generic "anything may be tarred, but when untarred must result in the file structure defined" applying to everything in the spec?

niemela:
I have created #117, with the exact same contents as this PR, and will merge it now.

Issue #48, opened 2023-07-22
Problem name in problem.yaml


niemela:
Should `name` actually be required?

We can (and do) get the name from `\problemname` in LaTeX. We could do something similar with Markdown.

How should we handle when the problem name is specified in both problem statement and `problem.yaml`?

niemela:
@pehrsoderman @austrin @ghamerly

pehrsoderman:
I think it makes sense to keep it, to have a canonical name for the problem. If you have multiple language statements for the problem they could have different names. Also, it is easier to parse the yaml with tooling, compared to the latex or markdown.

RagnarGrootKoerkamp:
What we do in BAPCtools:
- Specify the name in `name:` in `problem.yaml` as a string.
- In `problem.en.tex` we do `\problemname{\problemyamlname}`, where `\problemyamlname` is the string name as specified in the yaml.
- In case of multiple natural languages, `\problemyamlname` in `problem.xy.tex` is replaced by the value of `xy` in the `name:` dictionary.
- In case you want a dedicated latex name, simply put that as `\problemname{Problem $i$}`.

How do you feel about adding this to the spec?

RagnarGrootKoerkamp:
Alternatively, we could reuse the constants and have a 'magic constant' `XproblemnameX` that is replaced by the string (for the current language) specified in `problem.yaml`

jsannemo:
I like the magic constant, it's very cross-language

RagnarGrootKoerkamp:
New idea:
`\problemname` (or `\problemname{}`) renders the name as in the problem.yaml, and `\problemnam{\LaTeX name}` can be used to have a specific name that includes latex/math.

niemela:
> New idea:
> `\problemname` (or `\problemname{}`) renders the name as in the problem.yaml, and `\problemname{\LaTeX name}` can be used to have a specific name that includes latex/math.

I think this is the way to go (and specifcially `\problemname{}`). This also allows systems that want to have a fancy LaTeX name to support that by parsing the LaTeX or Markdown. We should (maybe?) add a requirement that the name provided in `problem.yaml` should be an acceptable fallback to whatever is specified in `\problemname`.

There are only ~4 problemnames on Kattis that tries to use "fancy LaTeX" in the problem name:
- So you want to be a $2^n$-aire?
- Perfect $p$th Powers
- Perfect $k$-ary Tree
- $S$-Nim

Based on this, I think this would be fine.

Issue #31, opened 2023-07-21
Problem statements needs to be better specified


niemela:
What exact LaTeX environment can/must be assumed.

niemela:
@austrin Could you help with this one?

jsannemo:
I think we should define a "HTML-friendly" subset, that implmentations must promise to be able to convert to HTML.

It should still be allowed to go outside of this subset. A CCS for example compile that LaTeX using a provided problem class name, and instead host a PDF upon installation.

evouga:
This seems *incredibly* daunting... a starting point might be the subset supported by mathjax, which is [quite extensive](https://onemathematicalcat.org//MathJaxDocumentation/TeXSyntax.htm).

In fact maybe the reasonable path here is to pick the same subset as a common TeX Javascript display engine like mathjax (which strongly encourages, but doesn't formally require, a CCS to use that engine).

niemela:
This is what Polygon supports: https://polygon.codeforces.com/docs/statements-tex-manual

niemela:
In addition to what Polygon supports we should probably support:
```
\begin{figure}
\begin{verbatim}
\begin{quote}
\begin{array}
\begin{minipage}
\begin{wrapfigure}
\section
\textwidth
\ldots
\noindent
```

RagnarGrootKoerkamp:
This is the class file we use for BAPC
https://github.com/RagnarGrootKoerkamp/BAPCtools/blob/master/latex/bapc.cls
Apart from that, we only have a light wrapper around problems
https://github.com/RagnarGrootKoerkamp/BAPCtools/blob/master/latex/problem.tex

Note that we also inject `\problemlabel` and `\problemauthor` that are mostly used in solution slides rather than statements.

PR #76, opened 2023-07-23
LaTeX problem statement spec

evouga:
As discussed at https://github.com/Kattis/problem-package-format/issues/31 I've started with Polygon's commands and added a few commonly-used commands from the Kattis database. Also attempted to document problem-package-specific commands the LaTeX class must provide.

### DIFF
--- a/spec/2023-07-draft/problem_package_format.md
+++ b/spec/2023-07-draft/problem_package_format.md
@@ -245,7 +245,7 @@ The LaTeX class shall provide the convenience environments `Input`, `Output`, an
  
  - `\illustration{width}{filename}{caption}`, a convenience command for adding a figure to the problem statement. `width` is a floating-point argument specifying the width of the figure, as a fraction of the total width of the problem statement; `filename` is the image to display and `caption` the text to include below the figure. The illustration should be flushed right with text flowing around it (as in a `wrapfigure`).
 
-Arbitrary LaTeX is not guaranteed to render correctly by HTML-based contest systems. However, contest systems must make a best effort to correctly render **at minimum** the following LaTeX subset when displaying a LaTeX problem statement:
+Arbitrary LaTeX is not guaranteed to render correctly by HTML-based judging systems. However, judging systems must make a best effort to correctly render **at minimum** the following LaTeX subset when displaying a LaTeX problem statement:
 
 1. All [MathJax-supported TeX commands](https://docs.mathjax.org/en/latest/input/tex/macros/index.html) within inline (`$ $`) and display (`$$ $$`) math mode.
 
--- a/spec/2023-07-draft/problem_package_format.md
+++ b/spec/2023-07-draft/problem_package_format.md
@@ -237,7 +237,7 @@ sample data. It is the contest system's responsibility to append the sample data
 
 ### LaTeX Environment and Supported Subset
 
-Problem statements provided in LaTeX must consist only of the problem statement body (i.e., the content that would be placed within a `document` environment). It is the contest system's responsibility to wrap this text in an appropriate LaTeX class.
+Problem statements provided in LaTeX must consist only of the problem statement body (i.e., the content that would be placed within a `document` environment). It is the judging system's responsibility to wrap this text in an appropriate LaTeX class.
 
 The LaTeX class shall provide the convenience environments `Input`, `Output`, and `Interaction` for delineating sections of the problem statement. It shall also provide the following commands:
 
--- a/spec/2023-07-draft/problem_package_format.md
+++ b/spec/2023-07-draft/problem_package_format.md
@@ -241,7 +241,7 @@ Problem statements provided in LaTeX must consist only of the problem statement 
 
 The LaTeX class shall provide the convenience environments `Input`, `Output`, and `Interaction` for delineating sections of the problem statement. It shall also provide the following commands:
 
- - `\problemname[name]`, which must be the first line of the problem statement. `name` is an optional argument giving a LaTeX-formatted problem name to be used in place of the `name` value from `problem.yaml` when rendering the problem statement header.
+ - `\problemname{name}`, which must be the first line of the problem statement. `name` gives a LaTeX-formatted problem name to be used when rendering the problem statement header. This argument can be empty, in which case the `name` value from `problem.yaml` is used instead.
  
  - `\illustration{width}{filename}{caption}`, a convenience command for adding a figure to the problem statement. `width` is a floating-point argument specifying the width of the figure, as a fraction of the total width of the problem statement; `filename` is the image to display and `caption` the text to include below the figure. The illustration should be flushed right with text flowing around it (as in a `wrapfigure`).
 
--- a/spec/2023-07-draft/problem_package_format.md
+++ b/spec/2023-07-draft/problem_package_format.md
@@ -232,16 +232,16 @@ reference auxiliary files as if the working directory is
 `<short_name>/problem_statement/`. Image file formats supported are `.png`,
 `.jpg`, `.jpeg`, and `.pdf`.
 
-The problem statements must only contain the actual problem statement, no
-sample data.
+For problem statements provided in LaTeX or Markdown, the problem statements must only contain the actual problem statement and no
+sample data. It is the contest system's responsibility to append the sample data, problem name and ID, time limit, and other contest headers/footers/logos. For problem statement PDFs, the problem statement is displayed unmodified and must include this information.
 
 ### LaTeX Environment and Supported Subset
 
-Problem statements provided in LaTeX must consist only of the problem statement body (i.e., the content that would be placed within a `document` environment). It is the contest system's responsibility to wrap this text in an appropriate LaTeX class and to append the sample data, problem name and ID, time limit, and other contest headers/footers/logos.
+Problem statements provided in LaTeX must consist only of the problem statement body (i.e., the content that would be placed within a `document` environment). It is the contest system's responsibility to wrap this text in an appropriate LaTeX class.
 
 The LaTeX class shall provide the convenience environments `Input`, `Output`, and `Interaction` for delineating sections of the problem statement. It shall also provide the following commands:
 
- - `\problemname{name}`, which must be the first line of the problem statement. `name` gives a LaTeX-formatted problem name to be used when rendering the problem statement header. This argument can be empty, in which case the `name` value from `problem.yaml` is used instead.
+ - `\problemname{name}`, which must be the first line of the problem statement. `name` gives a LaTeX-formatted problem name to be used when rendering the problem statement header. This argument can be empty, in which case the `name` value matching the problem statement's language from `problem.yaml` is used instead.
  
  - `\illustration{width}{filename}{caption}`, a convenience command for adding a figure to the problem statement. `width` is a floating-point argument specifying the width of the figure, as a fraction of the total width of the problem statement; `filename` is the image to display and `caption` the text to include below the figure. The illustration should be flushed right with text flowing around it (as in a `wrapfigure`).
 
--- a/spec/2023-07-draft/problem_package_format.md
+++ b/spec/2023-07-draft/problem_package_format.md
@@ -257,7 +257,7 @@ Arbitrary LaTeX is not guaranteed to render correctly by HTML-based judging syst
 
 5. The following font weight and size modifiers:  `\bf`, `\textbf`, `\it`, `\textit`, `\t`, `\tt`, `\texttt`, `\emph`, `\underline`, `\sout`, `\textsc`, `\tiny`, `\scriptsize`, `\small`, `\normalsize`, `\large`, `\Large`, `\LARGE`, `\huge`, `\Huge`.
 
-6. `\includegraphics` (from the package `graphicx`), including the Polygon-style workaround for scaling the image using `\def \htmlPixelsInCm`.
+6. `\includegraphics` from the package `graphicx`, including the Polygon-style workaround for scaling the image using `\def \htmlPixelsInCm`.
 
 7. The miscellaneous commands `\url`, `\href`, `\section`, `\subsection`, and `\epigraph`.
 
--- a/spec/2023-07-draft/problem_package_format.md
+++ b/spec/2023-07-draft/problem_package_format.md
@@ -232,14 +232,35 @@ reference auxiliary files as if the working directory is
 `<short_name>/problem_statement/`. Image file formats supported are `.png`,
 `.jpg`, `.jpeg`, and `.pdf`.
 
-
-A LaTeX file may include the Problem name using the LaTeX command
-`\problemname` in case LaTeX formatting of the title is wanted. <span
-class="not-icpc">If it's not included the problem name specified in
-`problem.yaml` will be used.</span>
-
 The problem statements must only contain the actual problem statement, no
 sample data.
+
+### LaTeX Environment and Supported Subset
+
+Problem statements provided in LaTeX must consist only of the problem statement body (i.e., the content that would be placed within a `document` environment). It is the contest system's responsibility to wrap this text in an appropriate LaTeX class and to append the sample data, problem name and ID, time limit, and other contest headers/footers/logos.
+
+The LaTeX class shall provide the convenience environments `Input`, `Output`, and `Interaction` for delineating sections of the problem statement. It shall also provide the following commands:
+
+ - `\problemname[name]`, which must be the first line of the problem statement. `name` is an optional argument giving a LaTeX-formatted problem name to be used in place of the `name` value from `problem.yaml` when rendering the problem statement header.
+ 
+ - `\illustration{width}{filename}{caption}`, a convenience command for adding a figure to the problem statement. `width` is a floating-point argument specifying the width of the figure, as a fraction of the total width of the problem statement; `filename` is the image to display and `caption` the text to include below the figure. The illustration should be flushed right with text flowing around it (as in a `wrapfigure`).
+
+Arbitrary LaTeX is not guaranteed to render correctly by HTML-based contest systems. However, contest systems must make a best effort to correctly render **at minimum** the following LaTeX subset when displaying a LaTeX problem statement:
+
+1. All [MathJax-supported TeX commands](https://docs.mathjax.org/en/latest/input/tex/macros/index.html) within inline (`$ $`) and display (`$$ $$`) math mode.
+
+2. The following text-mode environments: `itemize`, `enumerate`, `lstlisting`, `verbatim`, `quote`, `center`, `tabular`, `figure`, `wrapfigure` (from the `wrapfig` package).
+
+3. `\item` within list environments and `\hline`, `\cline`, `\multirow`, `\multicol` within tables.
+
+4. The following typesetting constructs: smart quotes  (`' '`, `<< >>`, ` `` '' `), dashes (`--`, `---`), non-breaking space (`~`), ellipses (`\ldots` and `\textellipsis`), and `\noindent`.
+
+5. The following font weight and size modifiers:  `\bf`, `\textbf`, `\it`, `\textit`, `\t`, `\tt`, `\texttt`, `\emph`, `\underline`, `\sout`, `\textsc`, `\tiny`, `\scriptsize`, `\small`, `\normalsize`, `\large`, `\Large`, `\LARGE`, `\huge`, `\Huge`.
+
+6. `\includegraphics` (from the package `graphicx`), including the Polygon-style workaround for scaling the image using `\def \htmlPixelsInCm`.
+
+7. The miscellaneous commands `\url`, `\href`, `\section`, `\subsection`, and `\epigraph`.
+
 
 ## Attachments
 


evouga:
@niemela @RagnarGrootKoerkamp

RagnarGrootKoerkamp:
Fixes #48: We decided to use `\problemname{}` (ie with an empty argument) to reuse the name from the `problem.yaml` for the current language.

RagnarGrootKoerkamp (approved):

spec/2023-07-draft/problem_package_format.md
```diff
@@ -232,14 +232,35 @@ reference auxiliary files as if the working directory is
 `<short_name>/problem_statement/`. Image file formats supported are `.png`,
 `.jpg`, `.jpeg`, and `.pdf`.
 
+For problem statements provided in LaTeX or Markdown, the problem statements must only contain the actual problem statement and no
+sample data. It is the contest system's responsibility to append the sample data, problem name and ID, time limit, and other contest headers/footers/logos. For problem statement PDFs, the problem statement is displayed unmodified and must include this information.
```
pehrsoderman, line 236: Weird line break

pehrsoderman (approved):

niemela (approved):

spec/2023-07-draft/problem_package_format.md
jsannemo, line 236: Problems are not necessarily used for contests, so "contest system" here and later is a very ICPC-y term.

spec/2023-07-draft/problem_package_format.md
```diff
@@ -232,14 +232,35 @@ reference auxiliary files as if the working directory is
...
-`\problemname` in case LaTeX formatting of the title is wanted. <span
-class="not-icpc">If it's not included the problem name specified in
-`problem.yaml` will be used.</span>
+### LaTeX Environment and Supported Subset
+
+Problem statements provided in LaTeX must consist only of the problem statement body (i.e., the content that would be placed within a `document` environment). It is the contest system's responsibility to wrap this text in an appropriate LaTeX class.
+
+The LaTeX class shall provide the convenience environments `Input`, `Output`, and `Interaction` for delineating sections of the problem statement. It shall also provide the following commands:
+
+ - `\problemname{name}`, which must be the first line of the problem statement. `name` gives a LaTeX-formatted problem name to be used when rendering the problem statement header. This argument can be empty, in which case the `name` value matching the problem statement's language from `problem.yaml` is used instead.
```
jsannemo, line 244: Does this mean I always have to include `\problemname{}` at the start of a statement when I don't want to use a name different from the one in problem.yaml? That seems weird?

spec/2023-07-draft/problem_package_format.md
```diff
@@ -232,14 +232,35 @@ reference auxiliary files as if the working directory is
...
+ 
+ - `\illustration{width}{filename}{caption}`, a convenience command for adding a figure to the problem statement. `width` is a floating-point argument specifying the width of the figure, as a fraction of the total width of the problem statement; `filename` is the image to display and `caption` the text to include below the figure. The illustration should be flushed right with text flowing around it (as in a `wrapfigure`).
```
jsannemo, line 246: I understand that this is used in plenty of problems because Kattis' problem2pdf environment supports this. I'm not sure that's a *good* idea for this to exist though.

I don't like that this subset should preferrably not put constraints upon implementers to provide this non-standard-LaTeX command just because of Kattis legacy reasons.

jsannemo (changes requested):
I think this is overall a good subset. Not too large, not too small.

My main concern is about how useful this will be in practice, contra the cost of adding a bunch of stuff to the spec.

The status quo today is that Kattis is the only system (AFAIK) that performs any LaTeX->HTML conversion, and thus the only system that benefits from authors restricting themselves (especially since we support a wider subset than polygon -- if they implement the difference, that's another story). This is not entirely true, since my own judge also performs this conversion and thus I would love if this subset became as widely used as possible to reduce work for me.

Here, I wonder what Kattis is going to do. @niemela If they drop its much, much richer subset for this, Kattis problem authors are basically losing out on formatting possibilities for no practical benefits assuming other judges don't actually care about this subset. If Kattis is not dropping it, have we just invented an arbitrary subset of LaTeX that nobody actually will use?

There is a remaining possible win: if problem authors *not targeting Kattis* restrict themselves to this subset. Then e.g. GCPC problems targeting Domjudge will actually be fully portable (so that both Kattis and me can "just install them"). The crux is that I think this in practice will only happen if the "default" tooling (i.e. problemtools & BAPCtools) is *at the very least* opinionated about this subset being recommended, and warning authors that their problem might not be installable on judges using HTML statements if they go outside of it. Is this something that problemtools and BAPCtools plan to do? @RagnarGrootKoerkamp @pehrsoderman 

**Tl;dr:**
- as a judging system implementer I love this: I want other authors to write statements I can turn to HTML
- as a problem author I love this: I can make my statements very portable between judges if I want to
- as a spec writer I would like some info about buy-in from the other parties to avoid unused bloat

spec/2023-07-draft/problem_package_format.md
niemela: What term do you suggest to use instead?

spec/2023-07-draft/problem_package_format.md
niemela: The fact that it's used a lot (which it is) is a reasonable argument for having something like this. This specific way of doing it seems reasonable. Do you have an alternate suggestion if not this?

RagnarGrootKoerkamp:
I can make a pass over my archives and see if they compile with a much reduced header that matches the above (removing most `\usepackage`). Most 'default latex' that remains and is actually used should fall in the mathjax category I think

spec/2023-07-draft/problem_package_format.md
jsannemo: I try to go with "judging system" somewhat inconsistently.

spec/2023-07-draft/problem_package_format.md
RagnarGrootKoerkamp: This is the thing that renders the header, and if we don't require this we'd have to either grep the file and manually prepend it if there's no match, or otherwise fiddle with this in latex somehow which I'm not sure how to do. Also, this nicely signals a new problem starts here and a title should be rendered.

niemela:
> Here, I wonder what Kattis is going to do. @niemela If they drop its much, much richer subset for this, Kattis problem authors are basically losing out on formatting possibilities for no practical benefits assuming other judges don't actually care about this subset. If Kattis is not dropping it, have we just invented an arbitrary subset of LaTeX that nobody actually will use?

I would really like it if we could make this subset a little bit larger and richer. It's a bit of an balancing act, of course. We want it as small as possible, but no smaller, I guess :).

spec/2023-07-draft/problem_package_format.md
niemela: I would be fine with this t.b.h.

spec/2023-07-draft/problem_package_format.md
RagnarGrootKoerkamp: We provide it in the bapctools class file as well. I like it a lot since its a command with clear intent and the right amount of customization, rather than manually having to figure out the right set of `figure`, `infludegraphics`, and `caption` commands each time you want to include a figure.

jsannemo: Isch? This command specifically is mostly used because of copy-pasting effects, starting out with a well-intended idea in the old problemset.cls.

I assume the same can be replaced with a wrapfig+include-graphics combination, with something like
```latex
\begin{wrapfigure}[R}
\includegraphics[width=*width*]{*filename*}
\small{*caption*}
\end{wrapfigure}
```
which doesn't seem too burdensome IMO.

jsannemo:
> > Here, I wonder what Kattis is going to do. @niemela If they drop its much, much richer subset for this, Kattis problem authors are basically losing out on formatting possibilities for no practical benefits assuming other judges don't actually care about this subset. If Kattis is not dropping it, have we just invented an arbitrary subset of LaTeX that nobody actually will use?
> 
> I would really like it if we could make this subset a little bit larger and richer. It's a bit of an balancing act, of course. We want it as small as possible, but no smaller, I guess :).

Was the answer "no, we won't drop things from the Kattis supported tex environment"? 

I mean, that is a completely fair thing to do. But if the above subset won't actually be that valuable in practice because there won't be any other consumers of LaTeX, other than people actually running problem2pdf, bapctools people, people running pdflatex themselves, should we just not restrict ourselves to this subset, but rather just take whatever subset of packages that Kattis supports right now and say: "this is what is supported"? That provides some value too: domjudge targeted statements can be verified to work on Kattis, and Kattis problem authors don't need to trial-and-error to know what the right environment is (need for a Kattis latex env documentation remains for as long as you support more than whatever subset is in this spec anyway IMO).

spec/2023-07-draft/problem_package_format.md
jsannemo: For problems of the old version, doing this replacement could even be part of the legacy->2023/07 converter script.

spec/2023-07-draft/problem_package_format.md
jsannemo: Personally I would rather scrap supporting latex problemnames then. Could @niemela give a few examples of where this is used today?

spec/2023-07-draft/problem_package_format.md
RagnarGrootKoerkamp: But why drop a nice and concise command? I think it's well established beyond just legacy

spec/2023-07-draft/problem_package_format.md
niemela: I listed *all* the ones I know of in [#48](https://github.com/Kattis/problem-package-format/issues/48#issuecomment-1646879460):
- So you want to be a $2^n$-aire?
- Perfect $p$th Powers
- Perfect $k$-ary Tree
- $S$-Nim

RagnarGrootKoerkamp: Removing this completely is also ok for me. I just don't like making it optional without having a clean implementation.

There are some examples in the corresponding issue.

spec/2023-07-draft/problem_package_format.md
jsannemo: > rather than manually having to figure out the right set of figure, infludegraphics, and caption commands each time you want to include a figure.

You have to do this every time you include a figure already though... something I would expect people do a lot more often than include an illustration.

I mean go ahead if you both really want this, I just think it's unnecessary to add something that's not even such a big win and as a consequence make the subset non-plain-LaTeX. That I *do* think there is a lot of value in for people who actually make PDFs of statements when grabbing random problem packages, who will be required to use tooling or define *this specific command* themselves (this is also a reason why I would like to scrap \problemname)

spec/2023-07-draft/problem_package_format.md
jsannemo: No I agree with you that optional is bad. Let's remove it then?

jsannemo: Or maybe I misunderstood what "this" you both were Ok with

niemela: > No I agree with you that optional is bad. 

Agree, optional is the worst option.

> Let's remove it then?

I'd lean towards keeping. @austrin @ghamerly @pehrsoderman Thoughts?

jsannemo: A compormise would be to include the string `\def\latexproblemname{your latex problemname}` in the statement for this.

evouga:
Re: the `\problemname` command: so is the consensus that this command is optional, and if provided, overrides the problem name? Or to remove it completely? I have no strong opinions here; I don't think it's too constraining to require plainly-formatted problem titles.

Re: the `\illustration` command: here I agree with Ragnar that this command is a big win. We want to make it as easy as possible for author to write high-quality problem statements that look good *both* in PDF and HTML. This is not entirely trivial when dealing with figures; compare for instance the example posted by @jsannemo to the actual implementation in BAPCtools which includes quite a bit of carefully-tuned spacing and alignment tweaks:
```
\newcommand{\illustration}[3]{
  \begin{wrapfigure}{r}{#1\textwidth}
	\vspace{-10pt}
    \includegraphics[width=#1\textwidth]{#2}
	\ifstrempty{#3}{
		\vspace{-25pt}
	}{
		\begin{flushright}
		\vspace{-9pt}
		\tiny #3
		\end{flushright}
		\vspace{-15pt}
	}
  \end{wrapfigure}
}
```
We don't want to require (or trust) authors to tune and test these things themselves when the system can provide reasonable defaults.

Re: warning on use of unsupported TeX: I think this is a great idea in theory, not least because even in my experience of judging Kattis-run contests, authors have historically used TeX commands that do not render correctly even in Kattis's HTML engine. In practice it may be difficult to enforce, other than by removing unsupported `\usepackage`s from the class file. Are we going to `grep` each TeX file for some keyword blacklist that we have to maintain ourselves...?

Re: expanding the supported LaTeX subset: no strong opinions here; go ahead and add anything you think is important. I do want to caution that if we hope to Polygon to allow importing problem-packages in the future, it's probably a good idea not to stray too far from their current supported subset.

niemela:
> In practice it may be difficult to enforce, other than by removing unsupported `\usepackage`s from the class file.

I think we should probably do that. But also really consider if there are more things we should add to the supported set.

I.e. if there is something we would feel *really* bad about removing, we should probably add it to the supported set instead.

> if we hope to Polygon to allow importing problem-packages in the future, it's probably a good idea not to stray too far from their current supported subset.

They have said that they are mot interested in supporting importing, only exporting. That said, I'm kinda hoping that can change in the future.

spec/2023-07-draft/problem_package_format.md
eldering: I agree with @jsannemo: try to keep the implementation as clean as possible. The `wrapfigure` is standard LaTeX; if we think that authors can't write this, then it signals that maybe we should switch to a different source format for problem texts? (That's maybe a bit tongue-in-cheek, but also somewhat serious.)

spec/2023-07-draft/problem_package_format.md
```diff
@@ -232,14 +232,35 @@ reference auxiliary files as if the working directory is
...
 
-A LaTeX file may include the Problem name using the LaTeX command
-`\problemname` in case LaTeX formatting of the title is wanted. <span
-class="not-icpc">If it's not included the problem name specified in
-`problem.yaml` will be used.</span>
+### LaTeX Environment and Supported Subset
+
+Problem statements provided in LaTeX must consist only of the problem statement body (i.e., the content that would be placed within a `document` environment). It is the contest system's responsibility to wrap this text in an appropriate LaTeX class.
+
+The LaTeX class shall provide the convenience environments `Input`, `Output`, and `Interaction` for delineating sections of the problem statement. It shall also provide the following commands:
```
eldering, line 242: All LaTeX commands/environments tend to be fully lower case. Why deviate from that here? Also, I think that if it is said that these must be provided, then at least something should be said how they should work (in terms of usage and display).

evouga: > standard LaTeX; if we think that authors can't write this, then it signals that maybe we should switch to a different source format for problem texts?

Both LaTeX and HTML were designed with the philosophy of divorcing content from presentation. Authors are supposed to write TeX without worrying about exactly where each equation will appear on the page etc. and leave that to automated tooling. Unfortunately standard TeX is designed for math equation and includes no support for figures. (Very clunky) figure support was retrofitted with packages like `wrapfig`.

Of course authors can figure out how to use `wrapfig` and may even be able to create a figure that looks great on exactly the one judging system they're writing for. If you want it to also look good when imported into Polygon, DOMjudge, etc. you need to give them *platform-independent* tools for creating a good-looking figure. This means a command that hides all of the implementation-dependent spacing and formatting needed to make a figure look good on Judge System X and instead requires the authors to specify only the content.

evouga: As for changing to a different problem statement language, I have no objections, except that (1) TeX is already widely-known and used on e.g. Polygon and (2) I'm not aware of a better alternative with great support for both math and figures. Maybe .docx comes closest.

spec/2023-07-draft/problem_package_format.md
niemela: "If they need this QoL improvement then we should drop LaTeX support" is a very much false choice. And yes, it was a *bit* tongue-in-cheek. More importantly though, this seems to only take future authors into account.

I think there are 4 things we should mainly consider when making choices regarding this:
1. Existing problem packages.
2. Ease of use for future problem package authors.
3. Ease of import from other formats.
4. Ease of export to other formats.

They are not necessarily equally important (I would say that 1 & 2 are the most important), and maybe 3 & 4 should be merged?

If we want to do thing that are bad for one or several of the things above, there are ways to counter it:
1. If we write a conversion script (that works), then point 1 doesn't really matter anymore. So anything that can be converted by a script should be fine.
2. If we write good documentation/tutorial/examples the easy of use increases (a lot) even if we pick a somewhat convoluted solution.
3. When there exist stable import tools then it becomes a bit less important that it was easy to create them.
4. Same as 3.

RagnarGrootKoerkamp:
Slightly related discussion on allowing explanations in between samples:
https://github.com/Kattis/problemtools/issues/21

@mzuenni suggests maybe we can allow `\sample{basename}` to include a sample into the latex. That's pretty simple to do from the latex side, but maybe not great for portability. (I'd say that as soon as you use `\sample{..}` once, no further samples are appended automatically anymore.)

spec/2023-07-draft/problem_package_format.md
eldering: Ok, I hadn't seen the tweaks implemented in BAPCtools to render the wrapfigure nicely, so fair enough. I guess then it would also be good to add something like that code as a reference for how this `\illustration` command could be implemented.
Should this specification then maybe contain a reference `.cls` file as an attachment?

niemela: > Should this specification then maybe contain a reference `.cls` file as an attachment?

This is what I thought we would have to do. I don't know if that a reasonable thing to do though?

RagnarGrootKoerkamp: Not sure we need to provide a reference implementation. Every tool can decide for itself how it likes to render 'a figure with width `x%` (of `\textwidth`) and caption `y`'

niemela: I agree that we don't "need" to, but if we don't we need to explain what is expected to be implemented by this. So providing a `.cls` coudl/would/should be less work? But it is also overspecifying a bit, which is not great.

I would say that "no specification" < "underspecified" < "overspecified" < "perfectly specified" :)

RagnarGrootKoerkamp: right sure; providing an example may be easier indeed, but (obviously) we should clearly mark is as example.

spec/2023-07-draft/problem_package_format.md
Tagl, line 236: Also is ID referring to shortname or UUID?
```suggestion
sample data. It is the judging system's responsibility to append the sample data, problem name and ID, time limit, and other contest headers/footers/logos. For problem statement PDFs, the problem statement is displayed unmodified and must include this information.
```

spec/2023-07-draft/problem_package_format.md
```diff
@@ -232,14 +232,35 @@ reference auxiliary files as if the working directory is
...
+For problem statements provided in LaTeX or Markdown, the problem statements must only contain the actual problem statement and no
+sample data. It is the contest system's responsibility to append the sample data, problem name and ID, time limit, and other contest headers/footers/logos. For problem statement PDFs, the problem statement is displayed unmodified and must include this information.
 
-A LaTeX file may include the Problem name using the LaTeX command
-`\problemname` in case LaTeX formatting of the title is wanted. <span
-class="not-icpc">If it's not included the problem name specified in
-`problem.yaml` will be used.</span>
+### LaTeX Environment and Supported Subset
+
+Problem statements provided in LaTeX must consist only of the problem statement body (i.e., the content that would be placed within a `document` environment). It is the contest system's responsibility to wrap this text in an appropriate LaTeX class.
```
Tagl, line 240:
```suggestion
Problem statements provided in LaTeX must consist only of the problem statement body (i.e., the content that would be placed within a `document` environment). It is the judging system's responsibility to wrap this text in an appropriate LaTeX class.
```

spec/2023-07-draft/problem_package_format.md
```diff
@@ -232,14 +232,35 @@ reference auxiliary files as if the working directory is
...
+
+The LaTeX class shall provide the convenience environments `Input`, `Output`, and `Interaction` for delineating sections of the problem statement. It shall also provide the following commands:
+
+ - `\problemname{name}`, which must be the first line of the problem statement. `name` gives a LaTeX-formatted problem name to be used when rendering the problem statement header. This argument can be empty, in which case the `name` value matching the problem statement's language from `problem.yaml` is used instead.
+ 
+ - `\illustration{width}{filename}{caption}`, a convenience command for adding a figure to the problem statement. `width` is a floating-point argument specifying the width of the figure, as a fraction of the total width of the problem statement; `filename` is the image to display and `caption` the text to include below the figure. The illustration should be flushed right with text flowing around it (as in a `wrapfigure`).
+
+Arbitrary LaTeX is not guaranteed to render correctly by HTML-based contest systems. However, contest systems must make a best effort to correctly render **at minimum** the following LaTeX subset when displaying a LaTeX problem statement:
```
Tagl, line 248:
```suggestion
Arbitrary LaTeX is not guaranteed to render correctly by HTML-based judging systems. However, judging systems must make a best effort to correctly render **at minimum** the following LaTeX subset when displaying a LaTeX problem statement:
```

spec/2023-07-draft/problem_package_format.md
```diff
@@ -232,14 +232,35 @@ reference auxiliary files as if the working directory is
...
+
+1. All [MathJax-supported TeX commands](https://docs.mathjax.org/en/latest/input/tex/macros/index.html) within inline (`$ $`) and display (`$$ $$`) math mode.
+
+2. The following text-mode environments: `itemize`, `enumerate`, `lstlisting`, `verbatim`, `quote`, `center`, `tabular`, `figure`, `wrapfigure` (from the `wrapfig` package).
```
Tagl, line 252:
```suggestion
2. The following text-mode environments from the `wrapfig` package: `itemize`, `enumerate`, `lstlisting`, `verbatim`, `quote`, `center`, `tabular`, `figure`, `wrapfigure`.
```

spec/2023-07-draft/problem_package_format.md
```diff
@@ -232,14 +232,35 @@ reference auxiliary files as if the working directory is
...
+
+3. `\item` within list environments and `\hline`, `\cline`, `\multirow`, `\multicol` within tables.
+
+4. The following typesetting constructs: smart quotes  (`' '`, `<< >>`, ` `` '' `), dashes (`--`, `---`), non-breaking space (`~`), ellipses (`\ldots` and `\textellipsis`), and `\noindent`.
```
Tagl, line 256: Does this include \dotsb and all the rest? Vertical and diagonal and such?

spec/2023-07-draft/problem_package_format.md
```diff
@@ -232,14 +232,35 @@ reference auxiliary files as if the working directory is
...
+
+5. The following font weight and size modifiers:  `\bf`, `\textbf`, `\it`, `\textit`, `\t`, `\tt`, `\texttt`, `\emph`, `\underline`, `\sout`, `\textsc`, `\tiny`, `\scriptsize`, `\small`, `\normalsize`, `\large`, `\Large`, `\LARGE`, `\huge`, `\Huge`.
+
+6. `\includegraphics` (from the package `graphicx`), including the Polygon-style workaround for scaling the image using `\def \htmlPixelsInCm`.
```
Tagl, line 260: 😄 
```suggestion
6. `\includegraphics` from the package `graphicx`, including the Polygon-style workaround for scaling the image using `\def \htmlPixelsInCm`.
```

spec/2023-07-draft/problem_package_format.md
niemela: ID here is certainly not referring to the UUID.

A follow up question though; is this meant to be read as a *requirement* in the judging system that it MUST append "sample data, problem name and ID, time limit, and other contest headers/footers/logos"? 

- "sample data" is absolutely required, you must be able to assume that the sample data will be added.
- "problem name" should likely be required, because that should be there (right?).
- "ID" seems to be an implementation detail? Why would it even be necessarily needed? It certainly could in some systems, but the problem will in many (but not all) use cases already be uniquely identified by the problem name. If the ID is not needed in a particular context I see no reason that we require it to be put there.
- "other contest headers/footers/logos" is clearly very vague. *Requiring* it does not really mean anything; a system that does not want to add it would just have to make sure to have not way to provide them. And if you have a way to provide them, surely you will be adding them, that is the only point of them after all?
- "time limit"? I don't know. It's definitely there in most use cases, but does it *have* to be? And why only the time limit, why not the memory limit, or the other limits? 

It seems to me that we should absolutely require "sample data" and probably "problem name", but all the others should be up to the system. If this is the case we kinda need to clarify this sentence.

spec/2023-07-draft/problem_package_format.md
niemela: Is it really correct to say that all of those are "from the `wrapfig` package"?

niemela: @Tagl I don't know if you were just out killing parenthesis, but this seems to change the meaning? Was that intended?

Tagl: Oh I just I misunderstood it then, I thought that was what it was saying. You can ignore this suggestion.

spec/2023-07-draft/problem_package_format.md
Tagl: Yeah I don't see a reason for requiring ID (shortname) from being displayed.

I guess "other contest headers/footers/logos" must mean attachments? What else is there?

I think all limits *should* be displayed in the judge, not just time and memory.
Not sure I'd go so far as to say that they *must*, but I think it's pretty important information to convey.

niemela: > I guess "other contest headers/footers/logos" must mean attachments? What else is there?

No, I'm quite certain this is referring to the contest specific headers and footers that are typically added to problem statements when used in a contest. These are not included in the problem package (nor should they).

> I think all limits _should_ be displayed in the judge, not just time and memory. Not sure I'd go so far as to say that they _must_, but I think it's pretty important information to convey.

I agree.

Tagl: > No, I'm quite certain this is referring to the contest specific headers and footers that are typically added to problem statements when used in a contest. These are not included in the problem package (nor should they).

In that case I don't think it is relevant here at all and can be skipped, or just state that we don't forbid systems from injecting things into the statements such as headers/footers.

niemela:
> I can make a pass over my archives and see if they compile with a much reduced header that matches the above (removing most `\usepackage`). Most 'default latex' that remains and is actually used should fall in the mathjax category I think

What were the results from this?

niemela:
I have created #120 resolving the conflicts in this PR.

spec/2023-07-draft/problem_package_format.md
niemela: I removed ", problem name and ID, time limit, and other contest headers/footers/logos." in #120.

RagnarGrootKoerkamp: Oh, I was reading this sentence more as a guidance to 'do not include time limit and headers in the problem.en.tex. The contest system will do this for you.'

But anyway it's also fine without.

PR #83, opened 2023-07-24
Cleanup legacy (and rewrite history a bit)

niemela:
Cleaning up the legacy version a bit. Also making some after the fact choices for what was in the `legacy` version. Updating the change log to match.


spec/legacy/problem_package_format.md
```diff
@@ -196,18 +196,18 @@ The problem statement of the problem is provided in the directory
...
 This directory must contain one file per language, for at least one
-language, named `problem.`<language>`.`<filetype>, that contains the
+language, named `problem.`\<language\>`.`\<filetype\>, that contains the
 problem text itself, including input and output specifications, but not
 sample input and output. Language must be given as the
 shortest ISO 639 code. If needed a hyphen and a ISO 3166-1 alpha-2 code
 may be appended to ISO 639 code. Optionally, the language code can
 be left out, the default is then English (`en`). Filetype can be either
-`tex` for LaTeX files<s>, `md` for Markdown,</s><s class='problemarchive'><s> or `pdf` for PDF</s></s>.
+`tex` for LaTeX files, `md` for Markdown, or `pdf` for PDF.
```
RagnarGrootKoerkamp: should `md` and `pdf` be `not-icpc`?

spec/legacy/problem_package_format.md
```diff
@@ -125,7 +125,7 @@ stated. Any unknown keys should be treated as an error.
...
 | - | - | - | - |
 | name | String |   | The name of the problem. |
-| <s class="kattis"> type                                      </s> | <s class="kattis"> String                               </s> | <s class="kattis"> pass-fail                                               </s> | <s class="kattis"> One of "pass-fail" and "scoring".                                                                                                                                                                                                                                                                                                              </s> |
+| <span class="not-icpc">type</span> | <span class="not-icpc">String</span> | <span class="not-icpc">pass-fail</span> | <span class="kattis">One of `pass-fail` and `scoring`.</span> |
```
RagnarGrootKoerkamp, line 119: Also replace the last `kattis` by `not-icpc`

spec/legacy/problem_package_format.md
```diff
@@ -419,8 +346,7 @@ to standard output.
 ## Input Validators
 
 Input Validators, for verifying the correctness of the input files, are
-provided in <s class="dep kattis">`input_format_validators/`</s>
```
simonlindholm, line 422: Old problems use this, don't remove it

spec/legacy/problem_package_format.md
```diff
@@ -308,14 +308,12 @@ group will be used. The format of `testdata.yaml` is as follows:
...
-| output\_validator<s class="dep kattis">\_flags</s></s> |  String or map with the keys "name" and "flags"</s> |  empty string  | arguments passed to the output validator for this test data group.</s> If a string this is the name of the output validator that will be used for this test data group. If a map then this is the name as well as the flags that will be passed to the output validator.</s>                    |
+| input\_validator\_flags |  String or map with the keys "name" and "flags" |  empty string  | arguments passed to the input validator for this test data group. If a string this is the name of the input validator that will be used for this test data group. If a map then this is the name as well as the flags that will be passed to the input validator.                      |
```
RagnarGrootKoerkamp, line 302: `arguments passed to *each* input validator for this test data group. If a map, then each key must be the name of an input validator, and the corresponding value are the flags.`

(Drop the part about 'string means name of validator'; that was never true.)

spec/legacy/problem_package_format.md
```diff
@@ -308,14 +308,12 @@ group will be used. The format of `testdata.yaml` is as follows:
...
+| output\_validator\_flags |  String or map with the keys "name" and "flags" |  empty string  | arguments passed to the output validator for this test data group. If a string this is the name of the output validator that will be used for this test data group. If a map then this is the name as well as the flags that will be passed to the output validator.                    |
```
RagnarGrootKoerkamp, line 303: Same here:
`arguments passed to *each* output validator for this test data group. If a map, then each key must be the name of an output validator, and the corresponding value are the flags.`

spec/legacy/problem_package_format.md
RagnarGrootKoerkamp: In that case maybe mention both `input_format_validators` and `input_validators` as names?

spec/legacy/problem_package_format.md
```diff
@@ -39,7 +39,7 @@ sequences defined by IEEE 754-2008 and may use up to double precision.
 
 There are a number of different kinds of programs that may be provided
 in the problem package; submissions, input validators, output
-validators<s class="kattis">, graders and generators</s>.
+validators<span class="not-icpc">, graders and generators</span>.
```
RagnarGrootKoerkamp, line 42: remove the generators here

spec/legacy/problem_package_format.md
niemela: Fixed.

spec/legacy/problem_package_format.md
niemela: I added back a mention of this.

spec/legacy/problem_package_format.md
niemela: I think it was. I'd like to leave it in for now, if we can confirm that it never was we can remove it later. Ok?

niemela: Same as above.

spec/legacy/problem_package_format.md
niemela: Same.

spec/legacy/problem_package_format.md
niemela: Fixed.

RagnarGrootKoerkamp (approved):

niemela:
Will look into the question (from @RagnarGrootKoerkamp) about "Drop the part about 'string means name of validator'; that was never true in practice, in my understanding." Not sure what exactly was actually supported by the different tools.

Will merge this now, and come back and fix that later if needed.

PR #86, opened 2023-07-24
Update examples to latest draft version

RagnarGrootKoerkamp:
`problem.yaml`:
- add required `name`
- add version
- add uuid
- add new time limit settings

directory:
- less indentation
- clarify that `submissions/*` is currently about the final verdict
- rename `output_validators` to `output_validator`

### DIFF
--- a/examples/directory_structure.md
+++ b/examples/directory_structure.md
@@ -1,43 +1,43 @@
----
-permalink: /examples/directory_structure
----
 # Directory structure
 
 ```text
 <short_name>/
-      problem.yaml - problem configuration file
-      problem_statement/
-              problem.tex - problem statement
-              - any files that problem.tex needs to include, e.g. images
-      data/
-              sample/
-                      *.in - sample input files
-                      *.ans - sample answer files
-              secret/
-                      *.in - input files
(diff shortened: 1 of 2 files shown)


niemela:
Don't we need to have one of these per version?

RagnarGrootKoerkamp:
Yeah not quite sure. Users should always use the latest version for new problems IMO, so not having examples for old versions discourages that.

But we could add examples for older versions if people feel like it.

niemela:
> Users should always use the latest version for new problems IMO, so not having examples for old versions discourages that.

That's a really good point. So let's start with having good examples for the current version then. (i.e. what you're doing).

What do you think about having `2023-07-draft` either in the file name or at the top of the file so that it's very clear that the example is specifically for that version?

RagnarGrootKoerkamp:
> What do you think about having `2023-07-draft` either in the file name or at the top of the file so that it's very clear that the example is specifically for that version?

yes done. Mentioned at top of directory example, and added  `problem_format_version: 2023-07-draft` also to the minimal `problem.yaml`.

niemela (approved):

PR #89, opened 2023-07-24
add text explaining variants at the top

### LINKED ISSUE #79: Render difference between `-icpc` and full specification
Currently there is no way to see which parts of the specification pertain to the ICPC subset short of reading HTML tags in the source. (Sometime, a pair of buttons in the left-hand margin enables this, but it’s fragile, implicit, and not obvious to outsiders.)

RagnarGrootKoerkamp:
Fix #79

@thorehusfeldt


niemela:
This was basically included in #91 .

PR #92, opened 2023-07-24
legacy: cleanup of input_validator_flags

### LINKED ISSUE #90: legacy: cleanup of `input_validator_flags`
Following up on https://github.com/Kattis/problem-package-format/pull/83#issuecomment-1648425327

As described in #26, ``input_validator_flags: <string>` was specified to be the name of the input validator to run.

The actual behaviour in `problemtools` seem to be to run all validators with the given string as flags.

I implemented some preliminary support for this at some point in `BAPCtools` that actually follows the spec, but I don't think that code was ever used. (BAPCtools doesn't properly support testgroups anyway at the moment.)

Can we fix the legacy spec to say that a string argument is a flag passed to all validators?

niemela:
Fixes #90


spec/changelog.md
```diff
@@ -27,6 +27,7 @@ sort: 1
 - Add `uuid` to `problem.yaml`.
 - Add `languages` to `problem.yaml`.
 - Add support for Markdown problem statements.
+- Change format for specifying multiple input validators with flags using `input_validator_flags` in `testdata.yaml`.
```
RagnarGrootKoerkamp, line 30: This is already mentioned above. But also, now that we have 'rewritten history' to never have had this 'string means validator name to run' in the first place, the point `update {input,output}_validator_flags` is kinda void.
I think we should keep here:
- only allow a single output validator (this implies a corresponding change of `ouput_validator_flags` to not allow maps)
- remove `validator_flags` from `problem.yaml`

and then optionally in the legacy changelog put something like:
- Change meaning of string argument to `{input,output}_validator_flags` from name of validator (never implemented) to flags to pass.

spec/legacy.md
```diff
@@ -266,7 +266,7 @@ The format of `testdata.yaml` is as follows:
...
-| input_validator_flags  | String or map with the keys "name" and "flags" | empty string | arguments passed to the input validator for this test data group. If a string this is the name of the input validator that will be used for this test data group. If a map then this is the name as well as the flags that will be passed to the input validator.                                                 |
+| input_validator_flags  | String or map with the keys "name" and "flags" | empty string | arguments passed to the input validator for this test data group. If a string then those are the flags that will be passed to each input validator for this test data group. If a map then this is the name as well as the flags that will be passed to the input validator.                                                 |
```
RagnarGrootKoerkamp, line 269: I think you can copy the text from the current draft spec here.

spec/legacy.md
```diff
@@ -266,7 +266,7 @@ The format of `testdata.yaml` is as follows:
...
 | output_validator_flags | String or map with the keys "name" and "flags" | empty string | arguments passed to the output validator for this test data group. If a string this is the name of the output validator that will be used for this test data group. If a map then this is the name as well as the flags that will be passed to the output validator.                                              |
```
RagnarGrootKoerkamp, line 270: This also needs changing. You can copy the `input_validator_flags` from the current draft and change input to output. (`output_validator_flags` in the new version has been changed to never take a map, which is not the case in the legacy version.)

Tagl (approved):

spec/changelog.md
niemela: I did the first changes and skipped the last...

spec/legacy.md
niemela: Done.

spec/legacy.md
niemela: Not sure what needs to be done here. Could you make a PR? Will merge this as is...

niemela:
Ok... that merge conflict was confusing... I think I got it right...

PR #101, opened 2023-07-29
Simplify scoring, as per previous discussion

jsannemo:


jsannemo:
The change for sample from `min` to `sum` is to let `sum` be more of a hint to systems to judge all the samples (which is a good thing here) rather than fail at first non-success (which `min` implies is "more allowed").

spec/2023-07-draft.md
```diff
@@ -530,23 +530,26 @@ where test cases are run in lexicographical order of their full file paths (note
 
 <div class="not-icpc">
 
-For scoring problems, the behaviour is configured by the following flags under `grading` in `testdata.yaml`:
+For scoring problems, the concept of verdict is less defined.
+In fact, some judging systems may choose to not even present a final verdict for a submission, instead only present for example the score and the verdicts on those test cases the submission failed.
```
niemela, line 534: Maybe nitpicking?:

I'm not sure the concept is "less defined". Maybe "less important" is more correct? The fact that some systems might choose to not even present a verdict seems to indicate that it's less important, right?

I certainly can see some ways that it is also "less defined" (or at least "less generally agreed upon how to exactly define it", which is not exactly the same thing.

spec/2023-07-draft.md
```diff
@@ -530,23 +530,26 @@ where test cases are run in lexicographical order of their full file paths (note
...
-- The _sample_ group: `score` is 0, `scoring` is `min`, `verdict` is `first_error`.
-- Other groups: score is `1`, `scoring` is `min`, verdict is `first_error`.
+- The _data_ and _secret_ group: `score` is 1, `aggregation` is sum.
+- The _sample_ group: `score` is 0, `scoring` is `sum`.
+- Other groups: score is `1`, `scoring` is `min`.
 - `max_score` is the sum of the `max_score` of the subresults if `scoring` is sum, and the minimum of the `max_score` of the subresults of `scoring` is `min`. For individual testcases, the `max_score` here means the `score` value of the group it is in.
 
-If `verdict` is `first_error`, only test cases in the group up to the first non-accepted case are judged.
-If `verdict` is `accept_if_any_accepted`, all test cases in the group are judged.
+They are chosen to minimize the configuration needed for a typical scoring pproblem that might appear in olympiad-style contests such as the International Olympiad in Informatics.
```
niemela, line 551:
```suggestion
They are chosen to minimize the configuration needed for a typical scoring problem that might appear in olympiad-style contests such as the International Olympiad in Informatics.
```

niemela, line 551: More nitpicking: I don't love the term "olympiad-style contests".

niemela:
> The change for sample from `min` to `sum` is to let `sum` be more of a hint to systems to judge all the samples (which is a good thing here) rather than fail at first non-success (which `min` implies is "more allowed"). Of course both behaviors are allowed with the defaults, since score is 0.

Oh... I think I love this!

Let me think a little bit though...

spec/2023-07-draft.md
jsannemo: Happy to take suggestions

spec/2023-07-draft.md
jsannemo: I think I meant to say *the verdict of a submission*.

niemela: > I think I meant to say _a verdict of a submission_.

Ok, yes, that would be better, but I felt that part was implied. I still think it's more "less important" than "less defined", in addition to the change above.

jsannemo: No objection, updated.

spec/2023-07-draft.md
niemela: Yeah, I know, it's easier to complain than to think of something better 😞 . I will think about it though...

RagnarGrootKoerkamp: Just write `defaults are typical for IOI-style problems`?

RagnarGrootKoerkamp:
This is fine with me, but I don't use this anyway.

jsannemo:
@RagnarGrootKoerkamp well, Thore might try to force it into BAPCtools so it can still affect you :)

RagnarGrootKoerkamp:
I'm fine with supporting whatever you think is reasonable for scoring problems :)

jsannemo:
ACK, then I'll pop you off as a reviewer!

spec/2023-07-draft.md
```diff
@@ -530,23 +530,26 @@ where test cases are run in lexicographical order of their full file paths (note
 
 <div class="not-icpc">
 
-For scoring problems, the behaviour is configured by the following flags under `grading` in `testdata.yaml`:
+For scoring problems, the concept of verdict is less important.
+In fact, some judging systems may choose to not even present a final verdict for a submission, instead only present for example the score and the verdicts on those test cases the submission failed.
```
Tagl, line 534:
```suggestion
In fact, some judging systems may choose to not even present a final verdict for a submission, but instead only present, for example, the score and the verdicts on the test cases the submission failed.
```

spec/2023-07-draft.md
```diff
@@ -530,23 +530,26 @@ where test cases are run in lexicographical order of their full file paths (note
...
+
+The scoring behaviour is configured by the following flags under `grading` in `testdata.yaml`:
```
Tagl, line 536: Someone claimed we're using American English, so we must write this *incorrectly*
```suggestion
The scoring behavior is configured by the following flags under `grading` in `testdata.yaml`:
```

spec/2023-07-draft.md
```diff
@@ -530,23 +530,26 @@ where test cases are run in lexicographical order of their full file paths (note
...
+| `max_score` | String                                  | A number specifying the maximum score allowed for this test group. It is an error to exceed this.                                                                                                                 |
+| `aggregation` | `sum` or `min`                          | If sum, the score is the sum of the subresult scores. If min, the score is the minimum of the subresult scores.                                                                                                   |
 
 The defaults are as follows:
 
-- The _data_ and _secret_ group: `score` is 1, `scoring` is sum, `verdict` is `accept_if_any_accepted`.
-- The _sample_ group: `score` is 0, `scoring` is `min`, `verdict` is `first_error`.
-- Other groups: score is `1`, `scoring` is `min`, verdict is `first_error`.
+- The _data_ and _secret_ group: `score` is 1, `aggregation` is sum.
+- The _sample_ group: `score` is 0, `scoring` is `sum`.
```
Tagl, line 547:
```suggestion
- The _sample_ group: `score` is 0, `aggregation` is `sum`.
```

spec/2023-07-draft.md
```diff
@@ -530,23 +530,26 @@ where test cases are run in lexicographical order of their full file paths (note
...
+- Other groups: score is `1`, `scoring` is `min`.
 - `max_score` is the sum of the `max_score` of the subresults if `scoring` is sum, and the minimum of the `max_score` of the subresults of `scoring` is `min`. For individual testcases, the `max_score` here means the `score` value of the group it is in.
 
-If `verdict` is `first_error`, only test cases in the group up to the first non-accepted case are judged.
-If `verdict` is `accept_if_any_accepted`, all test cases in the group are judged.
+They are chosen to minimize the configuration needed for a typical scoring problem that might appear in olympiad-style contests such as the International Olympiad in Informatics.
+There, points are given for passing all test cases in groups of testcases called subtasks.
+To achieve that behaviour, create test data groups `secret/subtask1`, `secret/subtask2`, ... and set the `score` value in the `secret/subtaskX/testdata.yamlsecret/`.
```
Tagl, line 553:
```suggestion
To achieve that behaviour, create test data groups `secret/subtask1`, `secret/subtask2`, ... and set the `score` value in the `secret/subtaskX/testdata.yaml`.
```

Tagl (approved):

spec/2023-07-draft.md
jsannemo: @niemela would you think that is OK? I avoided "IOI-style" to avoid the unexplained acronym "IOI", but maybe that was unnecessary

niemela: or `defaults are typical for problems in the style of the International Olympiad in Informatics`? I do agree that unexplained acronyms are bad. OTOH if you don't know what IOI is, than the sentence wont make any sense anyway.

RagnarGrootKoerkamp: We pretty much assume familiarity with ICPC anyway; it'll be fine

jsannemo: I think I agree with @RagnarGrootKoerkamp : pushed some changed wording.

niemela: > We pretty much assume familiarity with ICPC anyway; 

Why does "we assume familiarity with ICPC" imply that assuming familiarity with IOI is ok?

> it'll be fine

Yeah, I think so as well.

spec/2023-07-draft.md
```diff
@@ -530,23 +530,25 @@ where test cases are run in lexicographical order of their full file paths (note
...
-- Other groups: score is `1`, `scoring` is `min`, verdict is `first_error`.
+- The _data_ and _secret_ group: `score` is 1, `aggregation` is sum.
+- The _sample_ group: `score` is 0, `aggregation` is `sum`.
+- Other groups: score is `1`, `scoring` is `min`.
 - `max_score` is the sum of the `max_score` of the subresults if `scoring` is sum, and the minimum of the `max_score` of the subresults of `scoring` is `min`. For individual testcases, the `max_score` here means the `score` value of the group it is in.
 
-If `verdict` is `first_error`, only test cases in the group up to the first non-accepted case are judged.
-If `verdict` is `accept_if_any_accepted`, all test cases in the group are judged.
+They are chosen to minimize the configuration needed for a typical IOI-style problem, where points are given for passing all test cases in test data groups called subtasks.
+To achieve that behaviour, create groups `secret/subtask1`, `secret/subtask2`, ... and set the `score` value in the `secret/subtaskX/testdata.yaml`.
```
niemela, line 552: We should probably expand on common use cases somewhere. (That is not an argument against this change, just a thought...)

niemela (approved):

spec/2023-07-draft.md
jsannemo: I was that we can do that after Thore's restructuring. I kept it very brief as a placeholder reminder to do that.

PR #104, opened 2023-07-30
Minor editorial clarifications.

jsannemo:


spec/2023-07-draft.md
```diff
@@ -278,7 +278,7 @@ If a specific order is desired a numbered prefix such as `00`, `01`, `02`, `03`,
 
 <div class="not-icpc">
 
-In each test data group, a file `testdata.yaml` may be placed to specify how the result of the test data group should be computed.
+In each test data group, a UTF-8 encoded YAML file `testdata.yaml` may be placed to specify how the result of the test data group should be computed.
```
RagnarGrootKoerkamp, line 281: i think this is already specified at the top of the spec somewhere, that all files are UTF-8

jsannemo: This was for consistency with what we write on "problem.yaml". We specify that *text files* must be UTF-8 without a BOM, but not what files are "text files".

This is why I added the restriction that .interaction must be a text file. I'm not sure interactive problems are restricted to ASCII, but I think the .interaction has to be (giving semantic meaning to `<`, `>` and newlines is a weird thing to do in a general binary container).

RagnarGrootKoerkamp: Indeed, maybe stress that the `.interaction` should be `ASCII` only

jsannemo: Hmm, why wouldn't UTF-8 be OK?

jsannemo: Or maybe you meant that in reply to my erroneous usage of ASCII in my previous comment :)

RagnarGrootKoerkamp: Maybe utf8 is ok, but I have no idea if it is easy or hard to render it correctly, nor whether programs can parse it. But I suppose if people really want to they can use utf8

RagnarGrootKoerkamp: Do we actually specify that .in files are required to be text, and not binary?

jsannemo: No. For example, https://open.kattis.com/problems/coincounter has binary .in files.

niemela: I would rather have UTF-8 (and details similar to that) be defined in one place, as opposed to mentioned in reference to every file.

That said, that's a separate change from this, so we can do that later if we decide that we agree with me.

niemela (approved):

spec/2023-07-draft.md
simonlindholm: I would be very happy if we could require that .in files are text, binary stdin is problematic for many reasons. Especially if #69 happens and we can allow problems with binary files that way instead.

niemela: > I would be very happy if we could require that .in files are text, 

That's a fairly breaking change. We already have such problems.

> binary stdin is problematic for many reasons.

What are the main issues?

Would it help if binary `.in` were only allowed if some key is specified in `problem.yaml`? Something *like* `stdin: binary` that defaults to `stdin: text`.

jsannemo: Can this perhaps migrate to a new issue? It feels like a bigger discussion that this PR doesn't directly touch on either.

simonlindholm: Mainly two issues:
- some programming languages don't allow reading stdin as a binary file
- on Windows it requires obscure tricks to read from stdin without crlf->lf conversion, making it impossible for some people to test solutions locally

Banning the use of binary stdin would be a way of getting rid of such problems in the future. Although an opinionated problemtools warning would serve the same purpose, so we don't strictly need to ban it in the spec.

(<https://kth.kattis.com/problems/oldkattis.aes> is another problem that uses binary stdin/stdout.)

spec/2023-07-draft.md
```diff
@@ -234,14 +234,14 @@ The hint file is meant to be given as feedback, i.e. to somebody that fails to s
 The description file is a text file with filename extension `.desc` describing the purpose of an input file.
 The description file is meant to be privileged information that explains the purpose of the related test file, e.g. what cases it's supposed to test.
```
Tagl, line 235: Not your change, but here it says test file and above it says input file. We should be consistent.
```suggestion
The description file is meant to be privileged information that explains the purpose of the related input file, e.g. what cases it's supposed to test.
```

spec/2023-07-draft.md
```diff
@@ -234,14 +234,14 @@ The hint file is meant to be given as feedback, i.e. to somebody that fails to s
...
 
-The Illustration is an image file with filename extension `.png`, `.jpg`, `.jpeg`, or `.svg`.
+The Illustration is an image file with filename extension `.png`, `.jpg`, `.jpeg`, or `.svg` (at most one such image file may exist).
```
Tagl, line 237: I think this is an important detail.
```suggestion
The Illustration is an image file with filename extension `.png`, `.jpg`, `.jpeg`, or `.svg`.
Note that at most one such image file may exist for each input file.
```

Tagl (approved):

jsannemo:
I'm merging this; we can keep going about the discussion on the text file-iness of inputs elsewhere.

spec/2023-07-draft.md
niemela: > Can this perhaps migrate to a new issue? It feels like a bigger discussion that this PR doesn't directly touch on either.

Yes it should.

Also because this PR has already been merged.

PR #107, opened 2023-07-31
Remove validation_output

jsannemo:


jsannemo:
I don't understand the purpose of this key, so I am putting forward an aggressive proposal to remove it.

@niemela @eldering could you explain if this is a meaningful limit to have?

jsannemo:
Should I take the silence as no meaningful purpose exists? :)

niemela:
It has been used (and not only to set it to the suggested default). The most common value seems to be "1".

I honestly don't really understand why it's important though.

@pehrsoderman Thoughts?

eldering:
Similar to Fredrik: I'm also on holidays, so not able to keep up with all comments on all the PRs in this repo. I also don't see a use case for this, where it should ever be different from `output`, but also not sure...

niemela:
@jsannemo Was there a reason why you only suggested to remove `validation_output` and not also `validation_time` and `validation_memory`? I definitely agree that `validation_output` seems to be the least useful of the three, but honestly, they are all a bit questionable.

I ***think*** that the idea for the `validation_*` limits is to be able to specify when validation is expected to use more resources than what is normal. I don't understand why you would want to specify a lower limit. I.e. why would you want to say that the *submission* should fail if the validator uses too much resources? And if there truly are cases where you need that, the validator *could* do that check, it would not have to be set in `limits`.

The usage I have seen (i.e. what we have in the Kattis archive) is:
- `validation_time`:
  - ~30 instances of `240`
  - ~30 instances of `60` (the suggested default)
  - ~5 insances of `<= 5`
- `validation_memory`
  - ~25 instances of `2048` (the suggested default)
  - ~40 instances of `1024`
- `validation_output` 
  - ~10 instances of `8` (the suggested default)
  - ~50 instances of `1`

The only ones of these that would be "actually useful" ***if*** we buy my reasoning above would be the ~30 instances of `validation_time = 240`.

Tagl:
> I don't understand why you would want to specify a lower limit.

QoL for the people implementing the problem, they trust the judging system to respect the limits instead of making their own custom checks for these things. I mean it isn't that difficult, but it certainly easier with these options. That being said, I have never personally modified these limits, and never even considered modifying `validation_output` since my validators don't output anything except temporary debug information.

niemela:
> QoL for the people implementing the problem, they trust the judging system to respect the limits instead of making their own custom checks for these things.

I still don't (really) get it though. If the validator fails due to any of these limits that would be a Judge Error (right?), so it's not even useful for judging submissions.

OTOH, it is basically an `assert()` in the validator, i.e. "this should never happen, if it does report as Judge Error", and asserts are not useless.

niemela:
> @jsannemo Was there a reason why you only suggested to remove `validation_output` and not also `validation_time` and `validation_memory`? I definitely agree that `validation_output` seems to be the least useful of the three, but honestly, they are all a bit questionable.
> 
> I _**think**_ that the idea for the `validation_*` limits is to be able to specify when validation is expected to use more resources than what is normal. I don't understand why you would want to specify a lower limit. I.e. why would you want to say that the _submission_ should fail if the validator uses too much resources? And if there truly are cases where you need that, the validator _could_ do that check, it would not have to be set in `limits`.
> 
> The usage I have seen (i.e. what we have in the Kattis archive) is:
> 
>     * `validation_time`:
>       
>       * ~30 instances of `240`
>       * ~30 instances of `60` (the suggested default)
>       * ~5 insances of `<= 5`
> 
>     * `validation_memory`
>       
>       * ~25 instances of `2048` (the suggested default)
>       * ~40 instances of `1024`
> 
>     * `validation_output`
>       
>       * ~10 instances of `8` (the suggested default)
>       * ~50 instances of `1`
> 
> 
> The only ones of these that would be "actually useful" _**if**_ we buy my reasoning above would be the ~30 instances of `validation_time = 240`.

@jsannemo ^

Tagl:
What about validation memory in the context of input validators? Perhaps the default is not sufficient for the input validator so you want to specify a higher limit.

pehrsoderman:
I think all the validation limits are very useful for a system. It gives an upper limit for the amount of resources a problem will require for validation, and it also allows for problems that require additional resources for validation to signal this. I would prefer to keep all of them.

niemela:
Closing this PR.

PR #108, opened 2023-07-31
Add reasonable limits to limits.

jsannemo:


spec/2023-07-draft.md
```diff
@@ -178,10 +178,14 @@ A map with the following keys:
 | validation_memory  | optional, in MiB           | system default | 2048                   |
 | validation_output  | optional, in MiB           | system default | 8                      |
 
+All of these limits must be positive.
+`output` may not be smaller than the largest default answer file in the problem, unless a custom output validator is used.
```
Tagl, line 182: Looks kind of weird to start the sentence with `output`.
```suggestion
Additionally, `output` may not be smaller than the largest default answer file in the problem, unless a custom output validator is used.
```

spec/2023-07-draft.md
```diff
@@ -191,6 +195,8 @@ Only specify limits when the problem needs a specific limit, but do specify limi
 | ac_to_time_limit  | float    | 2.0     |
 | time_limit_to_tle | float    | 1.5     |
 
+Both of these limits must be at least 1.
```
RagnarGrootKoerkamp, line 198:
```suggestion
Both of these multipliers must be at least 1.
```

RagnarGrootKoerkamp (approved):

spec/2023-07-draft.md
Tagl, line 182: Do we want to clarify the specific case where the largest file is an empty file?
Can output be 0 in that case, thus breaking the statement about it being positive above?

Tagl: I guess that usually a custom output validator is in use when that is the case. But we do have this task: https://iceland.kattis.com/problems/iceland.flytibaka

jsannemo: Sure, but you don't *need* to lower the output limit for that problem :)

jsannemo: And anyhow, the default validator accepts for example extra spacing, right?

Tagl: That's true and yes the default validator does by default, you can switch it off.

spec/2023-07-draft.md
```diff
@@ -178,10 +178,14 @@ A map with the following keys:
 | validation_memory  | optional, in MiB           | system default | 2048                   |
 | validation_output  | optional, in MiB           | system default | 8                      |
 
+All of these limits must be positive.
```
niemela, line 181: This is certainly reasonable, almost *too* reasonable.

I'm generally skeptical of specifying *too* obvious things, since that implies that you can't assume any other obvious truths that are not explicitly stated. 

Also, any problem with negative limits would *always* fail, so the problem would never validate, right?

spec/2023-07-draft.md
```diff
@@ -178,10 +178,14 @@ A map with the following keys:
...
+Additionally, `output` may not be smaller than the largest default answer file in the problem, unless a custom output validator is used.
```
niemela, line 182: Similar point here in that breaking this rule would just create a problem that can't be validated, which catches the issue.

Or, are we saying that validation might not be applying these limits (and it really should)? If so, this *does* add some value.

spec/2023-07-draft.md
```diff
@@ -191,6 +195,8 @@ Only specify limits when the problem needs a specific limit, but do specify limi
 | ac_to_time_limit  | float    | 2.0     |
 | time_limit_to_tle | float    | 1.5     |
 
+Both of these multipliers must be at least 1.
```
niemela, line 198: I could *almost* see a use case for `ac_to_time_limit` < 1. But it's probably a bad idea...

niemela (changes requested):
I'm not really requesting any changes, only a discussion... (i.e. change my mind 😃)

spec/2023-07-draft.md
jsannemo: Hmm, I guess I didn't really comsider fully whether `submissions/` are subject to all these limits? To me, it's not fully obvious they should be by default, since you develop problems on a different machine than running it, which is mostly fine for TL thanks to multipliers

But it is actually a good point, even though it's nice to be able to verify this without running things, since DomJudge probably won't test submissions at import time. I'm happy to remove it if you are strongly against it.

I general, the reason I want to specify this in your word "obvious" behaviour is because whenever my verifyproblem implementation *fails* I problem, I want to be able to point to where in the spec that is forbidden. I think that's reasonable. Having negative limits I at least think should be forbidden (in my judge I would reject those anyway).

spec/2023-07-draft.md
jsannemo: This was mostly a jumping board into another comment I made on slack: whether the default here should not be a fixed system default, but automatically be adjusted to e.g. max(system default, 1.5x the default output size). One could of course argue that a validation tool is allowed to just do that anyway (since "system defaults" aren't explicitly stated to be static. If you agree that this would be OK for a tool to do, I'm happy dropping this too.

spec/2023-07-draft.md
jsannemo: :) but these I actually think should be specified, for the reason I stated. 

Mostly because doing allowing it to be < 1 breaks what we semantically claim `accepted` solutions do -- get Accepted, so we contradict ourselves about that.

spec/2023-07-draft.md
niemela: > To me, it's not fully obvious they should be by default,

No, I agree, it's not obvious.

> I'm happy to remove it if you are strongly against it.

I'm not *strongly* against it. This is a class of mistake that I'm generally scared about, based on previous experience.

> I general, the reason I want to specify this in your word "obvious" behaviour is because whenever my verifyproblem implementation _fails_ I problem, I want to be able to point to where in the spec that is forbidden. I think that's reasonable. Having negative limits I at least think should be forbidden (in my judge I would reject those anyway).

I very much agree that we want to make is so that whenever you fail validation of a problem there is a clear thing in the spec you can point to. That said, do you need this line for that? I would argue that if a problem specifies e.g. a negative memory limit you can fail it, even without running anything, because you know that it will use a least no memory, which is more than negative memory.

spec/2023-07-draft.md
niemela: I agree that it doesn't *explicitly* say that the system default is not dynamic, but if we want to allow that (and that sounds very reasonable, doesn't it?) I think we should probably mention that?

spec/2023-07-draft.md
niemela: Yes, I agree. I did say *almost* :).

RagnarGrootKoerkamp: We could even require them `>1`: having TL equal to slowest AC doesn't make sense. Alternatively we can add that the chosen timelimit is the first multiple of resolution strictly larger than the slowest AC.

niemela: That's also a good point.

That said, I would rather bake the > into the definition than require that `ac_to_time_limit` > 1. 

I.e. instead of:
- time limit must be >= slowest AC * `ac_to_time_limit` 
- `ac_to_time_limit` must be >1

We should say:
- time limit must be > slowest AC * `ac_to_time_limit`
- `ac_to_time_limit` must be >= 1

spec/2023-07-draft.md
jsannemo: Yes, I think it makes sense to mention.

There's still a question on whether we want to leave it to tooling to do the smart thing (i.e. set output limit to by default be > the min answer), or mandate that the "system default" will be such that it holds true. Wdyt?

spec/2023-07-draft.md
jsannemo: > That said, do you need this line for that?

I guess I don't *need* it. But if the layout for this spec was a bit different (i.e., when Thore's stuff is merged), it would be very natural to bake this in as part of the data type definition: "`blah`: a positive integer that ..." vs "`blah`: an integer".

spec/2023-07-draft.md
niemela: I think we should allow the tooling to (try to) be smart.

The intent with the "defaults" was to give a lot of freedom to systems to do "reasonable" choices. Most of the limits wont be very important in most cases anyway, and you're only supposed to set a value if you explicitly need a specific value.

spec/2023-07-draft.md
niemela: > But if the layout for this spec was a bit different (i.e., when Thore's stuff is merged), it would be very natural to bake this in as part of the data type definition: "`blah`: a positive integer that ..." vs "`blah`: an integer".

That's a *very* good point.

Maybe we should leave this open until *after* we merge in Thore's stuff? I am working on that...

niemela:
Closing now, let's wordsmith further (maybe reopening this PR later)?

PR #116, opened 2023-08-15
Clarify when code limit is applied in the case of included code

### LINKED ISSUE #115: Should included code count towards code limit?
Should it?
I'm leaning towards no.
Reason being that it is code provided by the authors of the problem, so it should not impact the submitter.
There can also be a great difference between the size of provided Java code vs size of provided Python code because Java standard library has a much wider variety of built in data structures.

Tagl:
Closes #115


eldering (approved):
LGTM

niemela (approved):

PR #117, opened 2023-09-16
.args and .files support

niemela:
Replaces #69

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -247,7 +247,28 @@ Public, i.e. non-secret, files to be made available in addition to the problem s
 The test data are provided in subdirectories of `data/`.
 The sample data in `data/sample/` and the secret data in `data/secret/`.
 
-All input and answer files have the filename extension `.in` and `.ans` respectively.
+All files and folders associated to a single test case have the same base name with varying extensions. The table below summarizes the supported test data:
+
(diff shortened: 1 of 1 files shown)


jsannemo:
I feel like this PR went in a bit quickly without proper review (including that on the PR it was based on)?

In particular, whitespace separated tokens is IMO clearly the wrong way to specify args (newline-separated args makes more sense?),  and I don't think "submission's working directory after compiling but before executing the submission" is a very good formulation: I think it suggests that per-case files are not cleared between cases (especially the 'overwriting' comment), but I don't think that's what's intended?

niemela:
> I feel like this PR went in a bit quickly without proper review (including that on the PR it was based on)?

This PR was just a "rebase", the contents is the same as the old PR. The old PR was approved (but couldn't be merged due to conflicts).

> In particular, whitespace separated tokens is IMO clearly the wrong way to specify args (newline-separated args makes more sense?), 

I agree, that sounds better. Are you then also implying that the arguments should be quoted (by the system)? Single or double quotes (single seems better to me)?   

> and I don't think "submission's working directory after compiling but before executing the submission" is a very good formulation: I think it suggests that per-case files are not cleared between cases (especially the 'overwriting' comment), but I don't think that's what's intended?

Well, no, I don't think it suggests that per-case files are not cleared at all. You're correct that that was not intended though.

PR #120, opened 2023-09-16
LaTeX problem statement spec

### LINKED ISSUE #31: Problem statements needs to be better specified

### LINKED ISSUE #48: Problem name in problem.yaml

niemela:
Replaces #76 

Fixes #31 
Fixes #48 

I would suggest that we merge this fairly soon if we are reasonably in agreement, and then create issues for remaining questions.

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -260,7 +260,7 @@ Auxiliary files needed by the problem statement files must all be in `<short_nam
 Image file formats supported are `.png`, `.jpg`, `.jpeg`, and `.pdf`.
 
 For problem statements provided in LaTeX or Markdown, the problem statements must only contain the actual problem statement and no sample data.
-It is the contest system's responsibility to append the sample data, problem name and ID, time limit, and other contest headers/footers/logos.
+It is the contest system's responsibility to append the sample data.
 For problem statement PDFs, the problem statement is displayed unmodified and must include this information.
 
 ### LaTeX Environment and Supported Subset
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -259,10 +259,36 @@ Auxiliary files needed by the problem statement files must all be in `<short_nam
 `problem.<language>.<filetype>` should reference auxiliary files as if the working directory is `<short_name>/problem_statement/`.
 Image file formats supported are `.png`, `.jpg`, `.jpeg`, and `.pdf`.
 
-A LaTeX file may include the Problem name using the LaTeX command `\problemname` in case LaTeX formatting of the title is wanted.
-<span class="not-icpc">If it's not included the problem name specified in `problem.yaml` will be used.</span>
+For problem statements provided in LaTeX or Markdown, the problem statements must only contain the actual problem statement and no sample data.
+It is the contest system's responsibility to append the sample data, problem name and ID, time limit, and other contest headers/footers/logos.
+For problem statement PDFs, the problem statement is displayed unmodified and must include this information.
 
-The problem statements must only contain the actual problem statement, no sample data.
+### LaTeX Environment and Supported Subset
+
+Problem statements provided in LaTeX must consist only of the problem statement body (i.e., the content that would be placed within a `document` environment).
+It is the judging system's responsibility to wrap this text in an appropriate LaTeX class.
+
+The LaTeX class shall provide the convenience environments `Input`, `Output`, and `Interaction` for delineating sections of the problem statement.
+It shall also provide the following commands:
+
+- `\problemname{name}`, which must be the first line of the problem statement.
+  `name` gives a LaTeX-formatted problem name to be used when rendering the problem statement header.
+  This argument can be empty, in which case the `name` value matching the problem statement's language from `problem.yaml` is used instead.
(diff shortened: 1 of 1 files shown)


spec/2023-07-draft.md
```diff
@@ -259,10 +259,36 @@ Auxiliary files needed by the problem statement files must all be in `<short_nam
...
+For problem statements provided in LaTeX or Markdown, the problem statements must only contain the actual problem statement and no sample data.
+It is the contest system's responsibility to append the sample data.
+For problem statement PDFs, the problem statement is displayed unmodified and must include this information.
+
+### LaTeX Environment and Supported Subset
+
+Problem statements provided in LaTeX must consist only of the problem statement body (i.e., the content that would be placed within a `document` environment).
+It is the judging system's responsibility to wrap this text in an appropriate LaTeX class.
+
+The LaTeX class shall provide the convenience environments `Input`, `Output`, and `Interaction` for delineating sections of the problem statement.
```
niemela, line 271: As @eldering pointed out; why do we use upper case for these function names? Seems to not be the norm for LaTex?

OTOH, *if* there is a good reason, why don't we use uppercase for `\problemname`?

niemela, line 271: Should we also add a `Scoring` section for describing the scoring? This is very commonly used in practice.

spec/2023-07-draft.md
```diff
@@ -259,10 +259,36 @@ Auxiliary files needed by the problem statement files must all be in `<short_nam
...
+- `\problemname{name}`, which must be the first line of the problem statement.
+  `name` gives a LaTeX-formatted problem name to be used when rendering the problem statement header.
+  This argument can be empty, in which case the `name` value matching the problem statement's language from `problem.yaml` is used instead.
+- `\illustration{width}{filename}{caption}`, a convenience command for adding a figure to the problem statement.
+  `width` is a floating-point argument specifying the width of the figure, as a fraction of the total width of the problem statement;
+  `filename` is the image to display and `caption` the text to include below the figure.
+  The illustration should be flushed right with text flowing around it (as in a `wrapfigure`).
+
+Arbitrary LaTeX is not guaranteed to render correctly by HTML-based judging systems.
+However, judging systems must make a best effort to correctly render **at minimum** the following LaTeX subset when displaying a LaTeX problem statement:
```
niemela, line 283: Do we want to strengthen this to say that valid problem packages must not use more than the following subset.

(If we do that we might want to increase the subset a bit, as mentioned in #76, it's a balancing act)

spec/2023-07-draft.md
Tagl: Commonly used but also under different names.
I don't think it is a bad idea, but I've seen "Scoring", "Grading", "Subtasks" and more.

spec/2023-07-draft.md
Tagl: The current statement at least isn't really imposing any restrictions on either side.
It's kinda just saying both sides should try to play nice with each other.
I think I lean towards a stronger wording.

spec/2023-07-draft.md
niemela: A quick and dirty search in the Kattis archive:
- Scoring: 151 times
- Grading: 15 times
- Subtasks: 27 times

Tagl: Personally, I like Scoring, it is general enough to cover everything and that's what we call it in problem.yaml.

Tagl (approved):

spec/2023-07-draft.md
thorehusfeldt: Just for the record, I advise against this. These are functionally empty environments (just wrappers around `\section`) that add nothing, impose unwelcome constraints (“what is the orthodox name of this section?”) onto traditions (like, say, IOI) that have existing but different terminological expectations, and make it harder to translate to other languages than English.

niemela: > These are functionally empty environments (just wrappers around `\section`) 

True

> that add nothing, 

No, they allow a system to know where these sections are. Cf. how Polygon does it, instead of one big LaTeX snippet as we do, they have one snippet for each of these sections. (Which really is pretty much the same thing. IMO, doing it their way makes more sense it you have a web UI, doing it our way makes more sense if you have a file format).

> impose unwelcome constraints (“what is the orthodox name of this section?”) onto traditions (like, say, IOI) that have existing but different terminological expectations, 

I think we have enough experience now to say that this is *not* an issue. T.b.h., it might be more of an issue with "Scoring" than the others, but I don't think it would be either.

> and make it harder to translate to other languages than English.

Not really, it could take an argument, just like `\problemname`?

RagnarGrootKoerkamp: i18n works fine for these. Anyway the latex template needs to be language-aware, so adding in a few more translations for these wasn't a problem, at least for BAPCtools.

niemela:
@austrin Do you have any thoughts on this?

pehrsoderman (approved):

PR #127, opened 2023-09-20
Update specification of the include directories and how it interacts …

### LINKED ISSUE #123: Drop languages setting in problem.yaml
Should we drop the `languages` setting the `problem.yaml`?

I don't really see a use case for this, it's been used 0 times on Kattis.

Can anybody else see a need for this?

Tagl:
…with the set of allowed programming languages / submission types

Fixes #123


RagnarGrootKoerkamp:
This might be specified elsewhere, but what exactly should `<language>` be? The file extension used for the language, or maybe the key of the entry `languages.yaml` (which is outside the spec)?

Tagl:
> This might be specified elsewhere, but what exactly should `<language>` be? The file extension used for the language, or maybe the key of the entry `languages.yaml` (which is outside the spec)?

The initial values in languages.yaml is not outside the spec, the valid values are given in a table above.

niemela (approved):

niemela:
@Tagl Could you update this PR to *not* remove `languages` from `problem.yaml` (see comment in #123). I would like to merge the remaining changes though.

Tagl:
@niemela I can't because I lost push privileges and I wasn't doing this on a fork. But you can run
```
git switch included_files
git revert ed092138a5450d247f7576091c59282498de0493
git push
```
Then you remove that commit, or I can make a new PR from a different branch with the same changes.

Issue #125, opened 2023-09-19
Allow for multiple sources?


niemela:
The original expectation for the `source` field was that it would typically specify the contest that the problem was created for. Naturally, some problems were not created for contests, but rather courses, or practice series, and that's also fine.

Lately (well, *mostly* lately) it has been that case that a given problem is created for multiple contest. In the past this was mostly due to new (or weak) contests using the problem set from another contest. That meant that the problem *set* was usually used as a whole, *and* there was a reasonable choice for source, since one of the contests had the problems, and shared it with another. Even in the past though there were a few cases were problem sets were created more jointly.

In any case, in the last several years, mostly in North America, problems have been developed jointly in a "Problem Pool" and then used in different configurations in several contests that run more or less at the same time.

The solution so far has been to set `source` to something like `2022 ICPC North America Regional Programming Contests (February 25, 2023)` ([you can see that on Kattis](https://open.kattis.com/problem-sources/2022%20ICPC%20North%20America%20Regional%20Programming%20Contests%20%28February%2025%2C%202023%29)).

To solve this we could:

1. Allow multiple sources, probably making it a "String or sequence of strings".
2. Ignore it, it's fine.
3. Rethink source?

Thoughts?

thorehusfeldt:
While you think about these things, also consider adding a `version` to `problem` (_not_ referring to problem format.)

Tagl:
When you say "rethink source", what do you mean?
What options are there other than no source, one source or multiple sources?

- No source sounds awful to me.
- One source works almost all the time, since in some (many?) of these cases there is some sort of umbrella source.
- Multiple sources sounds like a nice QoL change that does not get used very often, but I'm not seeing the implementation complexity change much so I think it's fine to support that.

Issue #46, opened 2023-07-22
Output-only problems


jsannemo:

evouga:
To be clear, this issue refers to problems where contestants submit the output itself (instead of a program that computes the output).

niemela:
@jsannemo Any ideas on how to do this best?

I don't think I would be against it, but if we don't know how to do it yet, maybe we should cut it for now?

jsannemo:
The obvious way is IMO to have a problem.yaml flag for this that signals "teams should create submissions by submitting the .ans files matching a subset of the .in files". Some details remain on what to do for a submission that doesn't include *all* .ans files.

niemela:
> The obvious way is IMO to have a problem.yaml flag for this that signals "teams should create submissions by submitting the .ans files matching a subset of the .in files". 

This property is orthogonal to `pass-fail` vs `scoring` so it needs to be something else than `type` (or we need 4 values for type).

It's somewhat related to `languages` in the sense that it is a bit like allowing *no* languages.

> Some details remain on what to do for a submission that doesn't include _all_ .ans files.

Uploading a set of files should have the same effect as if a program would have produced those files. Submission that are missing some files should be treated as if that file was empty, or alternatively always fail (for that case).

RagnarGrootKoerkamp:
Probably a new `output-only: true` flag is needed then?

So for pass-fail problems, the only way to solve the problem is to upload the set of all `.ans` files at ones and they must all be correct.

For scoring problems, you can't first submit `1.ans`, then `2.ans`, then `3.ans`, because your score would be `max(score 1.ans, score 2.ans, score 3.ans)`, instead of `sum(score 1.ans, score 2.ans, score 3.ans)`, so you have to incrementally submit more and more files. (Well, unless the judge does some special handling but that sounds like a lot of work.)

Tagl:
I'd say it's kind of like adding a single language that is similar to echo and restricting to that language. Isn't that how you'd accomplish this now, that is provide a driver in a specific language that simply outputs the contents of the correct file?

If I wanted to do something like this I'd just use include but it would be nicer if the language of the submission was Text or Files or something instead of C++/Python. I'm not sure this is really a new type of problem but rather just some sort of language option.

Tagl:
Alright, yes I do agree now that this is a different type.
This kind of falls under validation, but with drastic changes and is not really compatible with the other options there.

And I must say, I do love this idea. Allows for Capture the Flag, but is more powerful.

[...]

PR #128, opened 2023-09-22
Draft CUE schemas for 7505413

thorehusfeldt:
Here are two CUE schemas (for `problem.yaml` and `testdata.yaml`) that try to capture the state of the specification as of `2023-07-draft`,  21. Sep 2023, commit [7505413](https://github.com/Kattis/problem-package-format/commit/7505413f3e9d327e4f1f65e77950dbf6f012e64c).

I am quite happy with these, except for the following.

# 1. `problem_format_version`

Please seriously consider adopting semantic versioning, like `2.0.0-rc1` or `1.1-alpha`? We’re shooting ourselves in the foot with a home-made versioning tradition. It leads to confusion and frustration, and deliberately avoids lots of useful tools (in documentation and version control) that have been created to facility exactly our problem (namely communicating about versions)

# 2. `validation` 

This is a mess. Currently we have, at best,

```cue
validation?: *"default" | string
```
and a prose specification substrings of "custom, multipass, interactive, score” that is really hard to specify (much less provide editor support for.) Allowed is `validation: default` and `validation: score multipass interactive` but not `vaildation: interactive interactive default`.

I propose to actually use YAML for this, so we’d have 

```cue
validation?: *"default" |  close({
  multipass?: bool
  interactive?: bool
  score?: bool
})
```
so you write `validation: default` (or leave it out) or `validation: { interactive: True; score: True }`. Parsing is for free, specification is clear, editor support (including autocomplete) is out-of-the-box.

We can, as a compromise, do something like this:
```cue
  validation?: string
  _validation_ok: strings.Split(validation, " ") & (["default"] | ["custom", ...#validation_flags])
  _validation_flags: "interactive" | "score" | "multipass"
```

which would allow `default` and `custom` and `custom score interactive`, but not `default score` or `interactive` or `foo`. But this would still allow `custom score score score score` and disallow `interactive custom`. And it’s quite opaque, will give very little editor support, and feels like a bad use of strings for a use case that YAML exactly solves. 

The details of how (and if) to specify

```cue
if validation.score != _|_ {type: "scoring"}
```
(“scoring validators can only exist for scoring problems”) are downstream from this.

## 2.1

Do we really mean `score` instead of `scoring` here? That doesn’t fit, linguistically.

# 3. Grading

In `testdata.yaml`, we now have 

```yaml
grading?: {
  score?:       string
  max_score?:   string
  aggregation?: "sum" | "min"
```
I thought was called `scoring`? The phrase “grading“ makes little sense here to me, _grading_ is no longer a concept in the specification.

I think this should just be called `scoring` (the key has had that name earlier.)

# 4. Number types for `score`, `max_score`

See above, I think we mean

```yaml
score?:       number
max_score?:  number
```

I think it’s uncontroversial and have updated the PR with that change.

### DIFF
--- a/support/schemas/problem.cue
+++ b/support/schemas/problem.cue
@@ -1,0 +1,41 @@
+package problem_package
+
+#problem_settings: {
+	name!:                    string | close({[#language_code]: string})
+	problem_format_version?: *"legacy" | "draft" | =~"^[0-9]{4}-[0-9]{2}(-draft)?$"
+
+	author?:       #author_information | [...#author_information]
+	source?:       string
+	source_url?:   string // only allow if source exists
+	license?:      *"unknown" | "public domain" | #license_with_rights
+	rights_owner?: string
+	if rights_owner != _|_ {license?: #license_with_rights}
+
+	type?:       *"pass-fail" | "scoring"
+	validation?: *"default" | "custom" | close({["multipass" | "interactive" | "scoring"]: _})
+	limits?: {
+		time_multiplier?: {
+			ac_to_time_limit?:  *2.0 | number
+			time_limit_to_tle?: *1.5 | number
+		}
+		time_limit?:      number & >0
+		time_resolution?: *1.0 | number
+		[#other_limits]:  int
+	}
+	constants?: {[=~"^[a-z0-9]+$"]: number | string}
+	keywords?: string | [...string]
+	uuid?:     string
+	keywords?: string | [...string]
+	if validation.scoring != _|_ {type: "scoring"}
+	languages?: *"all" | [...string]
+}
+
+#problem_settings
+
+#author_information : string | {
+	name!: string
+	email?: string
+}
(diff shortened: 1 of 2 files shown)


niemela:
@thorehusfeldt I suggest creating separate issues for the questions that need further discussions. Mixing them all up in the same thread is just unnecessarily complicating things.

Some quick comments:

1. Are you referring to the format of `problem_format_version`, or the introduction (and format of) `version`, which I assume would be the version of the problem package, as suggested here: https://github.com/Kattis/problem-package-format/issues/125#issuecomment-1726171165?

2. Yes, this should be cleaned up. We should also consider the relationship between `type` and `validation`. I.e. some of the information that is now in `validation` could be in `type` (and *possibly* vice versa?) See also discussions in #46.

3. Do you have some suggestions?

4. Yes, I think we do.

niemela (approved):

PR #129, opened 2023-10-02
Make UUID required

### LINKED ISSUE #124: Make UUID required
Should we make UUID required?

It's one of those things that where:
- It's actually kinda useful for tooling downstream (i.e. problem archives and online judges) if it's actually used.
- It's not that useful at all for the people that would have to add them (i.e. authors).
- It would be very easy for tooling to generate and prefill UUIDs for users.
I.e. it could add some value to make it required?

Alternatively we could make the "recommended" (which doesn't really mean anything, but could probably help just as much as requiring them), and make sure that all "our" tools generate them.

We could also have the ICPC Problem Archive "require" them internally. Which both means nothing, because we would just have (internal) tooling that adds UUIDs when they are missing, so no-one else would have to do anything, and kinda useful, because it would create a public catalogue of UUID to known problem mappings.

Thoughts?

niemela:
Fixes #124


thorehusfeldt:
I tried to change the schema accordingly, `uuid?` needs to be `uuid!` (the exclamation point means “this key has to have a concrete value”), but I may no longer have push access.

niemela:
@thorehusfeldt Does the change to problem.cue I just added look good?

PR #140, opened 2023-12-03
Extend authors to credits

### LINKED ISSUE #67: Extend credit list
If a more granular set of credits should be supported, then what should be supported?

- Problem Authors (idea)
- Contributors (setup)
- Translators
- Testers
- (Dedications?)

Tagl:
Originally I thought to use `accreditation` instead of `person` but it felt a little long. Open to suggestions.
The table changed due to formatting, but the real changes are `author` -> `credits` and `rights_owner` was updated ever so slightly and maybe should be updated to reference credits.

Closes #67


RagnarGrootKoerkamp:
```yaml
credits:
  authors:
    - Ragnar
```
feels somewhat verbose. Would we like to allow the following:?

1. Allow omitting `authors` if it's the only field?

```yaml
credits:
    - Ragnar
```

2. Allow a string instead of list of strings for single-authors, i.e.
```yaml
credits:
  authors: Ragnar
```
or even
```yaml
credits: Ragnar
```
or
```yaml
credits:
  authors:
    name: Ragnar
    email: ragnar@gmail.com
```
or
```yaml
credits:
  name: Ragnar
  email: ragnar@gmail.com
```

Tagl:
I honestly think it is fine this way and isn't too verbose.
I'd definitely prefer not to mix the types too much but I do see your point.
I'd rather skip `credits` than default to `authors` myself, as I feel `authors` holds more meaning.
So instead of
```yaml
credits:
  authors:
  - Author 1
  contributors:
  - Contributor 1
```
becomes
```yaml
authors:
- Author 1
contributors:
- Contributor 1
```

niemela:
1. I generally think that "list of X or X (if there is only one)" is a good idea. The main benefit of YAML (over the "obvious" alternatives like JSON. TOML, or INI) is that it's easy to read and write for humans, allowing "list of X or X" seems in line with that.

2. I generally think that "map or X (where the main key maps to X)" is a good idea. It gets a bit complicated when X is also a map though, so maybe not always.

3. I generally think that fewer levels of maps is better, but OTOH, grouping vert same things in the same to level map gives some very helpful structure. So just dumping everything on the root level is not always best.

1 & 2 implies @RagnarGrootKoerkamp's suggestion, which I think is very good.

3 implies @Tagl's suggestion, which I'm torn about. I'm not sure id 3 is a good principle in this case. I think that maybe the `credits` grouping has some value? If we do it though I think we should still also do 1.

spec/2023-07-draft.md
```diff
@@ -111,43 +111,73 @@ The keys are defined as below.
...
+Allowed keys for credits.
+
+A person is specified as either a string with their full name (optionally with an email formatted as `Full Name \<fullname@problem.example\>`), or a map with keys `name` and `email`, which both have values of type string.
+
+| Key          | Type                                   | Default | Comments
+| ------------ | -------------------------------------- | ------- | --------
+| authors      | Person or sequence of persons                    |         | The people who conceptualized the problem.
+| contributors | Person or sequence of persons                    |         | The people who developed the problem package, such as statement, validators and test data.
+| testers      | Person or sequence of persons                    |         | The people who tested the problem package, for example by providing a solution and reviewing the statement.
+| translators  | Map of strings to sequences of persons           |         | The people who translated the statement to other languages. Each key must be an ISO 639 language code.
```
niemela, line 142: As discussed on call. Maybe add "acknowledgements"?

spec/2023-07-draft.md
```diff
@@ -111,43 +111,73 @@ The keys are defined as below.
...
+| name                                     | String or map of strings                                    |                                                          | Required. The name of the problem in each of the languages for which a problem statement exists. See below for details.
+| uuid                                     | String                                                      |                                                          | Required. UUID identifying the problem.
+| <span class="not-icpc">type</span>       | <span class="not-icpc">String</span>                        | <span class="not-icpc">pass-fail</span>                  | <span class="not-icpc">One of `pass-fail` and `scoring`.</span>
+| credits                                  | Map with keys as defined below                              |                                                          | Who should get credits. See below for details.
```
niemela, line 120: Allow for only supplying a single person, that would be the author.

niemela (changes requested):

thorehusfeldt:
So if I understand this correctly, you’re suggesting the following. In my own schema for `problem.yaml` I already have

```cue
// Current problem.yaml specification according to Thore
#problem_settings: {
    name!:   string | close({[#language_code]: string})
    author?: #author_information | [...#author_information]
    ...
}
#language_code:       =~"^[a-z]{2,4}(-[A-Z][A-Z])?$"
#author_information : string | {
    name!:  string
    email?: string
}
```
The new format would then be
```cue
#problem_settings: {  
    credits: string | {
        authors!: [...#author_information]
        contributors?: [...#author_information]
        testers?: [...#author_information]
        translators?: { [#language_code]: [...#author_information] }
    }
    ...
}
```
With `credits: "foo"` meaning `credits: { authors: [{name: "foo"}] }`.

An alternative would be to allow either of the old `author` _or_ the new `credits` (but then without the `string` option). This “either-of” thing is very easy to specify formally (even in JSON, called `oneOf`). Then you could do

```yaml
author:
  name: Alice
  email: alice@foo.bar
```
or
```yaml
credits: 
  authors:
    - name: Alice
     email: alice@foo.bar
```
but you can not have _both_ an `author` and a `credits` key in the same `problem.yaml`. Backwards compatible, light-weight, easy to specify (and provide editor support), easy to parse, unambiguous, unsurprising.

thorehusfeldt:
Playing around with the syntax, here are some examples:
```yaml
   name: Hello World
    credits:
      authors:
        - name: Maggie McAuthorson
          email: ma@authors.com
    ---
    name: Hello World
    credits:
      authors: [Peter, Paul, Mary]
    ---
    name: Hello World
    credits:
      authors: [Peter, Paul, Mary]
      testers:
        - name: Toby Tester
          email: toby@gmail.com
        - name: Eddie Errorfinder
          email: ee@hotmail.com
      translators:
        da: [Thore] # can't to "da: Thore"
        se:
          - name: Simon Lindholm
            email: sl@lindholm.com
```
(You can play around with this, interactively, yourself at https://cuelang.org/play/?id=dUjV5r1dnc_C#cue@export@yaml by editing the string `example` at the bottom and see what kinds of things validate.)

The one suggestion I have is to replace, say, `testers?: [...#person] }` by `testers?: #person | [...#person]` everywhere. (This allows single `#person` maps as well as lists of person maps, and introduces very little overhead.)

niemela:
That looks about right. One missing thing is that these should be the same:
```yaml
credits:
   authors: Person One <person1@example.com>
   testers: [Person Two <person2@example.com>, Person Three <person3@example.com]
---
credits:
   authors: 
      name: Person One 
      email: person1@example.com>
   testers:
      - name: Person Two 
        email: person2@example.com 
      - name: Person Three
        email: person3@example.com
```

spec/2023-07-draft.md
Tagl: Added

spec/2023-07-draft.md
Tagl: Added

niemela (approved):

PR #141, opened 2023-12-03
Add run_samples option in testdata.yaml

### LINKED ISSUE #40: Invalid sample files
We had some problems with randomized solutions where we guarantee n=1000, but that is unpractical in teh samples, so we had those with much smaller n, eg n=10.

It would be nice to officially support samples that are only shown, but not run.

I implemented this with a `.{in,ans}.statement` extension, but open to other options.

Tagl:
Closes #40

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -362,6 +362,7 @@ Unlike `.in` and `.ans` files for non-interactive problem, interactive `.in` and
 not in the problem statement, nor as part of sample input download.
 If you want to provide files related to interactive problems (such as testing tools or input files) you can use [attachments](#attachments).
 
+
 ### Test Data Groups
 
 <div class="not-icpc">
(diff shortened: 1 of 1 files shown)


RagnarGrootKoerkamp (approved):

RagnarGrootKoerkamp:
domjudge does not support `input_validator_flags`, but it doesn't matter because domjudge anyway doesn't do input validation. Could you add some text so say that `samples` **will** be input_validated, and that `input_validator_flags` can be used to modify this.

Tagl:
> Could you add some text so say that `samples` **will** be input_validated, and that `input_validator_flags` can be used to modify this.

Added

niemela (changes requested):
`run_samples` should be listed in the table (just above the "Skipping Execution of Samples" section).

Tagl:
> `run_samples` should be listed in the table (just above the "Skipping Execution of Samples" section).

Added

niemela (approved):

niemela:
There is maybe some wordsmithing and restructuring I would want to do, but this is the meaning we have agreed on. Will merge shortly.

Issue #29, opened 2023-07-21
Add support for bad input


niemela:
Allow specifying of test data that must be failed by some input validator.

Suggested by @jsannemo.

jsannemo:
**Decision**: `data/invalid_inputs`. Supports `input_validator_flags` in `testdata.yaml`:s

RagnarGrootKoerkamp:
Groups and testdata.yaml's are also allowed inside `data/invalid_inputs`

Issue #143, opened 2023-12-06
Add `data/invalid_output`


RagnarGrootKoerkamp:
Similar to `data/invalid_input/*.in`, it would be very nice to have `data/invalid_output/*.{in,ans,out}`.

For each testcase, a `testcase.in` and `testcase.out` **must** be present, and a `testcase.ans` **may** be present.
This then runs the `output_validator` on the given `out` file, which must exit with code `43` (meaning wrong answer). Otherwise it's an error.

cc @mzuenni

niemela:
The `.in` file for an invalid output, must validate as input correctly, right?

Could this be merged with the `invalid_input` directory somehow? (If so probably renaming it to `invalid`? 

Maybe the `invalid_output` logic applies when there is an `.out` and the `invalid_input` logic applies when there isn't.

mzuenni:
Yes, the `.in` file should be valid in this situation.
Bapctools originally handled input and output in the same directory (i.e. just a `bad` directory), but I think two distinct directories is cleaner.

thorehusfeldt:
I am converging to `data/invalid`, which contains invalid testcases (including default answer files.)

Consider the problem add two number with input format
```
4 
1 2
10 11
-1 5
0 0
```
and expected answer
```
3
21
4
0
```
In particular, the answer file must contain `n` many integers; the problem description would tell us this.

Then a tool can now reject an answer file for syntactic reasons, for instance rejecting the following testcase:
```
bad_output_format:
  in: |
    2
    10 11
    -5 4
  ans:
    21
```
(This is not science fiction; BAPCtools’ answer validators do exactly this.)

But for this to work, the answer validator needs access to both `testcase.in` and `testcase.ans`. (Of course, the output validator would need this as well.) 

—

I’m happy to restrict to the file types `data/invalid/**/*.in` and `data/invalid/**/*.ans` (but to make me really happy it should find `testdata.yaml`).

thorehusfeldt:
Here’s my suggestion:

> ### Invalid Testcases

> The `/data`  may contain an `invalid` directory containing testcases that must be rejected by validation. In particular, every `.in`-file in `/data/invalid` must be rejected by at least one input validator. The directory /data/invalid can organised into a tree-like structure similar to `/data/secret`, and contain flags in `testdata.yaml` files that are passed to the validators.

> Unlike `/data/sample` and `/data/secret`, there need not be a `bad-testcase.ans` file for every `bad-testcase.in`. 

[...]

Issue #98, opened 2023-07-28
When must default answer files exist?


thorehusfeldt:
Honest question. The specification is silent about when a test case must provide `.ans`.

1. I’m reasonably sure that for `validation: default` it is true that if `<base_name>.in` exists then `<base_name>.ans` must also exist. We should say that.
2. I _think_ this is true for `validation: custom`. This is highly implicit, but since `judge_answer` in the specification needs a value, the adventurous reader can guess that this value was taken to be `<base_name>.ans`. I’m not sure, though.
3. The answer is not “always”, because explicitly now, for `validation: interactive: True` there need be no such file in `data/sample`

Is the rule “whenever an input file `.in` exists then a default answer file `.ans` must also exist”? (I think so.)

RagnarGrootKoerkamp:
I like the simple requirement that a `.ans` must always exist, but indeed, for interactive problems we tend to just create empty files there, so I'm also OK with 'a `.ans` must exist unless the problem is interactive, in which case it may exist.'.

(Whether the file is 'default'/empty or not is independent of this. I don't think we should ever specify anything about its contents, only about its existence.)

Tagl:
I agree with Ragnar about the simple requirement, simple is good.
But if there is strong desire to make it optional for interactive problems then that's fine by me too.

jsannemo:
While interactive validator never strictly "need" the ans in the same way as a general custom validator, it certainly happens. For example, in https://open.kattis.com/problems/bitgame the `.ans` contains the answer (YES/NO) that is correct.

Thus I definitely prefer requiring `.ans` to exist; it feels a bit too risky to leave it to be empty by default instead (since we can't know if the author signals that intent or forgot a particular .ans file).

thorehusfeldt:
So, let me try:

* a test case `tc` is defined by the _contents_ of it input file (`tc.in`) (and not the filename or even pathname.)
* if `/path/to/tc.in` exists then `/path/to/tc.ans` must also exist.

RagnarGrootKoerkamp:
> * a test case `tc` is defined by the _contents_ of it input file (`tc.in`) (and not the filename or even pathname.)

How/where is this relevant? (It seems unrelated to this specific issue.)

> * if `/path/to/tc.in` exists then `/path/to/tc.ans` must also exist.

agreed

niemela:
> > * a test case `tc` is defined by the _contents_ of it input file (`tc.in`) (and not the filename or even pathname.)
> 
[...]

PR #148, opened 2024-01-23
Invalid testcases (input and ouput validation)

thorehusfeldt:
(28 Jan: Directory structure changed.)

This section extends the scope of invalid _inputs_ to invalid _testcases_. It _retains_ the semantics of what a single test case input file `invalid_testcase.in` should mean (namely, that it should be rejected by at least one input validator). It _adds_ semantics for invalid outputs consistent with the rest of the specification.

The resulting body text is between the next two rules:

---
### Invalid Test Cases

The `data` directory may contain directories of test cases that must be rejected by validation.
Their goal is to ensure integrity and quality of the test data and validation programs.

**Invalid input.** 
The files under `invalid_input`  are invalid  inputs. Unlike in `sample` and `secret`, there are no `.ans` files. Each `tc.in` under `invalid_input` must be rejected by at least one input validator.

**Invalid output.**
The test cases in `invalid_output` describe invalid outputs for non-interactive problems. They consist of three files. The input file `tc.in` must _pass_ input validation. The default answer file `tc.ans` must _pass_ output validation. The output file `tc.out` must _fail_ output validation.

In particular, for an existing feedback directory `dir`, 
```bash
output_validator tc.in tc.ans dir [flags] < tc.ans # MUST PASS
output_validator tc.in tc.ans dir [flags] < tc.out # MUST FAIL
```

<span class="not-icpc">
The directory `invalid` can be organised into a tree-like structure similar to `secret`, and contain flags in `testdata.yaml` files that are passed to the validators.
</span>

---

You can have, say for [howl](https://open.kattis.com/problems/howl), specified in yaml just to be terse:
```yaml
invalid_input:
  hw_in:
    in: "AOHW"
  consec_w_in:
    in: "AHOWW"
invalid_output:
  echo:
    in: "AAHOOW"
    ans: "AWAWHOO"
    out: "AAHOOW"
  short:
    in: "AAHOOW"
    ans: "AWAWHOO"
    out: "AHOW"
```

This _solves_ #143 .

Related: #98, closed issue  #29

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -422,18 +422,22 @@ The samples should still be validated by input and output validators.
 Note that the keys `input_validator_flags` and `output_validator_flags` in `testdata.yaml`, as described above,
 may be used to alter the behavior of the validators for the sample test data group.
 
-### Invalid Input Files
+### Invalid Test Cases
 
-In the `data/` directory, there may be an `invalid_inputs/` directory containing input files that must be rejected by at least one input validator.
-These are meant to only test the input validators, and are not used for judging.
-The rejected input files can be organized into a tree-like structure similar to the test data.
+The `data` directory may contain an `invalid` directory of test cases that must be rejected by validation.
+
+Invalid input
+: Unlike `sample` and `secret`, there does not have to be a `invalid-testcase.ans` file for every `invalid-testcase.in` in `invalid`. 
+Every such `.in`-file without a matching `.ans`-file must be _rejected_ by at least one input validator.
+
+Invalid output
+: If `invalid-testcase.ans` exists then `invalid-testcase.in` must exist and must _pass_ input validation.
+If `invalid-testcase.out` exists, then `invalid-testcase.ans` must exist, and the `.out`-file must be _rejected_ by the output validator. 
 
 <span class="not-icpc">
-
-By default, the `input_validator_flags` of the `data/testdata.yaml` apply to the invalid inputs as well.
-There may be additional `testdata.yaml` files (or symlinks) within the `data/invalid_inputs/` tree, but note that only the value of `input_validator_flags` will be used.
-
+The directory `invalid` can be organised into a tree-like structure similar to `secret`, and contain flags in `testdata.yaml` files that are passed to the validators.
 </span>
+
 
 <div class="not-icpc">
 
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -433,7 +433,8 @@ Unlike in `sample` and `secret`, there are no `.ans` files.
 Each `tc.in` under `invalid_input` must be rejected by at least one input validator.
 
 
-**Invalid output.**
+#### Invalid output
+
 The test cases in `invalid_output` describe invalid outputs for non-interactive problems.
 They consist of three files.
 The input file `tc.in` must _pass_ input validation. 
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -424,15 +424,28 @@ may be used to alter the behavior of the validators for the sample test data gro
 
 ### Invalid Test Cases
 
-The `data` directory may contain an `invalid` directory of test cases that must be rejected by validation.
+The `data` directory may contain directories of test cases that must be rejected by validation.
+Their goal is to ensure integrity and quality of the test data and validation programs.
 
-Invalid input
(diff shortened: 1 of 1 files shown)


RagnarGrootKoerkamp:
So to clarify, there are now two allowed patterns:
1. A single `<testcase>.in`, which should fail input validation.
2. All three of `<testcase>.{in,ans,out}`, where the `.out` should fail output validation. `<testcase>.in` must pass input validation.

Question: do we allow the second case also for `validation: default` problems? In that case we can check whether the default output validator correctly rejects some output.

thorehusfeldt:
> So to clarify, there are now two allowed patterns:

No, it’s more lenient. In particular, it is allowed to have `.ans` and `.in` (but not `.out`.)

> Question: do we allow the second case also for validation: default problems? In that case we can check whether the default output validator correctly rejects some output.

That is certainly allowed, and certainly something I’d expect to be used a lot, using exactly the above specification.

——

The above suggestion is a _minimal_ specification. We can extend with a third case. Maybe that is what you are looking for:

`.ans` exists, `.in` exist, but .`out` does _not_ exist. 

In this case. the `.in` file must _pass_ all input validators, and the output validator must _reject_ in the following invocation:
```bash
% output_validator tc.in tc.ans feedbackdir [flags] < tc.ans
```
Let me repeat that no distinction between default output validators and custom output validators is made.

I would be enthusiastic about adding this third case, but then the spec would have to talk about invocations of output validators quite early. I tried to keep the PR minimal (and still useful.)

thorehusfeldt:
If you like this presented as a table, here it is:
 
 | `.in` | `.ans` | `.out` | Meaning |
 | --- | --- | --- | --- |
 | X | | | _Invalid input_. Must fail input validation |
 | | X | | Forbidden (useless)|
 | | | X | Forbidden (useless) |
 | X | X | | Allowed, but no semantics implied. BAPCtools would answer-validate this. |
 |  | X | X | Forbidden (useless)|
 | X |  | X | Forbidden (useless)|
 |X |X |X | _Invalid output_, must fail output validation, pass input validation |
 
_Discussion_. Note that there _are_ useful semantics that _could_ be given to the cases I label “useless” (and therefore forbid them). For instance, we could say that a sole `.out` file is allowed and then should fail output validation _for all valid input files_. Mathematically, this makes sense.

But we cannot _implement_ such semantics in the current framework. For instance, we don’t have the concept of an output validator that fails _no matter what the input is_. We could have done this, for instance by adding some kind of `--input_oblivious` flag to the specification of output validators. But we haven’t done this. (We also cannot run the output validator on _all_ valid input files, because they might be difficult to generate  and there typically are too many. A compromise would be to run on all `.in` _in the repo_, but this quickly gets very messy, for instance because of flags.)

Hence I deem that none of that is worth doing and thus propose a minimally viable (and still very useful) definition above. More ambitious suggestions can certainly be made.

spec/2023-07-draft.md
```diff
@@ -422,19 +422,36 @@ The samples should still be validated by input and output validators.
...
-### Invalid Input Files
+### Invalid Test Cases
 
-In the `data/` directory, there may be an `invalid_inputs/` directory containing input files that must be rejected by at least one input validator.
-These are meant to only test the input validators, and are not used for judging.
-The rejected input files can be organized into a tree-like structure similar to the test data.
+The `data` directory may contain directories of test cases that must be rejected by validation.
+Their goal is to ensure integrity and quality of the test data and validation programs.
+
+**Invalid input.**
```
niemela, line 430:
```suggestion
#### Invalid input

```

spec/2023-07-draft.md
```diff
@@ -422,19 +422,36 @@ The samples should still be validated by input and output validators.
...
+**Invalid input.** 
+The files under `invalid_input`  are invalid  inputs.
+Unlike in `sample` and `secret`, there are no `.ans` files.
+Each `tc.in` under `invalid_input` must be rejected by at least one input validator.
 
-<span class="not-icpc">
 
-By default, the `input_validator_flags` of the `data/testdata.yaml` apply to the invalid inputs as well.
-There may be additional `testdata.yaml` files (or symlinks) within the `data/invalid_inputs/` tree, but note that only the value of `input_validator_flags` will be used.
+**Invalid output.**
```
niemela, line 436:
```suggestion
#### Invalid output

```

niemela (changes requested):

niemela (approved):

Tagl (approved):

PR #150, opened 2024-03-20
Clarify working directories for submissions and validators

### LINKED ISSUE #5: Specify working directory for validators, graders, and generators
Especially important when the program is a directory

Tagl:
Closes #5

Working directory includes submission files for execution of submissions and output validators.
This allows both the submission and the output validator to easily access files which are submitted, or even created by the submission.
For example, perhaps the submission must write to a file some extra information which the output validator then verifies.
The output validator can then also access its own directory using the argv method discussed in the issue.
For input validators, there is no executing submission, so for convenience of writing the input validator, having the working directory as the validator's directory is nice.

I believe these changes cover all that was discussed.

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -41,6 +41,7 @@ All floating point numbers must be given as the external character sequences def
 There are a number of different kinds of programs that may be provided in the problem package: submissions, input validators, and output validators.
 All programs are always represented by a single file or directory.
 In other words, if a program consists of several files, these must be provided in a single directory.
+In the case that a program is a single file, it is treated as if a directory with the same name takes its place, which contains only that file.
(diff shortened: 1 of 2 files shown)


spec/2023-07-draft.md
```diff
@@ -93,6 +94,11 @@ For languages where there could be several entry points, the default entry point
 | typescript  | TypeScript   |                     | .ts                       |
 | visualbasic | Visual Basic |                     | .vb                       |
 
+The binary (and other artifacts that result from compiling the program) must be placed in the program's directory (or a copy of it).
```
Tagl, line 97: As noted in the issue, we may want to skip mentioning this at all.

niemela (approved):

PR #153, opened 2024-03-21
Define test cases

### LINKED ISSUE #98: When must default answer files exist?

Tagl:
Hopefully closes #98

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -317,7 +317,7 @@ If the test case exists, then some of the other files are required to exist whil
 If the test case does not exist, then the other files must not exist.
 The table below summarizes the supported test data:
 
-| Extension                       | Described In         | Summmary                               |
+| Extension                       | Described In         | Summary                                |
 |---------------------------------|----------------------|----------------------------------------|
 | `.in`                           | Input                | Input piped to standard input          |
 | `.ans`                          |                      | Expected (AC) answer                   |
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -328,6 +328,13 @@ The table below summarizes the supported test data:
 | `.args`                         | Input                | Input passed as command-line arguments |
 | `.files`                        | Input                | Input available via file I/O           | 
 
+Judge systems may assume that the result of running a program on a test case is deterministic. 
+In other words, for any two test cases, if the contents of their `.in` and `.args` files,
+and `.files` folder are equivalent, then the input of the two test cases is equivalent. 
+Additionally, for any two test cases, if their input,
+<div class="not-icpc"> output validator flags (see test data groups below)</div>
+and the contents of their `.ans` files are equivalent, then the test cases are equivalent. 
+For that purpose, a judge system could choose to reuse the result of a previous run, or re-run the equivalent test case.
 
 ### Input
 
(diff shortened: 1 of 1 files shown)


spec/2023-07-draft.md
```diff
@@ -322,15 +328,24 @@ All files and folders associated to a single test case have the same base name w
...
+For that purpose, a judge system could choose to reuse the result of a previous run, or re-run the equivalent test case.
+
 ### Input
 
 Each test case can supply input via standard input, command-line arguments, and/or the file system. 
 These options are not exclusive. 
-For a test case with base name `test`, the file `test.in` (if it exists) is piped to the submission as standard input. 
-The submission's filename is passed as the first command-line argument, followed by the whitespace-separated tokens in `test.args` (if this file exists).
+For a test case with base name `test`, the file `test.in` is piped to the submission as standard input. 
+The submission's filename is passed as the first command-line argument, followed by the whitespace-separated tokens in `test.args`, if the file exists.
```
Tagl, line 344: I do not think this is the right wording.
This is correct in the case of a single file submission and, for example, python.
In general this is not true and I think we should simply skip mentioning the first argument or say that it follows from how programs are run in general, so the main entry point is the first argument, whether it be binary, main class or main file.

spec/2023-07-draft.md
```diff
@@ -322,15 +328,24 @@ All files and folders associated to a single test case have the same base name w
 | `.args`                         | Input                | Input passed as command-line arguments |
 | `.files`                        | Input                | Input available via file I/O           | 
 
+Judge systems may assume that the result of running a program on a test case is deterministic. 
+In other words, for any two test cases, if the contents of their `.in` and `.args` files,
```
niemela, line 332:
```suggestion
For any two test cases, if the contents of their `.in` and `.args` files,
```

spec/2023-07-draft.md
```diff
@@ -322,15 +328,24 @@ All files and folders associated to a single test case have the same base name w
...
+and `.files` folder are equivalent, then the input of the two test cases is equivalent. 
+Additionally, for any two test cases, if their input,
```
niemela, line 334:
```suggestion
This means that for any two test cases, if their input,
```

spec/2023-07-draft.md
```diff
@@ -322,15 +328,24 @@ All files and folders associated to a single test case have the same base name w
...
+<div class="not-icpc"> output validator flags (see test data groups below)</div>
+and the contents of their `.ans` files are equivalent, then the test cases are equivalent. 
+For that purpose, a judge system could choose to reuse the result of a previous run, or re-run the equivalent test case.
```
niemela, line 337:
```suggestion
The assumption of determinism means that a judge system could choose to reuse the result of a previous run, or to re-run the equivalent test case.
```

spec/2023-07-draft.md
niemela: That's a good point. I really don't know how to write that though... :)

niemela (changes requested):

spec/2023-07-draft.md
Tagl: Will make ticket for this

niemela (approved):

niemela:
I'll merge this. @Tagl please create a ticket for the reminding issue that you brought up.

PR #154, opened 2024-03-21
Clarify inheritance for testdata.yaml

### LINKED ISSUE #113: Test data settings in `testdata.yaml` and inheritance
How are test data settings inherited?

Currently we have the following explanations. In `legacy`,

> In each test data group, a file testdata.yaml may be placed to specify how the result of the test data group should be computed. If a test data group has no testdata.yaml file, the testdata.yaml in the closest ancestor group that has one will be used. If there is no testdata.yaml file in the root data group, one is implicitly added with the default values.

At my behest, we changed this slightly in `2023-07-draft`.

> […] If a test data group has no testdata.yaml file, the testdata.yaml in the closest ancestor group that has one will be used. If there is no testdata.yaml file in the root data group, one is implicitly added with the default values.

Alas, with the significant changes @jsannemo is proposing for the scoring/grading framework (see https://github.com/Kattis/problem-package-format/blob/master/spec/2023-07-draft.md#grading) , this is no longer good enough.

For instance, if `secret/group1/testdata.yaml` does not exist then  then its default value of `grading/aggregation` becomes that of its parent (by the above rule), which is `sum`. But the specification gives `min` a default of `grading.aggregation`. This is at best confusing, at worst gives unintended behaviour.

# Solution 1: wordsmithing

It may be possible to introduce super-precise wording for inheritance or “what default means” that ameliorates this problem. I am not too confident, however.

# Solution 2: different inheritance rules for validator flags and grading/scoring

The rule could be that validator flags _are_ inherited (useful) but grading flags aren’t (also useful). That’s much easier to communicate (and implement.)

# Solution 3: Separate files

3a. Validator flags and grading flags play very different roles, and neither is well-named by `testdata.yaml`. So why not drop `testdata.yaml` and instead use `validator_flags.yaml` and `scoring.yaml`. The intent is very clear, the rules could be very clearly explained, and ICPC-focussed tools and judges can completely ignore `scoring.yaml`.

3b. If it’s worth doing, it’s worth doing right:
* `input_validation.yaml` contains input validator flags. It’s useful for _construction_ tools like `problemtools` and `BAPCtools`. They don’t even need to be exported to a judge.
* `output_validation.yaml` contains output validator flags. In ICPC-traditions, it only appears in at toplevel `data`. Important for _judging_.
* `scoring.yaml` is as above. Important for judging, but only for `type:scoring`. DOM-judge never touches this.

Clean, obvious, easy to describe, easy to maintain, good separation.

Tagl:
Closes #113 

I simply wrote up the consensus according to #113 

I do though think it could be confusing that some things are inherited and others are not.

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -379,7 +379,7 @@ At the top level, the test data is divided into exactly two groups: `sample` and
 <div class="not-icpc">
 
 The <em>result</em> of a test data group is computed by applying a <em>grader</em> to all of the sub-results (test cases and subgroups) in the group.
-See [Grading](#grading "wikilink") for more details.
+See [Scoring](#scoring "wikilink") for more details.
 
 </div>
 
@@ -397,7 +397,7 @@ The format of `testdata.yaml` is as follows:
 
 | Key                    | Type                                | Default                 | Inheritance    | Comments
 | ---------------------- | ----------------------------------- | ----------------------- | -------------- | --------
-| grading                | Map                                 | [See Grading](#grading) | Not inherited  | Description of how the results of the group test cases and subgroups should be aggregated.
+| scoring                | Map                                 | [See Scoring](#scoring) | Not inherited  | Description of how the results of the group test cases and subgroups should be aggregated.
 | input_validator_flags  | String or map of strings to strings | empty string            | Inherited      | Arguments passed to each input validator for this test data group. If a string then those are the flags that will be passed to each input validator for this test data group. If a map then each key is the name of the input validator and the value is the flags to pass to that input validator for this test data group. Validators not present in the map are run without flags.
 | output_validator_flags | String                              | empty string            | Inherited      | Arguments passed to the output validator for this test data group.
 | run_samples            | Boolean                             | true                    | Not applicable | Signifies whether samples should be run, see below for details.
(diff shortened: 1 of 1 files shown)


thorehusfeldt:
Here are some cases that need to be considered:

1.

In `secret/tiny/`
```yaml
input_validator_flags: --max_n 100
```

In secret:
```yaml
output_validator_flags: space_change_sensitive
```

Q: Which output validator flags apply at `secret/group1/testcase1`?

2.

For `grading:` keys, I’m afraid your proposal this will not  give the behaviour intended by @jsannemo . I think that intention can only be realised if `score` is never inherited. (But I could be wrong.)

Tagl:
@thorehusfeldt You are correct, I have pushed a change that should fix it, but we could also rewrite it essentially as you wrote it in the issue.

niemela:
> I simply wrote up the consensus according to #113
> 
> I do though think it could be confusing that some things are inherited and others are not.

I agree that this is a good representation of the consensus in #113, I also agree that it ends up more complicated than what is ideal.

I suggest that we merge this so that we at least get to a consistent state, and then we could open up a separate discussion about simplifying it. (@Tagl could you do that).

niemela (approved):

PR #158, opened 2024-04-10
Change meaning of submissions/<verdict> directories

### LINKED ISSUE #87: Meaning of `submissions/<verdict>` directories
Currently they are specified as

- accepted: Accepted as a correct solution for all test files
- wrong_answer: Wrong answer for some test file, but is not too slow and does not crash for any test file
- time_limit_exceeded: Too slow for some test file. May also give wrong answer but not crash for any test file
- run_time_error: Crashes for some test file

This is basically what you get if you grade with 'priorities', where `AC < WA < TLE < RTE`. In practice though, ICPC uses 'first failure' mode.
This means that these directories do not actually correspond to the expected final vedict.

Does anybody implement priority-based verdicts for testing these submissions?
If not, I'd say we should change them to just mean `the final verdict (using whatever grading scheme is in place) must be X`

Tagl:
Closes #87

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -544,14 +544,14 @@ that file must have the language dependent name as given in the table referred a
 Correct and incorrect solutions to the problem are provided in subdirectories of `submissions/`.
 The possible subdirectories are:
 
-| Value                                             | Requirement                                                                                                   | Comment
-| ------------------------------------------------- | ------------------------------------------------------------------------------------------------------------- | -------
-| accepted                                          | Accepted as a correct solution for all test files                                                             | At least one is required.
-| <span class="not-icpc"> partially_accepted</span> | <span class="not-icpc"> Overall verdict must be Accepted. Overall score must be less than `max_score`.</span> | <span class="not-icpc"> Must not be used for pass-fail problems.</span>
-| wrong_answer                                      | Wrong answer for some test file, but is not too slow and does not crash for any test file                     |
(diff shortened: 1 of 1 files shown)


spec/2023-07-draft.md
```diff
@@ -544,14 +544,14 @@ that file must have the language dependent name as given in the table referred a
...
+| accepted                                          | Accepted as a correct solution for all test files                                                                  | At least one is required.
+| <span class="not-icpc"> partially_accepted</span> | <span class="not-icpc"> Overall verdict must be Accepted. Overall score must be less than `max_score`.</span>      | <span class="not-icpc"> Must not be used for pass-fail problems.</span>
+| wrong_answer                                      | Wrong answer for some test file. May be too slow or crash for other test files.                                    |
+| time_limit_exceeded                               | Too slow relative to the safety margin for some test file. May output wrong answers or crash for other test files. |
+| run_time_error                                    | Crashes for some test file. May output wrong answers or be too slow for other test files.                          | Very rarely useful.
+| rejected                                          | Is not accepted for any reason in the final verdict.                                                               |
```
niemela, line 554: Why did you not write this similar to the other? I.e. 
```suggestion
| rejected                                          | Is rejected (i.e. not accepted, for any reason) for some test file.                                                               |
```

niemela (changes requested):

spec/2023-07-draft.md
Tagl: This does not have the desired effect for scoring problems, right? We want rejected to mean that it wasn't accepted in the end.

niemela (approved):

PR #160, opened 2024-04-10
Add submission.yaml for example submission metadata

### LINKED ISSUE #132: Specifying entry point
In the list of programming languages we define "default entry points" for some languages. The fact that we talk about a *default* entry point implies that an entry point could be defined somehow. We don't ever touch on this in the spec. some systems have in the past used rules do detect or decide on the entry point, others (e.g. Kattis) allows the user to specify. This UI detail is an implementation detail and we don't have to specify it, *but* I think we do need to specify it for submissions and included files in the package.

So, what is a good way to specify an override for default entry point for submissions or included files? One obvious solution would be to have some yaml file, say `submission.yaml` with a single key `entry-point` to specify this value. It's a bit awkward, but the default value requires no file and would be the most common by far. It also implies that you can't include a file called `submission.yaml` (at least not for languages that needs an entry point), but that seems like a very minor issue.

Thoughts?

### LINKED ISSUE #136: Specifying language
Related to #132.

In the list of programming languages we define "file endings" for all languages. The implication is that this is how we determine what language should be used. For input/output validators we also allow build and run scripts, so oddities can always be handled, but that is not (and should not) be supported for submissions.

There are already collisions in the set of commonly used languages. Specifically Python 2 and Python 3 actually has the same file ending. In the legacy version we require a shebang to distinguish, in 2023-07-draft we currently require .py2 for Python 2. I would argue that both of these are workarounds, and further that the latter is even worse. What is good is that it defaults to Python, and you only have to do strange things if you want to use Python 2 (which you normally shouldn't).

I think that we opted for these workarounds because: 1. they are *almost* not needed anyway, it's just Python, and 2. the other (obvious) option is to have some kind of metadata file specifying this, and we wanted to avoid that if possible.

Regarding reason 1:
- There are actual languages that share file endings, where one is not (more or less) a "deprecated" version of the other (i.e. Python 2 vs Python3). One such example is Objective-C, Octave and Matlab. Kattis currently supports both Objective-C and Octave.
- On Kattis we already have some additional "languages" that share file endings. E.g. "C with GMP", "C++ with GMP", "C with Memory Check", "C++ with Memory Check", "Python 3 with NumPy".

Regarding reason 2:
- If we store end endpoint information discussed in #132, we could store this there as well. I.e. if there already is some metadata file then the additional cost for using it for this as well is minimal.

niemela:
Fixes #132 
Fixes #136

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -558,7 +558,7 @@ It is mandatory to provide at least one accepted solution.
 
 Metadata about the example submissions are provided in a UTF-8 encoded YAML file named `submissions.yaml` placed directly in the `submissions/` directory.
 The top level keys in `submissions.yaml` are globs matching example submissions.
-E.g. `accepted/*` would match any submission in the `submissions/accepted/` directory.
+For example, `accepted/*` would match any submission in the `submissions/accepted/` directory.
 
 Each glob maps to a map with keys as defined below specifying metadata for all submissions that is matched by the glob.
 
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -563,10 +563,11 @@ E.g. `accepted/*` would match any submission in the `submissions/accepted/` dire
 
 Each glob maps to a map with keys as defined below specifying metadata for all submissions that is matched by the glob.
 
-| Key        | Type   | Default                                                               | Comment
-| ---------- | ------ | --------------------------------------------------------------------- | --
-| language   | String | As determined by file endings given in the [language list](#programs) |
-| entrypoint | String | As specified in the [language list](#programs)                        |
+| Key        | Type                          | Default                                                               | Comment
+| ---------- | ----------------------------- | --------------------------------------------------------------------- | -------
+| language   | String                        | As determined by file endings given in the [language list](#programs) |
+| entrypoint | String                        | As specified in the [language list](#programs)                        |
+| authors    | Person or sequence of persons |                                                                       | Author(s) of submission(s)
 
 
 Submissions must read input data from standard input, and write output to standard output.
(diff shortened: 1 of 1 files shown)


spec/2023-07-draft.md
```diff
@@ -554,10 +554,19 @@ The possible subdirectories are:
 | run_time_error                                    | Crashes for some test case. May output wrong answers or be too slow for other test cases.                          | Very rarely useful.
 
 Every file or directory in these directories represents a separate solution.
-Same requirements as for submissions with regards to filenames.
 It is mandatory to provide at least one accepted solution.
 
-Submissions must read input data from standard input, and write output to standard output.
+Metadata abut the example submissions are provided in a UTF-8 encoded YAML file named `submissions.yaml` placed directly in the `submissions/` directory.
```
Tagl, line 559:
```suggestion
Metadata about the example submissions are provided in a UTF-8 encoded YAML file named `submissions.yaml` placed directly in the `submissions/` directory.
```

spec/2023-07-draft.md
```diff
@@ -554,10 +554,19 @@ The possible subdirectories are:
...
+The top level keys in `submissions.yaml` are. globs matching example submissions.
```
Tagl, line 560:
```suggestion
The top level keys in `submissions.yaml` are globs matching example submissions.
```

spec/2023-07-draft.md
```diff
@@ -554,10 +554,19 @@ The possible subdirectories are:
...
+E.g. `accepted/*` would match any submission in the `submissions/accepted/` directory.
```
Tagl, line 561:
```suggestion
For example, `accepted/*` would match any submission in the `submissions/accepted/` directory.
```

PR #168, opened 2024-04-12
Multipass clarifications

### LINKED ISSUE #149: Multipass clarifications
Right now the output validator is unable to save state between runs (except by writing it in the `nextpass.in`). However, it would be useful if the validator could save state between runs without the submission beieng able to access it.

Besides that i have a few additional questions on how multipass problems work:

1. how exactly is the validator invoked? `./valdiator <inputfile> <feedbackdir> [args] < <team.ans>`?
2. how should the `judgemessage.txt` be handled? Does the validator append new stuff or does the judging system append all generated files?

mzuenni:
Closes #149 .


Tagl (approved):

spec/2023-07-draft.md
```diff
@@ -720,11 +720,17 @@ In particular, the judging system must not assume that the validator program cre
...
+The time and memory limit apply for each invocation separately.
+
 To signal that the submission should be run again, the output validator must exit with code 42 and output the new input in the file `nextpass.in` in the feedback directory.
 Judging stops if no `nextpass.in` was created, or the output validator exited with any other code.
+Note that the `nextpass.in` will be removed before the next run.
 
 It is a judge error to create the `nextpass.in` file and exit with any other code than 42.
 
+All other files inside the feedback directory are guaranteed to persist between runs.
+In particular, the validator should only append text to the "judgemessage.txt" to provide combined feedback for all runs.
```
niemela, line 732: I *think* I agree with "should". 

The counterpoint is that I think rewriting the judgemessage in a way that combines the (relevant) information from previous passes but isn't technically an append if fine. But then, "should" isn't "must".

niemela (approved):

PR #169, opened 2024-04-12
Add solution descriptions in `solution/`

### LINKED ISSUE #70: Solution slides / editorials
We should define a place to store solution slides / editorials.

At a minimum in the form of a PDF, but preferably (?) as LaTeX and/or Markdown (i.e. similar to problem statements). This of course introduces more of the #31 issues.

niemela:
Closes #70 

If anything the solution files are somewhat underspecified. I think that's fine (for now), and it's definitely better than being overspecified.

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -345,6 +345,20 @@ However, judging systems must make a best effort to correctly render **at minimu
 
 Public, i.e. non-secret, files to be made available in addition to the problem statement and sample test data are provided in the directory `attachments/`.
 
+## Solution description
+
+A description of how the problem is intended to be solved is provided in the directory `solution/`.
+
+This directory must contain one file per language, for at least one language, named `solution.`\<language\>`.`\<filetype\>.
+Language must be given as the shortest ISO 639 code.
+If needed a hyphen and a ISO 3166-1 alpha-2 code may be appended to ISO 639 code.
+Optionally, the language code can be left out, the default is then English (`en`).
+The set of languages used can be different from what was used for the problem statement.
+Filetype can be either `tex` for LaTeX files, `md` for Markdown, or `pdf` for PDF.
+
+Auxiliary files needed by the problem statement files must all be in `solution/`.
+`solution.<language>.<filetype>` should reference auxiliary files as if the working directory is `solution/`.
+
 ## Test data
 
 The test data are provided in subdirectories of `data/`.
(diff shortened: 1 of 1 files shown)


spec/2023-07-draft.md
```diff
@@ -345,6 +345,20 @@ However, judging systems must make a best effort to correctly render **at minimu
 
 Public, i.e. non-secret, files to be made available in addition to the problem statement and sample test data are provided in the directory `attachments/`.
 
+## Solution description
+
+A description of how the problem is intended to be solved is provided in the directory `solution/`.
+
+This directory must contain one file per language, for at least one language, named `solution.`\<language\>`.`\<filetype\>.
+Language must be given as the shortest ISO 639 code.
+If needed a hyphen and a ISO 3166-1 alpha-2 code may be appended to ISO 639 code.
```
RagnarGrootKoerkamp, line 354: can we drop the technical sentences and just say 'similar to `problem.<lang>.<ext>`'?

spec/2023-07-draft.md
```diff
@@ -345,6 +345,20 @@ However, judging systems must make a best effort to correctly render **at minimu
...
+Optionally, the language code can be left out, the default is then English (`en`).
+The set of languages used can be different from what was used for the problem statement.
+Filetype can be either `tex` for LaTeX files, `md` for Markdown, or `pdf` for PDF.
+
+Auxiliary files needed by the problem statement files must all be in `solution/`.
+`solution.<language>.<filetype>` should reference auxiliary files as if the working directory is `solution/`.
+
```
RagnarGrootKoerkamp, line 361: Currently we're not saying anything about whether we expect `solution.en.tex` to be a fully compilable latex file or just a fragment (similar to `problem.en.tex`).
Maybe we can be more explicit that how they are used is up to tooling?

spec/2023-07-draft.md
niemela: Do you mean drop lines 353-4 and add ", similar to problem.<language>.<filetype>." at the end of line 352?

Yeah, that's better.

spec/2023-07-draft.md
niemela: My problem is that I don't know exactly what that would be.

My reasoning is that "just put whatever you got in this directory" is better than nothing (i.e. what we have now), so I think it's worth adding.

What you are suggesting would be even better though. Could you write a suggestion for what that would look like? If so, I suggest we still merge this and have that be a follow up PR.

RagnarGrootKoerkamp: I just mean to literally add some text that `How solution.en.tex is used is up to the user/tooling.`

I agree that specifying details is too much for now.

niemela: > I just mean to literally add some text that `How solution.en.tex is used is up to the user/tooling.`

Ah, yes, that would make sense.

niemela: ok... I did *something* along those lines.... ?!?

RagnarGrootKoerkamp: works for me :)

PR #171, opened 2024-04-12
Allow source and source_url to be sequences

### LINKED ISSUE #125: Allow for multiple sources?

niemela:
Fixes #125 

The dependency on sequence length between `source` and `source_url` is a bit ugly. An alternative solution would be to make source a sequence of maps with `name` and (optional) `url`.

E.g.:
```yaml
source:
  - name: NWERC 2020
    url: https://some.url.com/yup
  - name: SWERC 2020
    url: https://someother.url.com/here
```

Thoughts?


Tagl:
> An alternative solution would be to make source a sequence of maps with `name` and (optional) `url`.

I like that.

niemela:
Replaced by #181

PR #173, opened 2024-04-14
Remove ICPC subset

niemela:
Remove the markup of a "ICPC subset" from the new version of the standard. 

What was marked up was not really correct, in fact, there is not a clear consensus right now on what exactly is correct. Furthermore, especially after moving more information into type, it is now easier do describe subsets directly.

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -1,17 +1,12 @@
 ---
 layout: default
 title: 2023-07-draft (latest)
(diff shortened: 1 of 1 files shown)


RagnarGrootKoerkamp:
simplifying is always good :)

RagnarGrootKoerkamp (approved):

Tagl (approved):

PR #179, opened 2024-04-14
Change constants from {%x%} to {{x}}; allow unrecognised names

### LINKED ISSUE #176: Handle conflicts with {%name%} style constants
Currently the spec says:

> Names must match the following regex: `[a-zA-Z0-9_.-]`.
> *Constant sequences* are tokens (regex words) of the form `{%name%}`.
> The `name` in every constant sequence must be a valid constant name.

I interpret this to mean the following:

- In a prepossessing step we find every string matching `{%([^%]+)%}` (in all the relevant files).
- If any capturing group does not match `[a-zA-Z0-9_.-]*` that's an error (Side note: the first regex above is wrong, right?)
- If any capturing group is not listed as a constant in `problem.yaml` that is... unclear if it's an error.

I think what we actually want is:

> Names must match the following regex: `[a-zA-Z0-9_.-]*`.
> *Constant sequences* are tokens (regex words) of the form `{%name%}`,
> where `name` is one of the names defined in `constants`

This means that conflicts with other tools that use some `{% %}` syntax would not be a problem (for us) as long as we have not defined a constant with the exact name of what's between the `{% %}`. It also means that we can't make misspelled constant names an error (although we could warn).

Thoughts? 

@mzuenni @RagnarGrootKoerkamp

RagnarGrootKoerkamp:
Fix #176

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -272,8 +272,10 @@ If a list is given, the problem may only be solved using those programming langu
 
 A map of names to values. 
 Names must match the following regex: `[a-zA-Z0-9_.-]`. 
-*Constant sequences* are tokens (regex words) of the form {% raw %}`{%name%}`{% endraw %}.
-The `name` in every constant sequence must be a valid constant name.
+*Constant sequences* are tokens (regex words) of the form `{{name}}`,
+where `name` is one of the names defined in `constants`.
+Tags `{{xyz}}` containing a name that is not defined are not modified, but may be warned for.
+
 All constant sequences in the following files will be replaced by the value of the corresponding constant, formatted as described below:
   - problem statements
   - input and output validators


spec/2023-07-draft.md
```diff
@@ -272,8 +272,10 @@ If a list is given, the problem may only be solved using those programming langu
 
 A map of names to values. 
 Names must match the following regex: `[a-zA-Z0-9_.-]`. 
-*Constant sequences* are tokens (regex words) of the form {% raw %}`{%name%}`{% endraw %}.
-The `name` in every constant sequence must be a valid constant name.
+*Constant sequences* are tokens (regex words) of the form `{{name}}`,
+where `name` is one of the names defined in `constants`. Tags`{{xyz}}`
+containing a name that is not defined are not modified, but may be warned for.
```
niemela, line 277: Start a new line after `.`.

niemela (changes requested):

spec/2023-07-draft.md
mpsijm, line 277: Should there be a space after `Tags` here?

```suggestion
where `name` is one of the names defined in `constants`.
Tags `{{xyz}}` containing a name that is not defined are not modified, but may be warned for.
```

niemela (approved):

PR #181, opened 2024-04-15
Change source so it accepts a sequence of maps

Tagl:


pehrsoderman:
Duplicate of https://github.com/Kattis/problem-package-format/pull/171 ?

spec/2023-07-draft.md
```diff
@@ -109,8 +109,7 @@ Any unknown keys should be treated as an error.
...
-| source                                   | String                                                      |                                                          | Who should get source credit. This would typically be the name (and year) of the event where the problem was first used or created for.
-| source_url                               | String                                                      |                                                          | Link to page for source event. Must not be given if source is not.
+| source                                   | Map of strings or sequence of map of strings                |                                                          | If specified, then the required key `name` should typically map to the name (and year) of the event where the problem was first used or for which it was created, and the key `url` should map to a link to the event's page. A sequence is specified in the special case that the first use was actually multiple (typically simultaneous) events.
```
niemela, line 112: Maybe this should be "described below" instead. That could be done in a follow up PR though.

niemela (approved):

spec/2023-07-draft.md
Tagl: Should we add examples there too as with credits?

niemela (approved):

spec/2023-07-draft.md
niemela: Maybe? I mean, examples is always good, right?

PR #184, opened 2024-04-15
Split our ICPC subset from legacy version

niemela:
- Creating new "version" `legacy-icpc`, i.e. the ICPC subset of `legacy`.
- Retroactively "adding" `problem_format_version` to `legacy`. It does not make any difference since it's not required and defaults `legacy` (and must be set to `legacy`), but it allow future users of `legacy` to explicitly do so.
- Retroactively "adding" `problem_format_version` to `legacy-icpc`. This should really be required, since the default is `legacy`, but doing so would be more confusing.
- Making  `problem_format_version` required in `2023-07-draft`.
- Cleaning up some tables.


RagnarGrootKoerkamp:
So in my understanding, we'll have:
- `legacy`
- `legacy-icpc`
- `2023-07`
- no more `2023-07-icpc`, since it's too undefined anyway? Good for me; fewer versions is simpler.

Adding `problem_format_version` sgtm.

niemela:
> So in my understanding [...]

Correct.

`legacy-icpc` is a little bit pointless, but I think it makes things a little bit simpler for those that prefer to use that subset.

Tagl (approved):

PR #193, opened 2024-04-17
Add example problem packages

JoelNiemela:
Creates the `examples/problems` directory, which contains a set of example problem packages.

Note that this is very much a work in progress. The `pass-fail` package is the most complete, followed by `scoring`. The rest are very incomplete.


niemela:
This is WIP, *but* it says so in the README. I think it will be easier to improve on this if we merge it so I will do so.

PR #195, opened 2024-04-19
Allow constants longer than one character

evouga:


ghamerly (approved):
Looks good to me.

Tagl (commented):
Should names be allowed to start with a digit, dash, or period?

niemela:
> Should names be allowed to start with a digit, dash, or period?

...or consist only of digits? I think no. 

Maybe we want the same as for filenames?: `[a-zA-Z0-9][a-zA-Z0-9_.-]*[a-zA-Z0-9]`

Tagl:
That would allow digits at start, I think: `[a-zA-Z_][a-zA-Z0-9_]*` would be more appropriate.

niemela:
> That would allow digits at start, 

Good point. I just copied what we have for files. Should *that* change? It feels a little bit strange that the requirements are so different?

> I think: `[a-zA-Z_][a-zA-Z0-9_]*` would be more appropriate.

That does sound better.

evouga:
No, I think it’s ok for file names to start with numbers since this is super useful for ordering test cases.

spec/2023-07-draft.md
```diff
@@ -302,7 +302,7 @@ File endings in parenthesis are not used for determining language.
 ### Constants
 
 A map of names to values.
-Names must match the following regex: `[a-zA-Z0-9_.-]`.
+Names must match the following regex: `[a-zA-Z0-9_.-]+`.
```
niemela, line 305:
```suggestion
Names must match the following regex: `[a-zA-Z_][a-zA-Z0-9_]*`.
```

niemela (changes requested):

niemela (approved):

PR #215, opened 2024-08-16
Clarify that types can't be repeated and must not be incompatible

### LINKED ISSUE #202: Use YAML, not space-separated string of keywords, to specify problem type
The new specification of `type` avoids the use of YAML as a specification language, which leads a convoluted definition, unnecessarily lenient values, arbitrary implementations, and difficulty of giving editor support.

(For instance, `2023-07-draft-2024-05-05` uses inconsistent symmetry in its specification of “is incompatible with”, and would allow `type: "scoring scoring scoring"`. Is `"interactive multi-pass"` meant to be allowed?) While we’re at it, we manage to use `pass` with two different meanings_ (“the testcase passes, the problem uses two passes”) in the same value string – `pass-fail multi-pass` (!) That can’t be good.

Consider using YAML for this, since we’re using YAML for many other parts of the configuration files already. Here are some examples:

```yaml
judgement: score
---
# empty, defaults to a single-pass, non-scoring problem
---
# same as above
judgement: verdict
submission: executable
passes: single
---
submission: answer # submit-answer
---
passes: interactive
---
judgement: score
submission: answer
--- 
judgement: score
passes: multiple
---
# FORBIDDEN
submission: answer
passes: multiple
```

The above can be specified neatly  in CUE.
```cue
judgment: *"verdict" | "score"
submission: *"executable" | "answer"
if submission == "executable" {
    passes: *"single" | "multiple" | "interactive"
}
```

or less neatly in JSON (I can’t be bothered right now.) Thanks to the magic of `schemastore.org` The JSON spec would then _automatically_ make every JSON-schema-aware editor in the world (such as VS Code or Emacs of VIM) check that people’s configuration files are correct (without typos and without contradictions). I cannot stress enough how useful that would be, in particular to new authors.

## Caveat

The _actual names_ for fields and values are unimportant. For instance, we can do

```cue
scoring: *false | true
executable: *true | false
if executable {
    passes: *"single" | "multiple" | "interactive"
}
```

Please don’t let your opinions about _what the fields should be called_ get in your way of evaluating this proposal. It’s about _using YAML_ instead of putting all of this into a string with some extra rules.

Also, I can see a case for `rounds` instead of `passes`, mainly to really make sure we avoid the class with `pass-fail`. 

Say,
```cue
rounds: *1 | "multiple" | "interactive"
```

I consider these details secondary; they only become worth talking about after moving away from `"pass-fail multi-pass     pass-fail"`.)
 
## Benefits

1. Avoids various name clashes, if key/value names are picked carefully
2. Supports problem author via schema-aware editor, avoiding ill-formed problem specifications
3. Trivial to parse for the tool (which already parses yaml)
4. Shorter and clearer to write down
5. Automates (using `cue vet` or JSON), so we can automatically verify problem specifications.

niemela:
Fixes #202

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -91,7 +91,8 @@ Any unknown keys should be treated as an error.
 
 ### Type
 
-Allowed values for type.
+Must be either a single string, or a sequence of strings, from the table below, with no repetition.
+Two values that are listed as incompatible must not both be in the sequence.
 
 | Value         | Incompatible with       | Comments
 | ------------- | ----------------------- | --------


sjoqvist (approved):
:shipit:

PR #223, opened 2024-08-16
Use randomly generated UUIDs in the example packages' problem.yaml

JoelNiemela:
This uses randomly generated UUIDs if we want to use that instead. If not, feel free to close this PR.


RagnarGrootKoerkamp (approved):

RagnarGrootKoerkamp:
We should re-generate the sample UUIDs when people clone these problems as a starting point. How do we prevent accidentally reusing the ones here?

PR #224, opened 2024-08-16
Advice on where to put generators

### LINKED ISSUE #216: Add place to store generators
We have decided to not directly support generators in the format, and I think that's the right decision.

That said, problem authors quite often use generators, and when e.g. debugging a problem it is very useful to have access to whatever generator(s) was used.

The most common place to store these seems to be in `generators/` (which is were we were going to put it when we almost added it to the spec), followed by `gen/`, `generator.*`, `gen.*`, and `generator/`. They can also sometimes be found in `xtra/`, `tools/`, or `other/`.

I suggest we specify that generators should be stored in `generators/`, without any further requirements on how they should work. Maybe we should also suggest that `generators/README.*` should describe how to use them?

evouga:
Fixes #216

Add recommendation to put generators and invocation instructions in `generators/`

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -433,10 +433,6 @@ output validator flags (see test data groups below)
 and the contents of their `.ans` files are equivalent, then the test cases are equivalent.
 The assumption of determinism means that a judge system could choose to reuse the result of a previous run, or to re-run the equivalent test case.
 
-## Generators
-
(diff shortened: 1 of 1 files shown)


evouga:
Addresses https://github.com/Kattis/problem-package-format/issues/216

niemela (approved):

niemela:
Fixes #216

spec/2023-07-draft.md
```diff
@@ -433,6 +433,10 @@ output validator flags (see test data groups below)
 and the contents of their `.ans` files are equivalent, then the test cases are equivalent.
 The assumption of determinism means that a judge system could choose to reuse the result of a previous run, or to re-run the equivalent test case.
 
+## Generators
```
eldering, line 436: I think this needs one extra level:
```suggestion
### Generators
```

evouga:
I think the heading level is OK, but the location is bad. I've moved it down below the Input Data section.

niemela (approved):

PR #229, opened 2024-08-16
Add archivists to credits map

### LINKED ISSUE #206: Add `packaged_by`
Sometime the creator of a problem package is neither the `author` nor the `rights_owner`. It would be good to be able to record who created a package in those cases.

There are two obvious choices for how to do this:
1. Add (somethign like) `packager` to `credits` in `problem.yaml`.
2. Add `packaged_by` at the top level of `problem.yaml`.

I'm leaning towards option 2. I would argue that this information is more like `rights_owner` in that it should not typically be shown with credits, but is useful when reusing and or debugging a problem in the future.

Thoughts?

niemela:
Fixes #206

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -176,11 +176,7 @@ credits:
   acknowledgements:
     - Inspirational Speaker 1
     - Inspirational Speaker 2
-<<<<<<< HEAD
   packagers:
-=======
-  archivists:
->>>>>>> 8513d47 (Add archivists to credits map)
     - Package Creatorson
 ```
 which demonstrates all the available credit types.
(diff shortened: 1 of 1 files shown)


JoelNiemela (approved):

spec/2023-07-draft.md
```diff
@@ -138,6 +138,7 @@ A person is specified as either a string with their full name (optionally with a
 | testers           | Person or sequence of persons          |         | The people who tested the problem package, for example by providing a solution and reviewing the statement.
 | translators       | Map of strings to sequences of persons |         | The people who translated the statement to other languages. Each key must be a language code as described in [General Requirements](#General Requirements "wikilink").
 | acknowledgements  | Person or sequence of persons          |         | Extra acknowledgements or special thanks that do not fit the other categories.
+| archivists        | Person or sequence of persons          |         | The people who created the problem package and do not fit the other categories.
```
Tagl, line 141: Now there are two types that "don't fit the other categories", might want to be a bit more specific here.

pehrsoderman: Should not "acknowledgements" be "in addition to the previously mentioned", and "archivists" "If different from the problem authors and contributors"

eldering: I'd change it to `packagers` and the description to:
The people who created this problem package out of an existing problem.

niemela: Updated.

pehrsoderman (approved):

Tagl (approved):

PR #233, opened 2024-08-17
Add expectations

### LINKED ISSUE #145: Expectations 0.9.0 (minimal viable specification); roadmap for 0.9.x
# Expectations 0.9.0

This framework allows problem author to express their expecations for the behaviour of a submission on the test data.

_New since 0.7.x:_ Vastly reduced specification to minimum viable specification; introduced roadmap for iterative future development. Added interactive example so people can play around with the syntax on their own.

#### Test Case Verdict

The behaviour of a submission on a _single_ test case is summarised in a *verdict*.
The verdicts are:

* `AC`: Accepted. The submission terminates successfully. Its execution time is strictly within the time limit. The output vaidator accepts the submission output.
* `WA`: The submission terminates successfully. Its execution time is strictly within the time limimt. The output validator rejects the submission output.
* `TLE`: The submission does not terminate in time stricly within the time limit.
* `RTE`: The submission aborts stricly within the time limit with a runtime error.

#### Common Expectations for a Submission

The expected behaviour of a submission on the test data often falls into a number of common classes, such as

* `accepted`: Every test case is `AC`.
* `wrong answer`: Every test case receives `AC` or `WA`;  _some_ test case receives `WA`.

More generally, an expectation consists of a set of _permitted_ verdicts (by default, _all_ verdicts) and set of _required_ verdicts (by default, _no_ verdicts).

* All test case must receive one of the permitted verdicts.
* Some test case must receive one of the required verdicts.

Thus, the common expectations above can be spelt out in terms of lists of verdicts.
For instance, the expectation that submission `mysubmission.cpp` is accepted is expressed like this:

```yaml
mysubmission.py: accepted
```

which is the same as

```yaml
mysubmission.py:
  permitted: [AC]
```
Similarly, 

```yaml
mysubmission.py: time limit exceeded
```

is the same as

```yaml
mysubmission.py:
  permitted: [AC, TLE]
  required: [TLE]
```

We use YAML, so the above can also be written as

```yaml
mysubmission.py:
  permitted:
    - AC
    - TLE
  required:
    - TLE
```

#### Specifying Expectations for Submissions

Expectations for submissions can be provided in a file `/submissions/expectations.yaml`.
A common tradition is specified like this:

```yaml
accepted: accepted
wrong_answer: wrong answer
time_limit_exceeded: time limit exceeded
runtime_exception: runtime exception
```
This would associate the expectation “accepted” with the submission `/submissions/accepted/mysubmission.cpp`.
The expectations framework is agnostic about directory names; for instance you can put your crashing submissions in `/submissions/run_time_error/` and put other requirements on the submissions in `/submissions/mixed/`:

```yaml
run_time_error: runtime exception
mixed:
  permitted: [AC, WA, TLE]
  required: [WA]
```

For problems with subtasks, you can put all your non-100% submissions into `/submissions/partially_accepted`, or you can introduce `/submissions/subtask1` or even `/submissions/brute_force_subtask` for this problem.

#### Specification per Submission

You can target individual submissions by providing their name:

```yaml
mixed/alice.py:
  permitted: [AC, WA, TLE]
  required: [WA]
mixed/bob.py:
  permitted: [AC, WA, TLE]
  required: [TLE]
```

#### Specification per Test Data

Top-level (“root”) expectations apply to all test data, but you can be more fine-grained and specify expectations for subdirectories of `/data`.
For instance, if you want all submission in `wrong_answer` to pass the sample inputs, you’d write:

```yaml
wrong_answer:
  sample: accepted
  secret: wrong answer
```

#### Terminology 

To fix terminology, a _testcase_ is uniquely identifed by its path, which is a string satisfying

```cue
#testcasepath: ~= "^(secret|sample)" & #path
```

A _testcase verdict_ is one of
```cue
#verdict: 'AC' | 'RTE' | 'WA' | 'TLE'
```

The _testcase result_ of running the submission on a testcase consists of 

* its `#verdict`.

#### Pattern matching

Both submissions and testcases are identified by paths.
The path of a submission is relative to `/submissions/`, possibly including a suffix like `.py`.
The path of a testcase is relative to `/data/`, not including a suffix like `.in`.
The _parents_ of a path are the logical ancestors in the sense of Python's `pathlib.PurePath.parents`;
the parents of submission `accepted/th.py` is the singleton `["accepted"]`;
the parents of testcase `secret/group3/032-random-4` are `["secret", "secret/group3"]`.
A _pattern_ can contain the Unix-style wildcard `*`; it cannot be empty.

```cue
#filename =~ [a-zA-Z0-9][a-zA-Z0-9_.-]*[a-zA-Z0-9] # given by specification
#path: =~ "[a-zA-Z0-9_./-]+" # nonempty
#pattern: =~ "[a-zA-Z0-9_./*-]+" # can contain *
```

A path `P` _matches_ a pattern `Q` if `P` or any of the parents of `P` match `Q` in the sense of Python's `fnmatch.fnmatch`.
(In particular, it is case-insensitive.)

#### Expectations

An _expectation_ consist of any of:

* a nonempty set `P` of _permitted verdicts_. If not specified, `P` is the set of all verdicts.
* a set `R`, possibly empty, of _required verdicts_. If not specified, `R` is the empty set.

An expectation is _satisfied_ by a set `S` of testcase results if

* the verdict of _every_ result in `S` belongs to `P`,
* if `R` is not empty then  some verdict in `R` appears as the verdict of _at least one_ result in `S`,

An expectation associated with a pattern is satisfied if all testcases matching the pattern satisfy the expectation.
A submission must satisfy all expectations that it matches.

#### Schema

```cue
#pattern: =~"[a-zA-Z0-9_./*-]+" // can contain *

#registry: close({[#pattern]: #root_expectation})

#verdict: "AC" | "WA" | "RTE" | "TLE"

#root_expectation: #abbreviation | {
    #expectation
    [=~"^(sample|secret|\\*)" & #pattern]: #abbreviation | #expectation
}

#expectation: {
    permitted?: [...#verdict] // only these verdicts may appear
    required?:  [...#verdict] // at least one of these verdicts must appear
    }

#abbreviation: "accepted" |  // { permitted: [AC] }
    "wrong answer" |         // { permitted: [AC, WA]; required: WA }
    "time limit exceeded" |  // { permitted: [AC, TLE]; required: TLE }
    "runtime exception" |    // { permitted: [AC, RTE]; required: RTE }
    "does not terminate" |   // { permitted: [AC, RTE, TLE]; required: [RTE, TLE] }
    "not accepted" |         // { required: [RTE, TLE, WA] }
    "rejected"               // { required: [RTE, WA] }
```

## Examples

```yaml
# Simple examples for some common cases

a.py: accepted            # AC on all cases
b.py: wrong answer        # at least one WA, otherwise AC
c.py: time limit exceeded # at least one TLE, otherwise AC
d.py: runtime exception   # at least one RTE, otherwise AC
e.py: does not terminate  # at least one RTE or TLE, but no WA
f.py: not accepted        # at least one RTE, TLE, or WA

# submission are identified by path or parent:

wrong_answer: wrong answer # expectations "wrong answer" apply to "wrong_answer/th.py" etc.

# can also use globbing. Because we use YAML, better surround those strings in quotes

"accepted/*.py": accepted # matches {accepted/th.py, accepted/ragnar.py}
"*/th.py" : accepted    # matches {accepted/th.py, wrong_answer/th.py}.

# Abbreviations are just shorthands for richer maps 
# of "required" and "permitted" keys.
#
# For instance, these are the same:

th.py: accepted
---
th.py:
  permitted: [AC]
  required: [AC]
---

# Specify that a submission is failed by the output validator on some testcase
# These are the same:

wrong.py: wrong answer
---
wrong.py:
  permitted: [WA, AC]
  required: [WA]
---
wrong.py:
  permitted: # alternative yaml syntax for list of strings
    - WA
    - AC
  required: [WA]
---

# Specify that the submission fails, but passes the samples.
# These are the same, using the same abbreviations as
# above for "accepted" and "wrong answer"

wrong.py:
  sample: accepted
  secret: wrong answer
---
wrong.py:
  sample: 
    permitted: [AC]
    required: [AC]
  secret:
    permitted: [AC, WA]
    required: [WA]

# Expectations apply to testcases whose parent matches the pattern
funky.cpp:
  permitted: [AC, WA, RTE]
  secret:
      permitted: [AC, RTE, TLE] # TLE is forbidden at parent, so this does not suddenly allow TLE here; tool should warn
  secret/small: accepted # more restrictive than ancestor: this is fine
   
# Specification for testcases works "all the way down to the single tescase"
funkier.cpp:
  secret/huge_instances/disconnected_graph:
      permitted: [RTE, TLE]
        
# Testcases can also be addressed using the unix-like “globbing” wildcard `*`

brute_force.py:
  "*/*-small-*" : accepted # matches secret/group3/032-small-disconnected
```

See https://cuelang.org/play/?id=baug8IzTJVU#cue@export@cue to play with the above example in an interactive, editable, web-based validation tool.

# Roadmap 

The specification is expected to evolve as follows:

### 0.9.1 Specification of Submission Directories

Specify which directories under `/submissions` have predefined semantics,
and whether this is 
1. implemented as a default `submissions/expectations.yaml` file with predefined entries such as `accepted: accepted` (and which therefore can be changed), or
2. part of the problem package format specification (and therefore does not appear in `expectations.yaml` and thus cannot be changed.)

#### 0.9.2 Judge message

Extend `testcase result` to include the contents of  `judgemessage.txt`
Add expectations for judge message:

```cue
##expectation: {
    ...
    message?: string // this must appear in some judgemessage, case-insensitive
}
```

Should this match by substring, and case-insensitive? (Or by glob, regex, etc.)?
Possible rule:

* possibly a string `message`, the _required judge message_

* if `message` is specified then it appears (as a substring) in the judge message of _at least one_ result in `S`,

#### 0.9.3 Richer globbing

Specify if `*` matches `x/y`.
Allow `**`, `[abc]` and `[^abc]` as per Python’s `fnmatch`. 
Specify `{th.py, ragnar.py}`.  

#### 0.9.4 Scoring

Extend `testcase result` to include the score of the testcase.

```cue
#expectation: {
    ...
    score?: ##range // all scores must be in range
}

#range: number | [number, number]

... ##abbreviation | #range | #expectation
```

* possibly a `score` range, which is an interval `(a,b)` with `0 <= a <= b`.

* if `score` is specified then the score \(s\) computed for _every_ testcase in `S` satisfies `a <= s <= b`.

This requires understanding what we mean by “the score of a testcase”. (In particular, has the multiplier from `testdata.yaml` been applied?)

#### 0.9.5 Aggregate results

```cue
#expectation: {
    ...
    aggregate_verdict?: ##verdict
    aggregate_score?: ##range
}
```

#### 0.9.6 Timing

Introduce the restricted verdicts `AC!` and `TLE!`, or invent some other formalism.
Until then, problem development tools are encouraged to emit warnings when `accepted` and `time limit exceeded` expectations are no satisfied with margin.

niemela:
Progress on #145 

Adds a slightly simplified versions of "expectations".

Does not include:
- Expectations on test cases or groups
- Richer globbing
- Expectations on overall verdict
- Ability to express what submissions should be used for timing

Also, explanation should be extended.

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -608,7 +608,6 @@ The possible subdirectories are:
 | Value               | Requirement                                                                                                        | Comment
 | ------------------- | ------------------------------------------------------------------------------------------------------------------ | -------
 | accepted            | Accepted as a correct solution for all test cases.                                                                 | At least one is required.
-| partially_accepted  | Overall verdict must be Accepted. Overall score must be less than `max_score`.                                     | Must not be used for pass-fail problems.
 | rejected            | Is rejected (i.e., not accepted) for any reason in the final verdict.                                              |
 | wrong_answer        | Wrong answer for some test case. May be too slow or crash for other test cases.                                    |
 | time_limit_exceeded | Too slow relative to the safety margin for some test case. May output wrong answers or crash for other test cases. |
@@ -628,6 +627,16 @@ Each glob maps to a map with keys as defined below, specifying metadata for all 
 | language   | String                        | As determined by file endings given in the [language list](#languages) |
 | entrypoint | String                        | As specified in the [language list](#languages)                        |
 | authors    | Person or sequence of persons |                                                                        | Author(s) of submission(s)
+| permitted  | Sequence of strings           | All (i.e. `[AC, WA, TLE, RTE]`)                                        | Values must be non-repeating values from `AC`, `WA`, `TLE`, `RTE`
+| required   | Sequence of strings           | All (i.e. `[AC, WA, TLE, RTE]`)                                        | Values must be non-repeating values from `AC`, `WA`, `TLE`, `RTE`
+| score      | Float                         |                                                                        |
+| message    | String                        | Empty string                                                           | This must appear (case-insensitive) in some `judgemessage.txt`
+
+It is an error if the submissions matched by the glob: 
+- gets a verdict not included in `permitted` for any test case.
+- does not get at least some verdict included in `required` for some test case.
+- does not get a final score of `score`.
+- does not include the string in `message` (case-insensitive) as a substring of the `judgemessage.txt` for some test case. 
 
 ## Input Validators
 


spec/2023-07-draft.md
```diff
@@ -628,6 +627,16 @@ Each glob maps to a map with keys as defined below, specifying metadata for all
 | language   | String                        | As determined by file endings given in the [language list](#languages) |
 | entrypoint | String                        | As specified in the [language list](#languages)                        |
 | authors    | Person or sequence of persons |                                                                        | Author(s) of submission(s)
+| permitted  | Sequence of strings           | All (i.e. `[AC, WA, TLE, RTE]`)                                        | Values must be non-repeating values from `AC`, `WA`, `TLE`, `RTE`
+| required   | Sequence of strings           | All (i.e. `[AC, WA, TLE, RTE]`)                                        | Values must be non-repeating values from `AC`, `WA`, `TLE`, `RTE`
```
JoelNiemela, line 631: Shouldn't "All" here be "Any"?

RagnarGrootKoerkamp:
Can we keep #145 open?

spec/2023-07-draft.md
```diff
@@ -608,7 +608,6 @@ The possible subdirectories are:
 | Value               | Requirement                                                                                                        | Comment
 | ------------------- | ------------------------------------------------------------------------------------------------------------------ | -------
 | accepted            | Accepted as a correct solution for all test cases.                                                                 | At least one is required.
-| partially_accepted  | Overall verdict must be Accepted. Overall score must be less than `max_score`.                                     | Must not be used for pass-fail problems.
 | rejected            | Is rejected (i.e., not accepted) for any reason in the final verdict.                                              |
```
evouga, line 611: Could we instead define rejected as "Is rejected (i.e., not accepted) for any reason by at least one test case."? Then we do not have to define what it means for a scoring problem to have an overall verdict.

spec/2023-07-draft.md
```diff
@@ -628,6 +627,16 @@ Each glob maps to a map with keys as defined below, specifying metadata for all
 | language   | String                        | As determined by file endings given in the [language list](#languages) |
 | entrypoint | String                        | As specified in the [language list](#languages)                        |
 | authors    | Person or sequence of persons |                                                                        | Author(s) of submission(s)
+| permitted  | Sequence of strings           | All (i.e. `[AC, WA, TLE, RTE]`)                                        | Values must be non-repeating values from `AC`, `WA`, `TLE`, `RTE`
+| required   | Sequence of strings           | All (i.e. `[AC, WA, TLE, RTE]`)                                        | Values must be non-repeating values from `AC`, `WA`, `TLE`, `RTE`
+| score      | Float                         |                                                                        |
```
evouga, line 632: What happens if you specify a score for a pass-fail problem?

niemela:
> Can we keep #145 open?

Sure. I have changed to "progress on".

I would prefer to replace #145 with tickets for the remaining parts. That said we could certainly do it in the order: merge this, create new tickets, close #145.

spec/2023-07-draft.md
niemela: I think `All` is good.

spec/2023-07-draft.md
niemela: That should either be an explicit error (i.e. you can't specify score unless the type includes `scoring`), or an implicit error because whatever score you require here, the submissions will not get that.

RagnarGrootKoerkamp:
I think we should stay closer to the original expectation PRs. We want to specify expectations for individual testgroups and testcases, not just at the root level.

niemela:
Merging. @RagnarGrootKoerkamp will make follow up PR adding more from #145

spec/2023-07-draft.md
eldering, line 631: It's very implicit now that `permitted` and `required` refer to verdicts on testcases. I think I'd be in favour of calling them `permitted_verdicts` and `required_verdicts`, but at least the comment should clarify this.

spec/2023-07-draft.md
eldering, line 632: Was the option to support a score range dropped on purpose?

spec/2023-07-draft.md
```diff
@@ -628,6 +627,16 @@ Each glob maps to a map with keys as defined below, specifying metadata for all
...
+| message    | String                        | Empty string                                                           | This must appear (case-insensitive) in some `judgemessage.txt`
+
+It is an error if the submissions matched by the glob: 
+- gets a verdict not included in `permitted` for any test case.
+- does not get at least some verdict included in `required` for some test case.
+- does not get a final score of `score`.
+- does not include the string in `message` (case-insensitive) as a substring of the `judgemessage.txt` for some test case.
```
eldering, line 639: I think it might read better if these were formulated in the non-negated sense:
```
Each submission matched by the glob must:
- gets a verdict included in `permitted` for each test case.
- get at least some verdict included in `required` for some test case.
- get a final score of `score` if `score` is specified.
- include the string in `message` (case-insensitive) as a substring of the `judgemessage.txt` for some test case. 
```

RagnarGrootKoerkamp:
@eldering your comments should be addressed in #249

Issue #207, opened 2024-06-12
Add compiler information


niemela:
While discussing the problem format at NAC24 @evouga pointed out that it would often be useful to have more exact information about what exact compiler and language versions are expected/required for a problem. E.g. if a problem package uses feature from C++20 then it will not be possible to install it on a system that only offers C++17.

OTOH, for portability we don't want to *force* any exact versions for using a package, so I think this needs to be "informational".

Another challenge is that there is no universal way to describe what specific compiler/interpreter is being used (as far as I know). There is only the name, which is probably unique most of the time, at least if it's something popular.

We solved kinda the same problem in the [contest API](https://ccs-specs.icpc.io/draft/contest_api#languages) by specifying:
- `command`: Command to run
- `args`: Argument list for command. `{files}` denotes where to include the file list. Defaults to empty string.
- `version`: Expected output from running the version command. Defaults to empty string.
- `version_command`: Command to run to get the version. Defaults to `<command> --version`.

I think we should piggyback on this solution (and if we realize a way to improve on it, we should push that back upstream), since it really is solving the same problem, and since this means that contest API compliant systems (i.e. at least Kattis, DOMjudge, PC^2) would be able to easily produce this information.

Then the question is where to put this information. I see two main options:

Either, add a `languages` key to `prohblem.yaml` containing this info for languages used. That could look something like this:
```yaml
languages:
   cpp:
      command: gcc
      args: -O2 -Wall -static {files}
      version: gcc (Ubuntu 8.3.0-6ubuntu1) 8.3.0
   java:
      command: javac
      args: -O {files}
      version: javac 11.0.4
```
We might want to add both a compiler and runner, as we did with the contest API.

Note that this as written right now does *not* work, since we already use `languages` for something else. It might be possible and even suitable to combine them though?

Alternatively, if we are of the opinion that this is mostly (only?) about submissions, we could add the same information to `submissions.yaml`.

I'm leaning towards the first option.

Thouhgts?

Tagl:
I think adding information like C++20 vs C++14 or JDK8 vs JDK17 is definitely something useful to add.
These are even different programming languages, since they aren't always backwards compatible.
Compiler version is fine as informational, since there are cases where compiler intrinsics are used, but I think the important thing to note is what standard of the language is being used.
There are many compilers which support the same standard, does it matter that much whether it is clang or gcc, or which version of those, if it supports c++20?

Tagl:
[...]

PR #240, opened 2024-08-17
Metadata about submission language versions

evouga:
Resolves https://github.com/Kattis/problem-package-format/issues/207

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -617,8 +617,37 @@ Every file or directory in these directories represents a separate solution.
 It is mandatory to provide at least one accepted solution.
 
 Metadata about the example submissions are provided in a YAML file named `submissions.yaml` placed directly in the `submissions/` directory.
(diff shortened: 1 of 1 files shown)


JoelNiemela (commented):
Do we want to put this in `submissions.yaml`? `language_versions` might be used to describe which language version e.g. an output validator is written in.

evouga:
Hmm good point, maybe we want this in problem.yaml instead?

niemela:
> Hmm good point, maybe we want this in problem.yaml instead?

I think so.

This will certainly be *mostly* be useful for submissions, but *technically* it should really apply to all programs.

So... I'm a little bit torn, because it *almost* fits perfectly in `submission.yaml`... but I think it *should* be in `problem.yaml`.

evouga:
Now in problem.yaml

niemela (approved):

pehrsoderman:
Needs a rebase.

PR #249, opened 2024-08-17
Expectations, part 2

### LINKED ISSUE #145: Expectations 0.9.0 (minimal viable specification); roadmap for 0.9.x

### LINKED ISSUE #234: Notes from verdict aggregation discussion
For pass-fail problems:
- Support testing `permitted` and `required` verdicts of sets of testcases (a single testcase, or leaves under a group).
- No testgroup/final verdict aggregation -- only permitted/required on sets of testcases can be asserted.
- **Question:** Do we indeed not need final verdict assertions?

For scoring problems:
- No verdict aggregation.
- Allow asserting permitted/required vertices of (sets of) test cases.
- Allow asserting scores of testcases/testgroups/final score.

RagnarGrootKoerkamp:
I _think_ this includes everything we discussed, but surely some subtleties are still missing.

It would be good to add some more examples, and we should also put the formal cue spec somewhere.

This merges most of what is proposed in #145. Most notably, it does not (yet?) stabilize the 'abbreviations' `submission.py: accepted`. And the other difference is that we merge this into `submissions.yaml` instead of a dedicated `expectations.yaml`.

Also note that to preserve backwards compatibility with the previous definitions of the pre-defined directories, an `expectations.yaml` can be added that relaxes the meaning of `wrong_answer/`, `time_limit_exceeded/`, and `run_time_error/` to simply `permit` all verdicts. But this is not mentioned as such in this version of the spec itself.

Closes #234 closes #145

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -610,12 +610,14 @@ that file must have the language-dependent name as given in the table referred a
 
 ## Example Submissions
 
-Correct and incorrect solutions to the problem are provided in subdirectories of `submissions/`.
+Correct and incorrect solutions (file or directory programs) to the problem are
+provided in direct subdirectories of `submissions/`. That is
 By default, the possible subdirectories are as in the table below, but they can
-be customized, see [Default directories](#default-directories).
+be customized and more can be added, see [Default directories](#default-directories).
+Submissions directly in `submissions/`, or two (or more) subdirectories inside it are ignored.
 
 
-| Value                 | Requirement                                                                           | Comment                                                       |
+| Directory             | Requirement                                                                           | Comment                                                       |
 |-----------------------|---------------------------------------------------------------------------------------|---------------------------------------------------------------|
 | `accepted`            | Accepted as a correct solution for all test cases.                                    | At least one is required. Used to lower bound the time limit. |
 | `rejected`            | At least one case is not accepted.                                                    |                                                               |
@@ -627,11 +629,12 @@ Every file or directory in these directories represents a separate solution.
 It is mandatory to provide at least one accepted solution.
 
 Metadata about the example submissions are provided in a YAML file `submissions/submissions.yaml`.
-The top level keys in `submissions.yaml` are globs matching files or directories in `submissions/`.
+The top level keys in `submissions.yaml` are glob patterns matching files or directories under `submissions/`.
 For example, `accepted` and `accepted/*` match all submissions in the `submissions/accepted/` directory.
-See also [Globs](#globs)
+See also [Glob-patterns](#glob-patterns)
 
-Each glob maps to a map with keys as defined below, specifying metadata for all submissions that are matched by the glob.
+Each glob pattern maps to a map with keys as defined below, specifying metadata for all
+submissions that are matched by the glob pattern.
 
 | Key                  | Type                             | Default                                                                              | Comment                                                                                                   |
 |----------------------|----------------------------------|--------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------|
@@ -644,7 +647,7 @@ Each glob maps to a map with keys as defined below, specifying metadata for all 
 | `message`            | String                           | Empty string                                                                         | This must appear as a substring in at least one `judgemessage.txt`.                                       |
 | `use_for_time_limit` | Bool or string (`lower`/`upper`) | See below.                                                                           | Controls whether this submission is used to determine the time limit.                                     |
 
-Every submission matched by the glob must satisfy:
+Every submission matched by the glob pattern must satisfy:
 - all test cases must have only verdicts present in `permitted`;
 - at least one test case must have a verdict in `required`;
 - if given, the score must be in the given range or equal the given score;
@@ -676,10 +679,10 @@ solves_group_1.py:
     required: TLE
 ```
 
-#### Globs
+#### Glob patterns
 
-Globs can be used to apply restrictions to a subset of submissions.
-It is also possible to use globs to put restrictions on a subset of test
+Glob patterns can be used to apply restrictions to a subset of submissions.
+It is also possible to use glob patterns to put restrictions on a subset of test
 cases, for example when test groups are not used:
 ```yaml
 time_limit_exceeded/solves_easy_cases.py:
@@ -694,11 +697,11 @@ time_limit_exceeded/solves_easy_cases.py:
 This means that the submission must solve all samples and all easy cases, but
 must time out on at least one of the hard cases.
 
-Submission globs are matched against all paths to files and directories of
+Submission glob patterns are matched against all paths to files and directories of
 submissions inside and relative to the `submissions/` directory.
-Test case globs are matched against all paths of test groups and test cases relative to
+Test case glob patterns are matched against all paths of test groups and test cases relative to
 `submissions` and `data/` respectively (, excluding the trailing `.in`. Wildcards (`*`) only match within a file
-name (e.g., do not match `/`). A test case is matched by the glob if either
+name (e.g., do not match `/`). A test case is matched by the glo patternb if either
 itself or any of its parent test groups is matched by it, and similarly a
 submission is matched if either itself or a parent directory is matched.
 
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -624,6 +624,7 @@ Submissions directly in `submissions/`, or two (or more) subdirectories inside i
 | `wrong_answer`        | At least one case is wrong answer, and all cases are either wrong answer or accepted. | Used to lower bound the time limit.                           |
 | `time_limit_exceeded` | Too slow on at least one case, and all cases are either too slow or accepted.         | Used to upper bound the time limit.                           |
 | `run_time_error`      | Crashes for at least one case, and all cases either crash or are accepted.            | Used to lower bound the time limit.                           |
+| `brute_force`         | Never gives the wrong answer, but not accepted because run time error or timeout.     |                                                               |
 
 Every file or directory in these directories represents a separate solution.
 It is mandatory to provide at least one accepted solution.
@@ -733,7 +734,7 @@ run_time_error:
   required: RTE
 # Must not WA, but fail at least once.
 # Note that by default these are not used for determining the time limit.
-bruteforce:
+brute_force:
   permitted: [AC, RTE, TLE]
   required: [RTE, TLE]
 ```
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -619,30 +619,30 @@ be customized, see [Default directories](#default-directories).
 |-----------------------|---------------------------------------------------------------------------------------|---------------------------------------------------------------|
 | `accepted`            | Accepted as a correct solution for all test cases.                                    | At least one is required. Used to lower bound the time limit. |
 | `rejected`            | At least one case is not accepted.                                                    |                                                               |
-| `wrong_answer`        | At least one case is wrong answer, and all cases are either wrong answer or accepted. |                                                               |
+| `wrong_answer`        | At least one case is wrong answer, and all cases are either wrong answer or accepted. | Used to lower bound the time limit.                           |
 | `time_limit_exceeded` | Too slow on at least one case, and all cases are either too slow or accepted.         | Used to upper bound the time limit.                           |
-| `run_time_error`      | Crashes for at least one case, and all cases either crash or are accepted.            | Very rarely useful.                                           |
+| `run_time_error`      | Crashes for at least one case, and all cases either crash or are accepted.            | Used to lower bound the time limit.                           |
 
 Every file or directory in these directories represents a separate solution.
 It is mandatory to provide at least one accepted solution.
 
 Metadata about the example submissions are provided in a YAML file `submissions/submissions.yaml`.
 The top level keys in `submissions.yaml` are globs matching files or directories in `submissions/`.
-For example, `accepted` and `accepted/*` match all submission in the `submissions/accepted/` directory.
+For example, `accepted` and `accepted/*` match all submissions in the `submissions/accepted/` directory.
 See also [Globs](#globs)
 
 Each glob maps to a map with keys as defined below, specifying metadata for all submissions that are matched by the glob.
 
-| Key                  | Type                          | Default                                                                              | Comment                                                                                                   |
-|----------------------|-------------------------------|--------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------|
-| `language`           | String                        | As determined by file endings given in the [language list](../appendix/languages.md) |                                                                                                           |
-| `entrypoint`         | String                        | As specified in the [language list](../appendix/languages.md)                        |                                                                                                           |
-| `authors`            | Person or sequence of persons |                                                                                      | Author(s) of submission(s).                                                                               |
-| `permitted`          | Sequence of strings           | `[AC, WA, TLE, RTE]`                                                                 | All test cases must have a verdict in this subset of `AC`, `WA`, `TLE`, `RTE`.                            |
-| `required`           | Sequence of strings           | `[AC, WA, TLE, RTE]`                                                                 | At least one test case must have a verdict in this subset of `AC`, `WA`, `TLE`, `RTE`.                    |
-| `score`              | Float or list of two floats   |                                                                                      | The score of the submission equals the given number, or is in the given range. Only for scoring problems. |
-| `message`            | String                        | Empty string                                                                         | This must appear as a substring in at least one `judgemessage.txt`.                                       |
-| `use_for_time_limit` | Bool                          | See below.                                                                           | Controls whether this submission is used to determine the time limit.                                     |
+| Key                  | Type                             | Default                                                                              | Comment                                                                                                   |
+|----------------------|----------------------------------|--------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------|
+| `language`           | String                           | As determined by file endings given in the [language list](../appendix/languages.md) |                                                                                                           |
+| `entrypoint`         | String                           | As specified in the [language list](../appendix/languages.md)                        |                                                                                                           |
+| `authors`            | Person or sequence of persons    |                                                                                      | Author(s) of submission(s).                                                                               |
+| `permitted`          | Sequence of strings              | `[AC, WA, TLE, RTE]`                                                                 | All test cases must have a verdict in this subset of `AC`, `WA`, `TLE`, `RTE`.                            |
+| `required`           | Sequence of strings              | `[AC, WA, TLE, RTE]`                                                                 | At least one test case must have a verdict in this subset of `AC`, `WA`, `TLE`, `RTE`.                    |
+| `score`              | Float or list of two floats      |                                                                                      | The score of the submission equals the given number, or is in the given range. Only for scoring problems. |
+| `message`            | String                           | Empty string                                                                         | This must appear as a substring in at least one `judgemessage.txt`.                                       |
+| `use_for_time_limit` | Bool or string (`lower`/`upper`) | See below.                                                                           | Controls whether this submission is used to determine the time limit.                                     |
 
 Every submission matched by the glob must satisfy:
 - all test cases must have only verdicts present in `permitted`;
@@ -651,7 +651,7 @@ Every submission matched by the glob must satisfy:
 - if given, the `message` string must be included as a case-sensitive substring in the `judgemessage.txt` for at least one test case. 
 
 The tooling should check the constraints for consistency, such as that two
-disjoint `permitted` sets are applied to a single `(submission, testcase)` pair.
+disjoint `permitted` sets are never applied to a single `(submission, testcase)` pair.
 
 ### Groups
 The `permitted`, `required`, `score`, `message`, and `use_for_time_limit`
@@ -682,7 +682,7 @@ Globs can be used to apply restrictions to a subset of submissions.
 It is also possible to use globs to put restrictions on a subset of test
 cases, for example when test groups are not used:
 ```yaml
-time_lmit_exceeded/solves_easy_cases.py:
+time_limit_exceeded/solves_easy_cases.py:
   sample:
     permitted: AC
   secret/*-easy:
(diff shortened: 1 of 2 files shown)


spec/2023-07-draft.md
```diff
@@ -606,40 +606,159 @@ that file must have the language-dependent name as given in the table referred a
...
+
+| Value               | Requirement                                                                           | Comment                                                   |
+|---------------------|---------------------------------------------------------------------------------------|-----------------------------------------------------------|
+| accepted            | Accepted as a correct solution for all test cases.                                    | At least one is required. Used to lower bound time limit. |
+| rejected            | At least one case is not accepted.                                                    |                                                           |
+| wrong_answer        | At least one case is wrong answer, and all cases are either wrong answer or accepted. |                                                           |
+| time_limit_exceeded | Too slow on at least one case, and all cases are either too slow or accepted.         | Used to upper bound time limit.                           |
```
niemela, line 618: I still would like to allow WA, but if nobody agrees with me, so be it.

spec/2023-07-draft.md
```diff
@@ -606,40 +606,159 @@ that file must have the language-dependent name as given in the table referred a
...
+| required           | String or sequence of strings | `[AC, WA, TLE, RTE]`                                                   | At least one test case must have a verdict is shit subset of `AC`, `WA`, `TLE`, `RTE`. Must be a subset of `permitted`. |
+| score              | Float or list of two floats   |                                                                        | The score of the submission equals the given number, or is in the given range. Only for scoring problems.               |
+| message            | String                        | Empty string                                                           | This must appear as a substring in at least one `judgemessage.txt`.                                                     |
+| use_for_time_limit | Bool                          | See below.                                                             | Controls whether this submission is used to determine the time limit.                                                   |
+
+Every submission matched by the glob must:
+- only have verdicts present in `permitted`;
```
niemela, line 644: "all testcases must have verdicts from `permitted`".

spec/2023-07-draft.md
```diff
@@ -606,40 +606,159 @@ that file must have the language-dependent name as given in the table referred a
...
+- get at least one verdict in `required`;
```
niemela, line 645: "at least some testcase must have a verdict from `required`"

spec/2023-07-draft.md
```diff
@@ -606,40 +606,159 @@ that file must have the language-dependent name as given in the table referred a
...
+Submission globs are matched against all paths to files and directories of
+submissions inside and relative to the `submissions/` directory.
+Test case globs are matched against all paths of test groups and test cases relative to
+`submissions` and `data/` respectively (, excluding the trailing `.in`. Wildcards (`*`) only match within a file
+name (e.g., do not match `/`). A test case is matched by the glob if either
+itself or any of its parent test groups is matched by it, and similarly a
+submission is matched if either itself or a parent directory is matched.
+
+Using `**` to match any number of directories and `[xyz]` to match only a subset
+of characters is no supported.
```
niemela, line 699:
```suggestion
of characters is not supported.
```

spec/2023-07-draft.md
```diff
@@ -606,40 +606,159 @@ that file must have the language-dependent name as given in the table referred a
...
+
+TODO: Did we allow `{a,b}.py` in the end?
```
niemela, line 701: It would be nice.

spec/2023-07-draft.md
```diff
@@ -606,40 +606,159 @@ that file must have the language-dependent name as given in the table referred a
...
+  permitted: [AC, TLE]
+  required: TLE
+# All cases AC or RTE, at least one RTE.
+run_time_error:
+  permitted: [AC, RTE]
+  required: RTE
+# Must not WA, but fail at least once. TODO: name up for debate.
+bruteforce:
+  permitted: [AC, RTE, TLE]
+  required: [RTE, TLE]
```
niemela, line 727: I think this should be allowed to be `AC`.

spec/2023-07-draft.md
```diff
@@ -606,40 +606,159 @@ that file must have the language-dependent name as given in the table referred a
...
+```
+
+These requirements may be too strict. To e.g. allow TLE
```
niemela, line 730:
```suggestion
To e.g. allow TLE
```

spec/2023-07-draft.md
```diff
@@ -606,40 +606,159 @@ that file must have the language-dependent name as given in the table referred a
...
+- include the `message` string as a case-sensitive substring in the `judgemessage.txt` for at least one test case. 
+
+### Groups
+The `permitted`, `required`, `score`, `message`, and `use_for_time_limit`
+requirements can also be given for only a subset of test cases, by adding them
+under a key with the name of a test group (relative to `data/`). In this case,
+the `permitted`, `required`, `message`, and `use_for_time_limit` keys only apply
+to the set of test cases (recursively) in the given group.
+
+The `score` key puts a constraint of the aggregated score of a given test group,
```
eldering, line 656:
```suggestion
The `score` key puts a constraint on the aggregated score of a given test group,
```

spec/2023-07-draft.md
eldering, line 618:
```suggestion
| accepted            | Accepted as a correct solution for all test cases.                                    | At least one is required. Used to lower bound the time limit. |
| rejected            | At least one case is not accepted.                                                    |                                                           |
| wrong_answer        | At least one case is wrong answer, and all cases are either wrong answer or accepted. |                                                           |
| time_limit_exceeded | Too slow on at least one case, and all cases are either too slow or accepted.         | Used to upper bound the time limit.                       |
```

spec/2023-07-draft.md
RagnarGrootKoerkamp: as discussed, we leave it as is. Tooling can suggest to move to `accepted/` instead.

spec/2023-07-draft.md
```diff
@@ -606,40 +606,168 @@ that file must have the language-dependent name as given in the table referred a
...
+    permitted: [AC, TLE]
+    required: TLE
+```
+This means that the submission must solve all samples and all easy cases, but
+must time out on at least one of the hard cases.
+
+Submission globs are matched against all paths to files and directories of
+submissions inside and relative to the `submissions/` directory.
+Test case globs are matched against all paths of test groups and test cases relative to
+`submissions` and `data/` respectively (, excluding the trailing `.in`. Wildcards (`*`) only match within a file
```
eldering, line 695:
```suggestion
`data/` (, excluding the trailing `.in`. Wildcards (`*`) only match within a file
```

spec/2023-07-draft.md
```diff
@@ -606,40 +606,168 @@ that file must have the language-dependent name as given in the table referred a
...
+
+### Groups
+The `permitted`, `required`, `score`, `message`, and `use_for_time_limit`
+requirements can also be given for only a subset of test cases, by adding them
+under a key with the name of a test group (relative to `data/`). In this case,
+the `permitted`, `required`, `message`, and `use_for_time_limit` keys only apply
+to the set of test cases (recursively) in the given group.
+
+The `score` key puts a constraint on the aggregated score of a given test group,
+_not_ on the _set_ of test cases the group contains.
```
eldering, line 659: This doesn't specify how a score should be aggregated on a set of test cases (which is not exactly a test group).

spec/2023-07-draft.md
```diff
@@ -606,40 +606,168 @@ that file must have the language-dependent name as given in the table referred a
...
+# Note that by default these are not used for determining the time limit.
+bruteforce:
+  permitted: [AC, RTE, TLE]
+  required: [RTE, TLE]
+```
+
+The defaults can be overwritten in the `submissions.yaml` file by simply specifying the name of the directory.
+Keys that are not specified are inherited from the default configuration above.
+(This is mostly supported for backwards compatibility, and not recommended for
+normal usage.)
```
eldering, line 736:
```suggestion
This is supported for backwards compatibility, and not recommended for normal usage.
```

spec/2023-07-draft.md
```diff
@@ -606,40 +606,168 @@ that file must have the language-dependent name as given in the table referred a
...
+
+```yaml
+time_limit_exceeded:
+  permitted: [AC, WA, TLE]
+  required: TLE
+```
+
+(Note that using `time_limit_exceeded/*` instead would impose an _additional_
+requirement, instead of _replacing_ the original requirement.)
```
eldering, line 745:
```suggestion
Note that the glob `time_limit_exceeded/*` would impose an _additional_
requirement, instead of _replacing_ the original requirement.
```

spec/2023-07-draft.md
```diff
@@ -606,40 +606,168 @@ that file must have the language-dependent name as given in the table referred a
...
+
+### Timelimit inference
+Any submission that must satisfy a `required: TLE` requirement, i.e., must `TLE`
+on at least one test case, is used to provide an _upper bound_ on the time limit.
+Precisely, the time limit must be at most `T / time_limit_to_tle`, where `T` is
+the slowest runtime over the set of test cases to which the rule applies.
+(Note that this excludes submissions that e.g. have `required: [TLE, RTE]`.)
```
eldering, line 752:
```suggestion
Note that this excludes submissions that e.g. have `required: [TLE, RTE]`.
```

spec/2023-07-draft.md
```diff
@@ -606,40 +606,168 @@ that file must have the language-dependent name as given in the table referred a
...
+
+Any submission that is not permitted to get `TLE` at all (on some subset of cases), i.e., must satisfy a
+`permitted:` rule that does not contain `TLE`, is used to provide a _lower
+bound_ on the time limit. Precisely, the time limit must be at least `T *
+ac_to_time_limit`, where `T` is the slowest runtime over the set of test cases to
+which the rule applies.
+
+To _opt out_ of a (set of) submission(s) from influencing the time limit, set
+`use_for_time_limit: false` alongside the `permitted:` or `required:` key that
+satisfies the constraints above.
```
eldering, line 762:
```suggestion
To _opt out_ of a glob for submission(s) and optional subset of testcases from influencing the time limit, set
`use_for_time_limit: false` alongside the `permitted:` and/or `required:` keys.
Note that this means that if you want to exclude a submission completely,
then you must add `use_for_time_limit: false` to every glob that matches that submission.
```

spec/2023-07-draft.md
```diff
@@ -606,40 +606,168 @@ that file must have the language-dependent name as given in the table referred a
...
+
+To explicitly _opt in_ a (set of) submissions(s) to be used for determining the time limit,
+use `use_for_time_limit: lower` and `use_for_time_limit: upper`.
+The first is equivalent to a `permitted: [AC, WA, RTE]` constraint, and the
+second to a `required: [TLE]` constraint. The system may warn when this makes
+other constraints redundant.
```
eldering, line 768:
```suggestion
other constraints redundant and should error when it is inconsistent with other constraints.
```

spec/2023-07-draft.md
```diff
@@ -606,40 +606,174 @@ that file must have the language-dependent name as given in the table referred a
...
+|-----------------------|---------------------------------------------------------------------------------------|---------------------------------------------------------------|
+| `accepted`            | Accepted as a correct solution for all test cases.                                    | At least one is required. Used to lower bound the time limit. |
+| `rejected`            | At least one case is not accepted.                                                    |                                                               |
+| `wrong_answer`        | At least one case is wrong answer, and all cases are either wrong answer or accepted. | Used to lower bound the time limit.                           |
+| `time_limit_exceeded` | Too slow on at least one case, and all cases are either too slow or accepted.         | Used to upper bound the time limit.                           |
+| `run_time_error`      | Crashes for at least one case, and all cases either crash or are accepted.            | Used to lower bound the time limit.                           |
```
JoelNiemela, line 619: Shouldn't `bruteforce` be added to this table? (See "Default directories")

JoelNiemela (changes requested):

RagnarGrootKoerkamp:
merge?

niemela (approved):

Issue #138, opened 2023-12-01
Add visualizers


niemela:
*Reply to a comment in #5*
> This is mostly for [...] and visualizers

@RagnarGrootKoerkamp :
"visualizers" is a BAPCtools extension, right?

If yes, could you link to or describe how they are specified there?

niemela:
> or alternatively passing it on `stdin`.

That would be more similar to how the output validator works, which I think is a good thing.

In fact, an obvious alternate design would be to have the output validator be this program, by allowing it to create these files in the feedback directory. The only downside I can think of right away is that this makes it a lot less clear whether there *is* a visualizer or not. I assume that this is something that you have thought about, are there other reasons why that would be worse?

Tagl:
I tend to do
```
./visualizer testcase.ans testcase.png <  testcase.in
```
to visualize testcases as I generate them.
I never considered doing it for team submissions.
I think that's a great idea, but it is a bit of different thing since then you can't use the feedback directory directly.
I would then do it like
```
./visualizer testcase.in testcase.ans testcase.png < testcase.out
```
so I could call it easily when generating and it's easy to specify the image file into the feedback folder.

So I guess another downside is that if this is the output validator it's not as useful, or at least easy to use, when generating data files, I think.

niemela:
> I think that's a great idea, but it is a bit of different thing since then you can't use the feedback directory directly.

Sorry, why can't you use the feedback directory directly?

Tagl:
> Sorry, why can't you use the feedback directory directly?

When generating a testcase, would you specify `secret/group1` as the feedback directory then? Or some other directory? I could move the image file from the feedback directory to the appropriate group I guess.

niemela:
> When generating a testcase, would you specify `secret/group1` as the feedback directory then? Or some other directory? I could move the image file from the feedback directory to the appropriate group I guess.

You could do either of those things. We don't specify where a the feedback directory should be, only that it must exist and be provided to the validator.

RagnarGrootKoerkamp:
I don't like specifying testcase.png as an argument, since that requires
knowing the right extension. We often mix SVG and PNG.

niemela:
@RagnarGrootKoerkamp or @Tagl Would you be able to write up a draft PR for this?

RagnarGrootKoerkamp:
For output, I prefer if they write to some directory, and any file extension types recognised by the tooling are taken as visualizations.

So some options:
[...]

PR #253, opened 2024-08-17
Visualizer

evouga:
Fixes https://github.com/Kattis/problem-package-format/issues/138

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -33,7 +33,7 @@ used for distributing and sharing problems for algorithmic programming contests 
 
 ### Programs
 
-There are a number of different kinds of programs that may be provided in the problem package: submissions, input validators, output validators, and visualizers.
+There are a number of different kinds of programs that may be provided in the problem package: submissions, input validators, output validators, and output visualizers.
 All programs are always represented by a single file or directory.
 In other words, if a program consists of several files, these must be provided in a single directory.
 In the case that a program is a single file, it is treated as if a directory with the same name takes its place, which contains only that file.
@@ -65,7 +65,7 @@ The binary (and other artifacts that result from compiling the program) must be 
 Each submission must be run with a working directory that contains (a copy of) the submitted files and any compiled binaries,
 but it must not contain any of the files described in the "Test data" section.
 Each input validator must be run with a working directory that contains the files in the program directory of the input validator in question.
-Each output validator and visualizer must be run with a working directory that contains the submitted files and any compiled binaries of the submission being validated.
+Each output validator and output visualizer must be run with a working directory that contains the submitted files and any compiled binaries of the submission being validated.
 
 ## Problem Metadata
 
@@ -588,6 +588,10 @@ The directory `invalid` can be organized into a tree-like structure similar to `
 
 If any generator scripts were used to automate writing test cases, it is recommended to include the generator source code in the directory `generators/` along with invocation instructions in a file such as `generators/README.txt`. This information can be useful as a debugging aid and for archival completeness: judge systems are not responsible for executing the provided generators and all test data written by the generators must be included in the problem package.
 
+## Input Visualizer
+
+If a tool was used to automate creating test case illustration annotations, it is recommended to include the input visualizer source code in the directory `input_visualizer/` along with invocation instructions in a file such as `input_visualizer/README.txt`. 
+
 ## Included Files
 
 Files that should be included with all submissions are provided in one non-empty directory per supported language.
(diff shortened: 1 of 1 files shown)


spec/2023-07-draft.md
```diff
@@ -801,6 +801,10 @@ A judging system that implements this format must support the `judgemessage.txt`
 (I.e., content of the `judgemessage.txt` file, if produced by the validator, must be provided by the judging system to a human judge examining the submission).
 Having the judging system support other files is optional.
 
+The validator may create one or more image files in the feedback directory with name `teamimage.ext` and/or `judgeimage.ext`, where `ext` is an appropriate image file format extension.
```
pehrsoderman, line 804: What are the allowed extensions?

RagnarGrootKoerkamp:
@mzuenni @mpsijm see also the last message in the referenced issue. This is now speccing output visualisers, but probably we should also keep input visualisers in some way

spec/2023-07-draft.md
evouga: Do we need to specify, if we are not requiring judge systems to support any particular formats?

niemela: We specify supported image file formats in many places. I think it makes sense to do so here as well.

Currently we allow:
- `.png`, `.jpg`/`.jpeg`, `.pdf` for problem statement images.
- `.png`, `.jpg`/`.jpeg`, `.svg` for annotations.

Output visualisers images are basically the equivalent for "annotations", but for outputs (and nit inputs), so I think it makes sense to allow the exact same set of formats.

niemela (changes requested):
Specify allowed image formats (`.png`, `.jpg`/`.jpeg`, `.svg`)

niemela (approved):

niemela (approved):

Tagl:
> JPEG images group together the following filename extensions:
> 
>     .jpg
>     .jpeg
>     .jpe
>     .jif
>     .jfif
>     .jfi

Of course it isn't easy

eldering:
> > JPEG images group together the following filename extensions:
> > ```
> > .jpg
> > .jpeg
> > .jpe
> > .jif
> > .jfif
> > .jfi
> > ```
> 
> Of course it isn't easy

I agree with Fredrik that just `.jpg` and `.jpeg` are needed.

Tagl (approved):

spec/2023-07-draft.md
```diff
@@ -848,6 +856,20 @@ Almost all test cases failed — are you even trying to solve the problem?
 A validator program is allowed to write any kind of debug information to its standard error pipe.
 This information may be displayed to the user upon invocation of the validator.
 
+### Output Visualizer
+
+An output visualizer is an optional [program](#programs) that is run after every invocation of the output validator in order to generate images illustrating the submission output. 
+A visualizer program must be an application (executable or interpreted) capable of being invoked with a command line call. It is invoked using the same arguments as the output validator. 
+It must be provided as the directory `output_visualizer/`.
+It must either be written in C++ or Python 3,
```
RagnarGrootKoerkamp, line 864: Looking over recent problems, we have used
- `asymptote`, some very specific graphic drawing language
- `matplotlib`
- a custom c++ lib that writes svg files
- a custom python function to write pngs.

Restricting to python/c++ by default sgtm, although these may want to depend on libs that need to be installed by the system.

spec/2023-07-draft.md
```diff
@@ -801,6 +805,10 @@ A judging system that implements this format must support the `judgemessage.txt`
 (I.e., content of the `judgemessage.txt` file, if produced by the validator, must be provided by the judging system to a human judge examining the submission).
 Having the judging system support other files is optional.
 
+The validator may create one or more image files in the feedback directory with name `teamimage.ext` and/or `judgeimage.ext`, where `ext` is one of: `png`, `jpg`, `jpeg`, or `svg`.
```
RagnarGrootKoerkamp, line 808: Also say that the output validator may create additional metadata files that can help the visualizer

spec/2023-07-draft.md
evouga: Should we say something about this in the spec? (Do we want to require matplotlib?)

RagnarGrootKoerkamp (approved):

RagnarGrootKoerkamp:
We also discussed allowing passing additional flags to the visualizer, e.g. to disable visualising large cases with number of nodes `>= n`. Then that could also be changed by a contest system during the contest.

evouga:
What is "number of nodes"?

Can the judge system just ignore/delete image files it deems too large? Or is the idea to save on CPU cost of judging the submission?

RagnarGrootKoerkamp:
As in, for example the number of nodes in an input graph. Indeed, the idea is to not spend time rendering cases that are way too large so that they take very long to render anyway and probably wouldn't be helpful to inspect anyway.

PR #260, opened 2024-08-17
First draft for static validation

### LINKED ISSUE #130: Support for static validator(s)
A static validator is a programs that validates a submission independent of test data. I.e. it would be a program that gets the source code submitted and nothing else and either accepts or rejects the submission.

This could be used in a teaching context, either to require a code standard, or for (approximating) something like "Solve this using two for loops". It could also be used for things like "You're not allowed to use BigInteger".

I don't see how you could ever make this language agnostic? I.e. it would always be used together with `languages`.

I don't think this would ever be useful in a contest?

(This *could* be used for #46 as well, but I think that would be a very bad way to do that)

@thorehusfeldt @ghamerly @Tagl Thoughts?

Tagl:
Might close #130

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -1172,8 +1172,8 @@ A static validator may be provided under the `static_validator` directory, simil
 
 Each test group may define a static validation test case.
 A static validation test case is defined within a group's `testdata.yaml` file by specifying the key `static_validation`.
-If a map is specified, it may have two keys `flags`, and in the case of scoring test groups, `score`.
-The key `flags` maps to a string which represents the additional arguments passed to the static validator in this group's static validation test case.
+If a map is specified, it may have two keys `args`, and in the case of scoring test groups, `score`.
+The key `args` maps to a string which represents the additional arguments passed to the static validator in this group's static validation test case.
 The key `score` maps to a float which represents both the maximum score achievable for the static validation test case and the default score assigned in case the static validator accepts the submission for that test case.
 The static validator can override this score by outputting a value to `score.txt` in the feedback directory, the same as an output validator.
 It can have the value of `false` meaning there is no static validaton, `true` meaning that static validation is enabled with no score defined and no flags.
@@ -1205,4 +1205,4 @@ The meaning of the parameters listed above are:
   so that simply appending a filename to feedback_dir gives the path to a file in the feedback directory.
   
 - additional_arguments:
-  in case the problem specifies additional validator_flags, these are passed as additional arguments to the validator on the command line.
+  in case the static validation test case specifies additional args, these are passed as additional arguments to the validator on the command line.
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -563,6 +563,7 @@ Key                     | Type                                | Default         
 `scoring`               | Map                                 | [See Verdict/Score Aggregation](#verdict/score-aggregation) | Not inherited  | Description of how the results of the group test cases and subgroups should be aggregated. This key is only permitted for the `secret` group and its subgroups.
 `input_validator_args`  | String or map of strings to strings | empty string                            | Inherited      | Arguments passed to each input validator for this test data group. If a string then those are the flags that will be passed to each input validator for this test data group. If a map then each key is the name of the input validator and the value is the flags to pass to that input validator for this test data group. Validators not present in the map are run without flags.
 `output_validator_args` | String                              | empty string                            | Inherited      | Arguments passed to the output validator for this test data group.
+| static_validation      | Map                                 | [See Static Validation](#static-validator) | Not applicable | Configuration of static validation test data node.
 `full_feedback`         | Boolean                             | `false` in `secret`, `true` in `sample` | Inherited      | See [Test Case Configuration](#test-case-configuration).
 
 ### Invalid Test Cases
@@ -969,7 +970,7 @@ The meaning of the parameters listed above are:
 - feedback_dir:
   a string which specifies the name of a "feedback directory" in which the validator can produce "feedback files" in order to report additional information on the validation of the output file.
   The feedbackdir must end with a path separator (typically '/' or '\\' depending on operating system),
-  so that simply appending a filename to feedbackdir gives the path to a file in the feedback directory.
+  so that simply appending a filename to feedback_dir gives the path to a file in the feedback directory.
 
 - additional_arguments:
   in case the problem specifies additional validator_args, these are passed as additional arguments to the validator on the command line.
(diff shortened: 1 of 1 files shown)


spec/2023-07-draft.md
```diff
@@ -927,3 +928,45 @@ The defaults are as follows:
...
+The map can have two keys `flags`, and in the case of scoring test groups, `score`.
+The key `flags` maps to a string which represents the additional arguments passed to the static validator in this group's static validation test case.
+The key `score` maps to a float which represents both the maximum score achievable for the static validation test case and the default score assigned in case the static validator accepts the submission for that test case.
+The static validator can override this score by outputting a value to `score.txt` in the feedback directory, the same as an output validator.
+Aggregation is then applied to the test case in the same manner as other test cases.
+It is an error to assign a score in a pass-fail test group.
+
+### Invocation
+
+When invoked the static validator will be passed at least two command line parameter.
```
RagnarGrootKoerkamp, line 954:
```suggestion
When invoked the static validator will be passed at least two command line parameters.
```

spec/2023-07-draft.md
```diff
@@ -927,3 +928,45 @@ The defaults are as follows:
...
+
+The validator should be possible to use as follows on the command line:
+```sh
+<static_validator_program> submission_dir feedback_dir [additional_arguments]
+```
+
+The meaning of the parameters listed above are:
+
+- submission_dir:
+  a string specifying the absolute path to a directory containing a copy of the submission
```
RagnarGrootKoerkamp, line 964: for single-file submissions, is this the path to the file directly? Otherwise, should we pass the entrypoint / main file?

spec/2023-07-draft.md
```diff
@@ -781,7 +782,7 @@ The meaning of the parameters listed above are:
 - feedback_dir:
   a string which specifies the name of a "feedback directory" in which the validator can produce "feedback files" in order to report additional information on the validation of the output file.
   The feedbackdir must end with a path separator (typically '/' or '\\' depending on operating system),
```
JoelNiemela, line 784:
```suggestion
  feedback_dir must end with a path separator (typically '/' or '\\' depending on operating system),
```

evouga:
Assuming the scoring changes get merged, `score.txt` should be in [0,1] and scales the maximum point value.

spec/2023-07-draft.md
JoelNiemela, line 954:
```suggestion
When invoked, the static validator will be passed at least two command line parameters.
```

spec/2023-07-draft.md
JoelNiemela, line 964: "Submission files" might be more clear.
```suggestion
  a string specifying the absolute path to a directory containing a copy of the submission files
```

spec/2023-07-draft.md
```diff
@@ -927,3 +928,45 @@ The defaults are as follows:
...
+```
+
+The meaning of the parameters listed above are:
+
+- submission_dir:
+  a string specifying the absolute path to a directory containing a copy of the submission
+
+- feedback_dir:
+  a string which specifies the name of a "feedback directory" in which the validator can produce "feedback files" in order to report additional information on the validation of the submission.
+  The feedbackdir must end with a path separator (typically '/' or '\\' depending on operating system),
```
JoelNiemela, line 968:
```suggestion
  feedback_dir must end with a path separator (typically '/' or '\\' depending on operating system),
```

spec/2023-07-draft.md
```diff
@@ -927,3 +928,45 @@ The defaults are as follows:
...
+
+A static validator is a program that is given the submission files as input and can analyze the contents to accept or reject the submission.
+Optionally, the static validator may assign a score to the submission.
+By default there is no static validator.
+A static validator may be provided under the `static_validator` directory, similar to a custom output validator.
+
+### Static Validation Test Cases
+
+Each test group may define a static validation test case.
+A static validation test case is defined within a group's `testdata.yaml` file by specifying the map `static_validation`.
```
JoelNiemela, line 944: Doesn't this mean that there's no way to have a static validator with no flags for pass-fail problems?
(Or alternatively: doesn't this force all pass-faill static validators to have `flags` set? Which feels unnecessary.)

spec/2023-07-draft.md
JoelNiemela: > for single-file submissions, is this the path to the file directly?

As I understand it, no. I also think linking directly to the file will make things more complicated for very little gain.

> Otherwise, should we pass the entrypoint / main file?

That is probably a good idea in either case?

evouga:
Do we need a submission_dir? The output validator is already required to run in a working directory containing the submitted files. We could do the same here?

spec/2023-07-draft.md
Tagl: You can specify flags as the empty string. But perhaps having an enabled: true wouldn't be bad to be clear.

spec/2023-07-draft.md
Tagl: Maybe we both pass in path to main file (code) and entrypoint (binary)?

JoelNiemela: I don't think the binary is needed?

I think the most obvious thing to do would be to pass in `submission_dir` and `entrypoint` (which would contain the name of the mainfile/mainclass (the exact meaning is language-dependent)).

If we think the binary is needed, I think the most stable (across languages) solution is to handle that similarly to the source code: passing in `compilation_dir` and `program` (I don't think the binary is needed though?)

Tagl:
> Do we need a submission_dir? The output validator is already required to run in a working directory containing the submitted files. We could do the same here?

That's right, I have fixed this.

spec/2023-07-draft.md
JoelNiemela: Never mind — as per @evouga's comment below, `submission_dir` isn't needed.

I think `entrypoint` alone would suffice, but if not, `entrypoint` and `binary` (or something like that) would also make sense.

spec/2023-07-draft.md
JoelNiemela: Alternatively, `static_validation: true` could be used when no relevant options should be set. (I'm not sure if that is more or less ugly than having an `enabled` flag...)

spec/2023-07-draft.md
Tagl: Entrypoint in C/C++/Rust is the binary
Entrypoint in Java/Scala is the main class
Entrypoint in Python/Bash/PHP is the main code file

Are we sure entrypoint suffices?

niemela: > Are we sure entrypoint suffices?

I think it needs the language and the entrypoint.

Tagl: Language and entry point added instead

spec/2023-07-draft.md
Tagl: I have no strong opinion, we use shorthands in YAML in many places so just true is fine here I think.

Tagl: Setting false would explicitly disable it then, which is essentially a no-op

spec/2023-07-draft.md
```diff
@@ -563,6 +564,7 @@ Key                     | Type                                | Default
 `scoring`               | Map                                 | [See Verdict/Score Aggregation](#verdict/score-aggregation) | Not inherited  | Description of how the results of the group test cases and subgroups should be aggregated. This key is only permitted for the `secret` group and its subgroups.
 `input_validator_args`  | String or map of strings to strings | empty string                            | Inherited      | Arguments passed to each input validator for this test data group. If a string then those are the flags that will be passed to each input validator for this test data group. If a map then each key is the name of the input validator and the value is the flags to pass to that input validator for this test data group. Validators not present in the map are run without flags.
 `output_validator_args` | String                              | empty string                            | Inherited      | Arguments passed to the output validator for this test data group.
+| static_validation      | Map or boolean                      | false                   | Not applicable | Configuration of static validation test data node. See [Static Validation](#static-validation)
```
niemela, line 567:
```suggestion
`static_validation`    | Map or boolean                      | false                   | Not applicable | Configuration of static validation test data node. See [Static Validation](#static-validation)
```

niemela (changes requested):

niemela (approved):

PR #271, opened 2024-08-18
Remove paragraph restricting languages to include file languages

JoelNiemela:
Languages used to be restricted to the languages that `/include` files were available for — since the `languages` key in `problem.yaml` has been added, this is no longer needed and actually make internal contradictions possible (since the set of allowed languages is set by two different factors, which may contradict each other).

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -654,8 +654,6 @@ If a tool was used to automate creating test case illustration annotations, it i
 Files that should be included with all submissions are provided in one non-empty directory per supported language.
 Files that should be included for all languages are placed in the non-empty directory `include/default/`.
 Files that should be included for a specific language, overriding the default, are provided in the non-empty directory `include/<language>/`.
-If either no included files are provided or default files are provided, then the set of languages in which a submission is allowed to be is unrestricted.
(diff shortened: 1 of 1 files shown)


spec/2023-07-draft.md
```diff
@@ -654,8 +654,6 @@ If a tool was used to automate creating test case illustration annotations, it i
 Files that should be included with all submissions are provided in one non-empty directory per supported language.
 Files that should be included for all languages are placed in the non-empty directory `include/default/`.
 Files that should be included for a specific language, overriding the default, are provided in the non-empty directory `include/<language>/`.
-If either no included files are provided or default files are provided, then the set of languages in which a submission is allowed to be is unrestricted.
-However, if included files are provided in specific languages, but no default files are provided, then the set of languages is restricted to the specified languages.
```
niemela, line 658: I think it should be an error if `languages` is inconsistent with `include`?

I think the only way it can be inconsistent is if any language has an include, but that language is not allowed?

JoelNiemela: Fixed:

It is now specified that `include/<language>/` may only be specified for `<language>`s that are specified by `languages` in `problem.yaml`.

niemela (approved):

PR #276, opened 2024-08-18
Explicitly mention CC license versions

eldering:
This means that if the owner of a problem package has licensed his problem hypothetically in the future under CC by-sa 5, then he can still select `cc -by-sa` here without the specification implying that it would actually be licensed under version 4.

For future versions of the specification we can bump the CC license version to the latest version (if any new version was released), assuming that we want to require problem owners to always license under the latest version of a CC license.

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -259,15 +259,15 @@ source:
 License under which the problem may be used.
 Must be one of the values below.
 
-| Value         | Comments                                                                           | Link                                             |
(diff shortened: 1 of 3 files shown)


Tagl:
Should we not allow typing the extra 4 to specify an exact version?

niemela:
> Should we not allow typing the extra 4 to specify an exact version?

I don't think that's needed.

niemela (approved):

Issue #279, opened 2024-08-18
Symlinks can't reach outside of package


niemela:
We should specify that symlinks can't refer outside of a package.

(Do we also need to specifically state that symlinks are allowed?)

This should be done for both `legacy`, `legacy-icpc`, and `draft-*`.

PR #281, opened 2024-08-18
Allow symlinks unless they point outside the problem package

evouga:
Fixes https://github.com/Kattis/problem-package-format/issues/279


niemela (changes requested):
This should also be backported to `legacy` and `legacy-icpc`.

niemela (approved):

PR #291, opened 2024-08-18
Revisit samples

### LINKED ISSUE #263: Mixing of .in and .interaction samples
Would it be reasonable that a problem being interactive means you should only be allowed to have .interaction samples (no .in)?

If not, should you ever be allowed to mix .in and .interaction? has this *ever* occured?

RagnarGrootKoerkamp:
Fix #263, hopefully

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -654,12 +654,12 @@ in `group1`, but times out on at least one case in `group2`.
 ```yaml
 solves_group_1.py:
   sample:
-    permitted: AC
+    permitted: [AC]
   secret/group1:
-    permitted: AC
+    permitted: [AC]
   secret/group2:
     permitted: [AC, TLE]
-    required: TLE
+    required: [TLE]
 ```
 
 #### Glob patterns
@@ -670,12 +670,12 @@ cases, for example, when test groups are not used:
 ```yaml
 time_limit_exceeded/solves_easy_cases.py:
   sample:
-    permitted: AC
+    permitted: [AC]
   secret/*-easy:
-    permitted: AC
+    permitted: [AC]
   secret/*-hard:
     permitted: [AC, TLE]
-    required: TLE
+    required: [TLE]
 ```
 This means that the submission must solve all samples and all easy cases, but
 must time out on at least one of the hard cases.
@@ -698,22 +698,22 @@ By default, the following requirements are defined:
 ```yaml
 # All cases must be accepted.
 accepted:
-  permitted: AC
+  permitted: [AC]
 # At least one case is not accepted.
 rejected:
   required: [RTE, TLE, WA]
 # All cases AC or WA, at least one WA.
 wrong_answer:
   permitted: [AC, WA]
-  required: WA
+  required: [WA]
 # All cases AC or TLE, at least one TLE.
 time_limit_exceeded:
   permitted: [AC, TLE]
-  required: TLE
+  required: [TLE]
 # All cases AC or RTE, at least one RTE.
 run_time_error:
   permitted: [AC, RTE]
-  required: RTE
+  required: [RTE]
 # Must not WA, but fail at least once.
 # Note that by default these are not used for determining the time limit.
 brute_force:
@@ -729,7 +729,7 @@ normal usage.
 ```yaml
 time_limit_exceeded:
   permitted: [AC, WA, TLE]
-  required: TLE
+  required: [TLE]
 ```
 
 Note that the glob `time_limit_exceeded/*` would impose an _additional_
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -580,7 +580,7 @@ Sample test cases can be used in three places:
 By default the sample data for all three cases is taken from the `.in` and `.ans` file pairs under `data/sample`.
 Some problems require (slightly) different data in each of these cases.
 We allow customizing this in subdirecties `data/sample/statement` and `data/sample/download`.
-In this case it recommended to symlink identical files from these subdirectories to those in `data/sample`.
+In this case it is recommended to symlink identical files from these subdirectories to those in `data/sample`.
 
 #### Samples used by the judge
 The `data/sample` directory contains test cases similar to those in `data/secret`.
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -591,21 +591,21 @@ cases (for example when the samples shown in the PDF are not actually valid case
 
 #### Samples shown in the problem statement
 By default, the `.in` and `.ans` pairs from `data/sample` are shown in the
-statement.
+problem statement.
 This can be customized by creating the `data/sample/statement` directory. If
 this directory exists, _only_ those testcases present in this directory are shown, and the
 test cases in `data/sample` are ignored for purposes of the problem statement.
 This directory is _required_ for interactive problems.
 
-For each test case, it must contain one of the following for default or custom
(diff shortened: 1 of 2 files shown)


niemela (approved):

niemela (approved):

spec/2023-07-draft.md
```diff
@@ -601,6 +571,74 @@ In particular, for an existing feedback directory `dir`,
...
+  submission (`>`).
+
+#### Samples available for download
+When `data/sample/download` exists, the files in there are available for
+download.
+If this directory does not exist, the contents of `data/sample/statement` are
+available for download, with any `.out` files renamed to `.ans`
+(replacing existing `.ans` files). When that also does not exist, the test cases in
+`data/sample` are available for download. `data/sample/download` is required for
+interactive problems.
```
simonlindholm, line 618: This seems like a harsh requirement. What about cases where you don't provide a testing tool and thus don't have anything useful for contestants to download? Leaving data/sample/download/ empty is unfortunately not a viable option with git.

spec/2023-07-draft.md
```diff
@@ -601,6 +571,74 @@ In particular, for an existing feedback directory `dir`,
...
+Some problems require (slightly) different data in each of these cases.
+We allow customizing this in subdirecties `data/sample/statement` and `data/sample/download`.
+In this case it is recommended to symlink identical files from these subdirectories to those in `data/sample`.
+
+#### Samples used by the judge
+The `data/sample` directory contains test cases similar to those in `data/secret`.
+Every submission is run against these.
+
+`data/sample` must not contain test groups. Also, it does not contain any test
+cases (for example when the samples shown in the PDF are not actually valid cases).
```
simonlindholm, line 590: I don't understand the second sentence here.

spec/2023-07-draft.md
```diff
@@ -601,6 +571,74 @@ In particular, for an existing feedback directory `dir`,
...
 
+### Samples
+Sample test cases can be used in three places:
+- All submissions are run on the samples, and possibly feedback is provided to teams.
+- The samples are shown in the problem statement.
+- The samples are available for download.
+
+By default the sample data for all three cases is taken from the `.in` and `.ans` file pairs under `data/sample`.
+Some problems require (slightly) different data in each of these cases.
+We allow customizing this in subdirecties `data/sample/statement` and `data/sample/download`.
```
simonlindholm, line 582: subdirectories*

spec/2023-07-draft.md
```diff
@@ -601,6 +571,74 @@ In particular, for an existing feedback directory `dir`,
...
+#### Samples shown in the problem statement
+By default, the `.in` and `.ans` pairs from `data/sample` are shown in the
+problem statement.
+This can be customized by creating the `data/sample/statement` directory. If
+this directory exists, _only_ those testcases present in this directory are shown, and the
+test cases in `data/sample` are ignored for purposes of the problem statement.
+This directory is _required_ for interactive problems.
+
+For each test case, `data/sample/statement` must contain one of the following for default or custom
+validation (non-interactive) problems:
```
simonlindholm, line 601: Does this mean that an .interaction file can only exist if there is an .in file in data/sample? That seems very restrictive -- I find that it somewhat frequently happens that you want to add an illustrative (perhaps incomplete because a full interaction might involve e.g. 10^5 queries?) sample interaction that doesn't have a corresponding .in.

spec/2023-07-draft.md
```diff
@@ -601,6 +571,74 @@ In particular, for an existing feedback directory `dir`,
...
+- A `.in` and `.ans` file, which are both shown.
+- A `.in`, `.ans`, and `.out` file, of which the `.in` and `.out` file are shown, and the
+  `.ans` file is used for validation.
+For interactive problems, it must contain one of:
+- A `.in` and `.out` file, which are both shown, neither of which is validated.
+- A `.interaction` file that contains lines starting with `<` and `>`,
+  containing an interaction log with output from the interactor (`<`) and
+  submission (`>`).
```
simonlindholm, line 609: (completely unrelated to the spec changes, but: it might be nice to allow "..." ellipsis lines too)

spec/2023-07-draft.md
niemela: Would there be no relevant files at all to download in these cases?

With the current spec, I believe you would get the `.in`, and `.ans` (I think?)

spec/2023-07-draft.md
niemela: I think this used to say "Also, it may not contain ..." but that is unclear. I think it should be changed to something like "Also, it is allowed to not contain ..." (but that's a bit awkward.

@RagnarGrootKoerkamp ?

spec/2023-07-draft.md
niemela: As written I would interpret it to mean that, yes.

That was not the intention. You should be allowed to "create" new "test cases" for the statement only by adding them here.

spec/2023-07-draft.md
simonlindholm: I would say there are no relevant files to download, indeed. In theory you could provide the .interaction file, but I think it'd be rather useless and probably confusing.

With current Kattis at least you're not allowed to download samples at all for interactive problems, a change which iirc was made ~3 years ago, and definitely one for the better, because before that we regularly had contestants confused about what the `.in` and `.ans` files meant (they weren't show in the statement online or their format explained in any way). I don't know if the old spec said anything about downloading samples in the first place.

RagnarGrootKoerkamp: > Leaving data/sample/download/ empty is unfortunately not a viable option with git.
You can make a `data/sample/download/.gitkeep`, and then we could specify that this particular `.gitkeep` file is ignored in the download? (Kinda ugly, I know.)

The reason that I made `data/sample/download` **required** is indeed that `data/sample/statement` is **never** what you want to offer for download. But I hadn't considered that Kattis already doesn't make anything available for download anyway. Domjudge definitely does allow downloading the 'wrong' `.in` and `.ans` currently.

Instead we could say that for interactive problems, `data/sample/download` doesn't fall back to `data/sample/statement` but instead is empty by default?

spec/2023-07-draft.md
RagnarGrootKoerkamp: Yes, what Fredrik said.

spec/2023-07-draft.md
RagnarGrootKoerkamp: Oh, the `for each test case` here is meant to be parsed as something like 'for each basename for which a `data/sample/statement/basename.ext` is found with some recognised extension, ...'.
Maybe we should additionally clarify that no other files may exist at all anyway.

spec/2023-07-draft.md
simonlindholm: That works, or you could say that it falls back to /statement but that .interaction files are excluded from download.

RagnarGrootKoerkamp: Hmm that could maybe work. I'm wondering how often the 'internal' `data/sample/*.{in,ans}` files for interactive problems have the same syntax as the public version for the testing tool.

Ah actually: providing `.in` and `.ans` really only makes sense if you have a testing tool, which should anyway go into `data/sample/download`.
So either you do have a testing tool, and you choose the `.in` and `.ans` you want to provide with the corresponding 'public' syntax (which may or may not be the same as the 'private' syntax of `data/sample/*.{in,ans}`), or you don't have a testing tool, and in that case providing `.in` and `.ans` files is kinda useless anyway.

So yes, making `data/sample/download` empty by default for interactive problems instead of requiring it sounds good.

PR #322, opened 2024-08-19
Reorder tables to make them more consistent

pehrsoderman:


niemela (changes requested):
I don't think tables should be listed in alphabetical order, but rather in "logical" order.

("logical" is of course a bit less well defined, but e.g. the current order is typically, the order in which things are defined in the spec.)

PR #325, opened 2024-08-20
Fix uses of i.e. and e.g.

JoelNiemela:


niemela (approved):

eldering:
More a discussion point, but I think it would be better to just write them all out to `that is` and `for example`.

Tagl:
> More a discussion point, but I think it would be better to just write them all out to `that is` and `for example`.

I was just trying to convince them of this in person 😄

eldering:
> > More a discussion point, but I think it would be better to just write them all out to `that is` and `for example`.
> 
> I was just trying to convince them of this in person 😄

I created https://github.com/Kattis/problem-package-format/issues/326

Issue #341, opened 2024-08-24
Why the subtle difference between "submissions" and "other programs"


niemela:
In "Overview" / "Programs" / "Submissions" we say:
> ... working directory that contains (a copy of) the submitted files, any compiled binaries and compilation byproducts ...

But in "Overview" / "Programs" / "Other Programs" we say:
> ... working directory that contains the submitted files and any compiled binaries ...

Are the subtle difference on purpose? I.e.
- "(a copy of)" vs not
- "compiled binaries and compilation byproducts" vs "compiled binaries"

I don't think so? Maybe the "copy" part?

evouga:
Has this issue been resolved?

niemela:
Yes, closing.

Issue #330, opened 2024-08-20
Make program (validators et.c.) environment more strictly defined.


niemela:
Comment from @gkreitz:
> It is super annoying that validators do not seem to have any documentation on what they need or what language they are written in. Supplying a `run` and/or `build` script without any guarantees on the environment other than the existence of `cc` and `cpp` just feels bad. As someone building a backend, I want to know what language the validator is written in, and what it needs for its runtime. As someone building a validator, I don't see how I would ever use `run` and `build` to safely do anything other than run C/C++. I'm very unsure what problem was solved by adding `run` and `build` instead of re-using the same system as is in place for submissions.

@eldering I think you have been pushing mostly for the necessity of this? @Tagl you as well?

Tagl:
My main use case for this has been linking to a library with C++ for the output validator where it saves a massive amount of time and ensures quality of the validator. The most recent use case was the use of the library Eigen for https://open.kattis.com/problems/flaedasmidi because our original validator was numerically unstable, and we needed something more robust.

That being said, this sort of thing I encounter very rarely, but sometimes I need to do it and the only other way to make the problem work would be to either implement the library myself or try to retrofit it to the problem package, both of which would take much more time than reasonable.
I am fine with limiting to specific commands, but it should not be too restrictive.
Allowing Python 3, C, and C++ as the only general purpose languages for validators is also something I support, with a possible addition of other high performance languages such as Rust if that becomes high in demand.

In Python 3 you can simply add the library as a directory within your validator and import it, but in C and C++ it's much nicer to be able to build the library as needed and link to it.

I do think this is something that may be best handled by allowing the things we know we need and then if we encounter more needs in the future we can add those to the spec with great hesitance to avoid too many required dependencies.

niemela:
> with a possible addition of other high performance languages such as Rust if that becomes high in demand.

There is always the *possibility* to add more languages in the future.

eldering:
I think @Tagl's case is already an important one: you might need to include a library: eigen, libgmp, matplotlib, etc that is not and should not be available in a contest language, and it doesn't make sense to add a contest language just to be able to build your validator (or other type of program).

Another example I've seen, are custom validators that consist of multiple programs chained together with a shell script. For example, first a checktestdata program to validate the syntax and then another program to validate the semantics. We've in the past also combined two separate validators into a single script.
The bottom line is that supporting scripts gives a lot of flexibility.

[...]

PR #342, opened 2024-09-17
Program clarifications

evouga:
Clarify program requirements; addresses https://github.com/Kattis/problem-package-format/issues/330 and https://github.com/Kattis/problem-package-format/issues/341

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -118,10 +118,11 @@ The language of files is inferred from their extension as listed in the [languag
 Each program must be run in a working directory with the following contents and **nothing else**:
 
 - For input validators: the files in the program directory of the input validator in question.
-- For submissions, ouput validators, output visualizers, and static validators: the submitted files, any compiled binaries of the submitted files and compilation byproducts, as well as any [included files](#included-files).
+- For submissions: the submitted files, any compiled binaries of the submitted files, any [included files](#included-files), and the contents of the `.files` directory of the test case being tested (if this directory exists).
+- For ouput validators, output visualizers, and static validators: the submitted files, any compiled binaries of the submitted files, as well as any included files.
 
 Please note that in particular:
-- the working directory for submissions **must not** contain any of the [test data](#test-data) files;
+- the working directory for submissions **must not** contain any of the [test data](#test-data) files, except for the contents of the test case `.files` directory;
 - except for input validators, the files in a program's directory are not included in the working directory.
 
 ## Problem Metadata
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -82,7 +82,9 @@ In the case that a program is a single file, it is treated as if a directory wit
 The name of the program, for the purpose of referring to it within the package, is the base name of the file or the name of the directory.
 There can't be two programs of the same kind with the same name.
 
-#### Submissions
+#### Languages and Compilation
+
+##### Submissions
 
 The language of a submission program is determined by the `language` key in `submissions.yaml` if present; 
 otherwise, by comparing the file extensions of the submission file(s) to those specified in the [languages table](../appendix/languages.md). 
@@ -92,10 +94,7 @@ If a single language can't be determined, building fails.
 For languages where there could be several entry points, the entry point specified by the `entrypoint` key in `submissions.yaml` is used if present;
 otherwise, the default entry point in the [languages table](../appendix/languages.md) is used.
 
-Each submission must be run with a working directory that contains (a copy of) the submitted files, any compiled binaries and compilation byproducts, as well as any [included files](#included-files), and nothing else. 
-In particular the working directory must not contain any of the [test data](#test-data) files.
-
-#### Other Programs
+##### Other Programs
 
 Other programs (validators and visualizers) provided as a directory may include one of two POSIX-compliant shell scripts, `build` and `run`.
 If at least one of these two files is included:
@@ -105,19 +104,25 @@ If at least one of these two files is included:
 2. Then, the `run` file (which now exists, and is an executable binary or POSIX-compliant shell script)
    will be used as the validator or visualizer program.
 
-Scripts may assume that a C and C++ compiler are available on the system search path, aliased to `cc` and `cpp` respectively.
+Scripts may assume a POSIX-compliant shell and that a Python 3 interpreter, C compiler, and C++ compiler are available on the system search path, aliased to `python`, `cc`, and `cpp` respectively. Contest systems may provide additional libraries or tools. Problem packages with `build` or `run` scripts are strongly encouraged to include a `README` file in the program directory documenting any such additional dependencies.
 
 Programs that do not include a `build` or `run` script must have one of the following forms:
-- a single Python 3 file.
-- a directory containing multiple Python 3 source files, two of which are `__init__.py`, (defining a module), and `__main__.py` (which will be used as the program entry point).
+- a single Python 3 file;
+- a directory containing multiple Python 3 source files, two of which are `__init__.py` (defining a module) and `__main__.py` (which will be used as the program entry point);
 - a single C or C++ source file, or a directory containing one or more such files.
 
 The language of files is inferred from their extension as listed in the [languages table](../appendix/languages.md).
 
-Each input validator must be run with a working directory that contains the files in the program directory of the input validator in question.
+#### Working Directory
 
-Each output validator, output visualizer, and static validator must be run with a working directory that contains the submitted files and any compiled binaries of the submission being validated.
-(Note in particular that this working directory does **not** contain the files in the program directory of the output validator, output visualizer or static validator.)
+Each program must be run in a working directory with the following contents and **nothing else**:
+
+- For input validators: the files in the program directory of the input validator in question.
+- For submissions, ouput validators, output visualizers, and static validators: the submitted files, any compiled binaries of the submitted files and compilation byproducts, as well as any [included files](#included-files).
+
+Please note that in particular:
+- the working directory for submissions **must not** contain any of the [test data](#test-data) files;
+- except for input validators, the files in a program's directory are not included in the working directory.
 
 ## Problem Metadata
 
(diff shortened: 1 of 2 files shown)


spec/2023-07-draft.md
```diff
@@ -105,19 +104,25 @@ If at least one of these two files is included:
 2. Then, the `run` file (which now exists, and is an executable binary or POSIX-compliant shell script)
    will be used as the validator or visualizer program.
 
-Scripts may assume that a C and C++ compiler are available on the system search path, aliased to `cc` and `cpp` respectively.
+Scripts may assume a POSIX-compliant shell and that a Python 3 interpreter, C compiler, and C++ compiler are available on the system search path, aliased to `python`, `cc`, and `cpp` respectively. Contest systems may provide additional libraries or tools. Problem packages with `build` or `run` scripts are strongly encouraged to include a `README` file in the program directory documenting any such additional dependencies.
```
niemela, line 107:
```suggestion
Scripts may assume a POSIX-compliant shell and that a Python 3 interpreter, C compiler, and C++ compiler are available on the system search path, aliased to `python3`, `cc`, and `cpp` respectively. 
Contest systems may provide additional libraries or tools. 
Problem packages with `build` or `run` scripts are strongly encouraged to include a `README` file in the program directory documenting any such additional dependencies.
```

I think `python3` is more common than `python`? This is what you would have in a minimal(-ish) Ubuntu image, and that seems like a reasonable basis?

I'm not sure what "Contest systems may provide additional libraries or tools" really means? I would argue that dropping that line does not change the meaning at all (we are definitely not saying that everything not mentioned in the spec is forbidden), and keeping it does not make anythign more clear (at least not to me).

spec/2023-07-draft.md
```diff
@@ -105,19 +104,25 @@ If at least one of these two files is included:
...
 
 Programs that do not include a `build` or `run` script must have one of the following forms:
-- a single Python 3 file.
-- a directory containing multiple Python 3 source files, two of which are `__init__.py`, (defining a module), and `__main__.py` (which will be used as the program entry point).
+- a single Python 3 file;
+- a directory containing multiple Python 3 source files, two of which are `__init__.py` (defining a module) and `__main__.py` (which will be used as the program entry point);
```
niemela, line 111: Wait, are we *requiring* a `__init__.py`? Why should be do that?

spec/2023-07-draft.md
```diff
@@ -105,19 +104,25 @@ If at least one of these two files is included:
...
 
 The language of files is inferred from their extension as listed in the [languages table](../appendix/languages.md).
 
-Each input validator must be run with a working directory that contains the files in the program directory of the input validator in question.
+#### Working Directory
+
+Each program must be run in a working directory with the following contents and **nothing else**:
+
+- For input validators: the files in the program directory of the input validator in question.
+- For submissions, ouput validators, output visualizers, and static validators: the submitted files, any compiled binaries of the submitted files and compilation byproducts, as well as any [included files](#included-files).
```
niemela, line 121: ...and the files in `.files`, right?

spec/2023-07-draft.md
```diff
@@ -105,19 +104,25 @@ If at least one of these two files is included:
...
 
-Each output validator, output visualizer, and static validator must be run with a working directory that contains the submitted files and any compiled binaries of the submission being validated.
-(Note in particular that this working directory does **not** contain the files in the program directory of the output validator, output visualizer or static validator.)
+Please note that in particular:
+- the working directory for submissions **must not** contain any of the [test data](#test-data) files;
```
niemela, line 124: Except `.files`, which *must* be there.

spec/2023-07-draft.md
niemela: What is meant by "compilation byproducts"?

niemela (changes requested):

spec/2023-07-draft.md
pehrsoderman: To ensure it is treated as a module in Python, which gives more consistency. You could also forbid it, forcing everybody to use namespace packages, but you should pick one.

evouga: What's the conclusion here?

evouga:
> What is meant by "compilation byproducts"?

I've removed this clause. (But is there any case where there are some dynamic libraries, or other files produced during compilation other than the submitted files and the compiled binary, that need to be in the folder?)

spec/2023-07-draft.md
pehrsoderman: My opinion is that forcing the use of a `__init__.py`, which results in a regular package is more pytonic, and probably somewhat less error prone. Also, I can't imagine much use of the features of namespace packages, like including directories external to the package directory into the package. However, my opinion is not very strong. If somebody can make a good argument for using namespace packages here I am willing to listen.

niemela:
> Yes, I agree the line does not actually specify anything. However (see also the discussion at #330) in practice people will want to link against additional libraries in their `build` and `run` scripts; my intent here is to (1) explicitly acknowledge that this is allowed and (2) encourage the best practice of documenting the needed libraries in a standard place.

I would argue that that discussion actually implies the opposite (and *I* would definitely argue for the opposite). I.e. if you want an additional library it should be *included* in the package (which kinda implies that "Contest systems may *not* provide additional libraries or tools", although it's strange to say that you can't provide *additional* things, when you have not *exactly* specified what should be provided).

evouga:
> > Yes, I agree the line does not actually specify anything. However (see also the discussion at #330) in practice people will want to link against additional libraries in their `build` and `run` scripts; my intent here is to (1) explicitly acknowledge that this is allowed and (2) encourage the best practice of documenting the needed libraries in a standard place.
> 
> I would argue that that discussion actually implies the opposite (and _I_ would definitely argue for the opposite). I.e. if you want an additional library it should be _included_ in the package (which kinda implies that "Contest systems may _not_ provide additional libraries or tools", although it's strange to say that you can't provide _additional_ things, when you have not _exactly_ specified what should be provided).

There is some elegance to that approach, but as a practical matter do you *really* want individual forks of Eigen, matplotlib, libpng, etc. (and all dependencies they recursive pull in) across some fraction of your problem packages?

niemela:
> There is some elegance to that approach, but as a practical matter do you _really_ want individual forks of Eigen, matplotlib, libpng, etc. (and all dependencies they recursive pull in) across some fraction of your problem packages?

Short answer: yes.

Long answer:

I see three possible approaches:

1. systems are allowed to provide *anything*, and similarly, problem packages are allowed to depend on *anything*. Pro: this is very simple for contest runners. A problem setter need only to verify with the tech staff for the contest they are running that the packages they need are available. Con: packages (that use this) become very unportable.
2. problem packages must include whatever they depend on (but as long as they do, they can depend on anything). Pro: this is still fairly simple (but less so), *and* it is reasonably portable. Con: packages become much larger. Popular dependencies in large archive could be stored *many* times.
3. problem packages may depend on some things they don't include, but must declare this (somehow). Pro: as simple as 1 for setters (and contest runners), and (almost) as portable as 2. Con: we have to set up (and keep up to date if there isn't something we could piggyback on) some kind of registry for libraries (and versions of libraries). Given versions it's not even certain that it's that much smaller than 2 for large archives.

I think 3 is *best*, but needs to be done well, and that's hard. 1 is worst; it's fine for single contests, but terrible for reuse and archives (i.e. "distributing and sharing problems" which is the stated purpose of the format). 2 is pretty much as good as 1 for single contests, and based on what has been used *so* far wouldn't be too bad for large archives either. Also, large archvie *could* use caching to make it better if it turns out to be bad (although, not quite as easily as with 3).

So, 2 is a lot easier than 3, and almost as good.

evouga:
But Option 2 also places more burden on the package writers. "this is still fairly simple": only for header-only libraries like Eigen, though? `libpng` is probably already more annoying than package authors are realistically willing to put up with; you need its source, the source of `zlib`, and some meta-scripts that configure, build, and install both of them reliably.

And from a quick look at `matplotlib` it seems a real nightmare to build from source without `pip`, requiring things like `FreeType` in addition to a dozen Python packages.

We rejected including even just a language version string for the package programs, under the argument that (1) it's too onerous on the package authors and (2) "surely" it won't be too difficult for an archivist in 20 years to figure out how to build the problem package. This dependency-bundling seems a *lot* more effort (albeit, only for problem packages that need libraries for visualizers etc. which are a minority).

I haven't needed external libraries for problems I've written (except maybe in generators, which are outside the scope of the spec) so I don't have much of a dog in this race---I'm fine with requiring all dependencies to be bundled, if this sounds reasonable to @eldering and others in the linked issue. Though I'm skeptical authors will obey such requirements in practice.

spec/2023-07-draft.md
evouga: @niemela Are you ok with keeping the __init__.py requirement?

niemela: Yes, I don't have a strong opinion on this. I was mostly surprised because it's a new requirement. If this seems reasonably to Python users, that I'm all for.

evouga:
What was the consensus here? Should I remove the line:
> Contest systems may provide additional libraries or tools
?

niemela:
> What was the consensus here? Should I remove the line:
> 
> > Contest systems may provide additional libraries or tools
> 
> ?

Yes. Repeating my argument from above:

> I'm not sure what "Contest systems may provide additional libraries or tools" really means? I would argue that dropping that line does not change the meaning at all (we are definitely not saying that everything not mentioned in the spec is forbidden), and keeping it does not make anythign more clear (at least not to me).

spec/2023-07-draft.md
```diff
@@ -105,19 +104,28 @@ If at least one of these two files is included:
 2. Then, the `run` file (which now exists, and is an executable binary or POSIX-compliant shell script)
    will be used as the validator or visualizer program.
 
-Scripts may assume that a C and C++ compiler are available on the system search path, aliased to `cc` and `cpp` respectively.
+Scripts may assume a POSIX-compliant shell and that a Python 3 interpreter, C compiler, and C++ compiler are available on the system search path, aliased to `python3`, `cc`, and `cpp` respectively. 
+Contest systems may provide additional libraries or tools.
```
niemela, line 108:
```suggestion
```

niemela (dismissed):

niemela:
I dropped the line:

> Contest systems may provide additional libraries or tools

At least for now. I will merge this now. If someone really want that line, please create a new PR.

Issue #328, opened 2024-08-20
The change of semantics of `score.txt` is a bit subtle


niemela:
Comment from @gkreitz:
> The complete change of semantics for score.txt is gonna be a complete pain to deal with (it's either a float in [0,1] if the test group score is bounded, or the actual score). It would have been much better IMHO to separate those two modes of operations into two files (score.txt vs scaledscore.txt ?) to not break every legacy problem

Thoughts on this?

Tagl:
I think separating is fine. Might make auto-updating packages easier too.

eldering:
I have no strong opinions on the matter, but I have to say that when I was reading the specification it was not very clear due to the complex inter-dependence of "unbounded" and scores in subgroups, etc. So I think that indeed separating this will also make it clearer.

niemela:
We did this. Closing.

PR #345, opened 2024-09-18
Score revamp

evouga:
Addresses https://github.com/Kattis/problem-package-format/issues/328

Also, tweaked the Maximum Score Inference section to fix what I think were some errors there---someone please double-check.

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -1123,7 +1123,7 @@ It is a judge error if `S > M` for a group with bounded maximum score and `sum` 
 
 ### Scoring Test Cases
 
-Only test cases in test case group with `sum` or `min` aggregation receive a score. 
+Only test cases in test case groups with `sum` or `min` aggregation receive a score. 
 
 The score of a failed test case is always 0.
 
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -1135,7 +1135,8 @@ A custom output validator may produce a `score.txt` or `unbounded_score.txt` fil
 It is a judge error if:
 - an output validator accepts a test case in an unbounded group and does not produce an `unbounded_score.txt`;
 - an output validator produces a `score.txt` for a test case in a group with `passs-fail` aggregation or with unbounded maximum score;
-- an output validator produces a `unbounded_score.txt` for a test case in a group with bounded maximum score.
+- an output validator produces a `unbounded_score.txt` for a test case in a group with bounded maximum score;
+- the score of a test case exceeds its maximum score.
 
 ### Scoring Test Groups
 
@@ -1150,7 +1151,9 @@ Otherwise the group score is 0.
 - If a group uses `sum` aggregation, the group score is the sum of the scores of its test cases and subgroups.
 - If a group uses `min` aggregation, then the group score is the minimum of these scores.
 
-The submission score is the score of the `secret` group. It is a judge error if the score of any group or subgroup exceeds its maximum score.
+The submission score is the score of the `secret` group. 
+
+It is a judge error if the score of any group or subgroup exceeds its maximum score.
 
 ### Required Dependent Groups
 
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -1109,11 +1109,11 @@ as is the maximum score of each test case in the group:
 Group Maximum Score | Aggregation Type     | Maximum Score of Test Case / Subgroup
 ------------------- | -------------------- | -------------------------------------
 `unbounded`         | any                  | `unbounded`
-bounded value `M`   | `sum` or `pass-fail` | `(M - S)/(A + T)`
-bounded value `M`   | `min`                | `M - S`
+bounded value `M`   | `pass-fail`          | 1
+bounded value `M`   | `sum`                | `(M - S)/(A + T)`
+bounded value `M`   | `min`                | `M`
 
-where the group has `T` test cases, `A` subgroups without a provided `score`, and whose other subgroups have maximum scores that sum to `S`.
-It is a judge error if `S > M`. This formula evenly distributes a group's leftover maximum points to its test cases and subgroups with unspecified maximum score.
+where the group has `T` test cases, `A` subgroups without a provided `score`, and whose other subgroups have maximum scores that sum to `S`. This formula evenly distributes a group's leftover maximum points to its test cases and subgroups with unspecified maximum score. It is a judge error if `S > M` for a group with bounded maximum score and `sum` aggregation.
 
 ### Scoring Test Cases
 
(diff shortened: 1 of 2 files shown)


spec/2023-07-draft.md
```diff
@@ -1109,38 +1109,40 @@ as is the maximum score of each test case in the group:
 Group Maximum Score | Aggregation Type     | Maximum Score of Test Case / Subgroup
```
niemela, line 1109: The last column should probably say "Default Maximum Score ..."? It is said in the paragraph just above, but it would maybe (?) make it just a little bit more clear.

spec/2023-07-draft.md
```diff
@@ -1109,38 +1109,40 @@ as is the maximum score of each test case in the group:
...
 ------------------- | -------------------- | -------------------------------------
 `unbounded`         | any                  | `unbounded`
```
niemela, line 1111: It says "any" which isn't *wrong*, but `pass-fail` is not allowed if score is `unbounded`. Maybe we should change "any" to "`sum` or `min`"?

spec/2023-07-draft.md
```diff
@@ -1109,38 +1109,40 @@ as is the maximum score of each test case in the group:
...
-bounded value `M`   | `sum` or `pass-fail` | `(M - S)/(A + T)`
-bounded value `M`   | `min`                | `M - S`
+bounded value `M`   | `pass-fail`          | 1
```
niemela, line 1112: In fact the maximum score does not matter in this case, but I guess 1 is as good a value as any? It might be easier to understand `aggregation` = 'pass-fail` if we don't pick a value here, but rather say that it doesn't matter?

spec/2023-07-draft.md
```diff
@@ -1109,38 +1109,40 @@ as is the maximum score of each test case in the group:
...
+bounded value `M`   | `sum`                | `(M - S)/(A + T)`
+bounded value `M`   | `min`                | `M`
 
-where the group has `T` test cases, `A` subgroups without a provided `score`, and whose other subgroups have maximum scores that sum to `S`.
-It is a judge error if `S > M`. This formula evenly distributes a group's leftover maximum points to its test cases and subgroups with unspecified maximum score.
+where the group has `T` test cases, `A` subgroups without a provided `score`, and whose other subgroups have maximum scores that sum to `S`. This formula evenly distributes a group's leftover maximum points to its test cases and subgroups with unspecified maximum score. It is a judge error if `S > M` for a group with bounded maximum score and `sum` aggregation.
```
niemela, line 1116:
```suggestion
where the group has `T` test cases, `A` subgroups without a provided `score`, and whose other subgroups have maximum scores that sum to `S`. 
This formula evenly distributes a group's leftover maximum points to its test cases and subgroups with unspecified maximum score. 
It is a judge error if `S > M` for a group with bounded maximum score and `sum` aggregation.
```

niemela: Should it also be a judge error if a subgroup maximum score is higher than the (bounded) maximum score for a group with  `min` aggregation?

spec/2023-07-draft.md
```diff
@@ -1109,38 +1109,40 @@ as is the maximum score of each test case in the group:
...
 
 ### Scoring Test Cases
 
 The score of a failed test case is always 0.
 By default, the score of an accepted test case is its maximum score, computed as described above.
-A custom output validator may produce a `score.txt` file for a test case:
+A custom output validator may produce a `score.txt` or `unbounded_score.txt` file for a test case:
```
niemela, line 1122: I think this fix goes the wrong way. This basically *renames* the old `score.txt` to `unbounded_score.txt` and creates a *new* file that takes the old name `score.txt`.

I think what we wanted was that `score.txt` should keep it's current semantics, i.e. it always contains the actual score, while a new file (say `score_ratio.txt`, alternatives to ratio coul be "multiplier", "factor", "relative"?) always contains a value between 0 and 1 (inclusive).

spec/2023-07-draft.md
```diff
@@ -1109,38 +1109,40 @@ as is the maximum score of each test case in the group:
...
 
 - for test cases in a group with bounded maximum score, `score.txt` must contain a single floating-point number in the range `[0,1]`.
   The score of the test case is this number _multiplied_ by the test case maximum score.
-
-- for test cases in unbounded groups, `score.txt` must contain a non-negative floating-point number.
+- for test cases in unbounded groups, `unbounded_score.txt` must contain a non-negative floating-point number.
```
niemela, line 1126: Revert this based on above.

spec/2023-07-draft.md
```diff
@@ -1109,38 +1109,40 @@ as is the maximum score of each test case in the group:
...
   The score of the test case is that number.
 
-It is a judge error if an output validator accepts a test case in an unbounded group and does not produce a `score.txt`.
-It is also a judge error if an output validator produces a `score.txt` for a test case in a group with `passs-fail` aggregation.
+It is a judge error if:
+- an output validator accepts a test case in an unbounded group and does not produce an `unbounded_score.txt`;
```
niemela, line 1130:
```suggestion
- an output validator accepts a test case in an unbounded group and does not produce an `score.txt`;
```

spec/2023-07-draft.md
```diff
@@ -1109,38 +1109,40 @@ as is the maximum score of each test case in the group:
...
+- an output validator produces a `score.txt` for a test case in a group with `passs-fail` aggregation or with unbounded maximum score;
```
niemela, line 1131:
```suggestion
- an output validator produces a `score.txt` for a test case in a group with `passs-fail` aggregation;
```

spec/2023-07-draft.md
```diff
@@ -1109,38 +1109,40 @@ as is the maximum score of each test case in the group:
...
+- an output validator produces a `unbounded_score.txt` for a test case in a group with bounded maximum score.
```
niemela, line 1132:
```suggestion
- an output validator produces a `score_ratio.txt` for a test case in a group with unbounded maximum score.
```

spec/2023-07-draft.md
```diff
@@ -1109,38 +1109,40 @@ as is the maximum score of each test case in the group:
...
```
niemela, line 1133: It should also be a judge error is an output validator produces a `score.txt` outside [0, maximum] (or a `score_ratio.txt` outside [0, 1]), but I'm not sure if it should be listed *here*?

niemela (changes requested):

spec/2023-07-draft.md
evouga: Yes.

evouga:
Do we want to forbid producing `score.txt` or `score_multiplier.txt` for failed test cases? Or is to more convenient/backward-compatible to allow this (but ignore the files)?

evouga:
Are there any remaining issues with this PR?

niemela:
> Do we want to forbid producing `score.txt` or `score_multiplier.txt` for failed test cases? Or is it more convenient/backward-compatible to allow this (but ignore the files)?

That is a good question. I will move this to a new issue (that we may or may not do anything about, but we can at least discuss it) So that the rest of the changes in this PR can be merged.

Issue #338, opened 2024-08-23
make data for testcases (again)


mzuenni:
With the last changes `data/samples/downloads/` was introduced, which is useful to overwrite testcases for the sample.zip.
However, it was decided that also non testcase related stuff should be placed there. But I think it is better if files in `data` should correspond to testcases .

For clarification `/data/sample/<xxx>.in` and `/data/sample/<xxx>.ans` both belong to the testcase `/sample/<xxx>` and files like `/data/sample/downloads/<xxx>` and `/data/sample/statement/<xxx>` also belong to that testcase.
But a file like `/data/sample/downloads/testing_tool.py` does not belong to any testcases.
Unfortunately this has the following side effects:
 - tooling can not detect if there is a missing testcase `/data/samples/testing_tool`
 - tools generating testcases can no longer detect outdated files

And this does not only happen for testing tools but anything that does not match a testcases.
Therefore I would propose to have `attachments` for stuff that does not belong to a testcase and make `/data` only for files to belong to a specific testcases.

RagnarGrootKoerkamp:
More generally, we should consider what exactly we intend to be the difference between `/attachements` and `/data/sample/download`.
Why do we present downloads in two places anyway in the UI? If we merge those from a user point of view, we can simply put the testing tool in `attachments` again.

Say you have some dictionary data attachement. Why should that not be included in the 'sample download' directly. Also if you run a contest and provide a zip with all samples, it does make sense for attachments to also be included right there, right?

If we do want to keep the attachments vs samples download, we could also consider a new top level `/testing_tool` directory, or simply `/downloads`.

evouga:
Has this issue been resolved? The current draft spec does have `attachments`.

evouga:
We do say
> If you want to provide files related to interactive problems (such as testing tools or input files), you can use data/sample/download
and I agree that `attachments` is a better place for testing tools.

RagnarGrootKoerkamp:
No this is still open. We have have overshot a bit when we were writing this up in Stockholm. I'm not on top of this atm though.

@mzuenni: Is all you are proposing that the testing tool moving out of `data/samples`, and the rest there stays the same? Or would you prefer to also move 'testing-tool-only inputs' back to `attachments/`?

mzuenni:
I am not sure what you mean with "testing-tool-only inputs". I would say that everything that does not belong to one specific testcase does not belong in data. (in bapctools generator.yaml terms: everything that is not numbered should not be in data)

[...]

PR #346, opened 2024-09-25
Cleanup of the Sample section

evouga:
I've done a pass on this section. The requirements here are quite complex so please double-check everything carefully.

In particular:
 - I've encouraged use of attachments where appropriate (per https://github.com/Kattis/problem-package-format/issues/338).
 - Is it true that all submissions are run on the test cases in data/sample? Including in scoring problems (despite these test cases not being included in the score) and interactive problems?
 - I've softened the hard requirement that interactive problems must have `sample/statement` and `sample/download`.
 - Clarified that if both `.interaction` and `.in` and `.out` is provided, only the former is shown (is this true)?
 - A strict reading of the spec suggested that interactive problems could only have a single test case in `sample/statement`. I assume that was unintentional?
 - Clarified that downloads do not include the `.interaction` files by default (is this true?)
 - Clarified what is means to validate the `.ans` files.
 - Clarified that `validator_args` in `data/sample` also apply to `data/sample/statement` (is this true?)

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -596,70 +596,70 @@ The directory `invalid` can be organized into a tree-like structure similar to `
 ### Samples
 
 Sample test cases can be used in three places:
-- All submissions are run on the samples, and possibly feedback is provided to teams.
-- The samples are shown in the problem statement.
-- The samples are available for download.
+- As test cases for team submissions (with feedback possibly provided to the teams).
+- As sample input and output displayed in the problem statement.
+- As sample input and output files available for download.
 
 By default the sample data for all three cases is taken from the `.in` and `.ans` file pairs under `data/sample`.
 Some problems require (slightly) different data in each of these cases.
-We allow customizing this in subdirectories `data/sample/statement` and `data/sample/download`.
+We allow customizing which data is used for each purpose using the subdirectories `data/sample/statement` and `data/sample/download`.
 In this case it is recommended to symlink identical files from these subdirectories to those in `data/sample`.
 
-#### Samples used by the judge
+#### Samples For Judging Team Submissions
 
 The `data/sample` directory contains test cases similar to those in `data/secret`.
-Every submission is run against these.
+Every submission is run on these test cases (but sample test cases do not contribute to the problem score for [scoring problems](#scoring-problems)).
 
-`data/sample` must not contain test groups.
-Also, it does not _have to_ contain any test cases, for example when the samples shown in the PDF are not actually valid cases.
+`data/sample` must not contain test groups. It may be missing (for problems with no samples) or empty.
 
-#### Samples shown in the problem statement
+#### Samples Shown in the Problem Statement
 
 By default, the `.in` and `.ans` pairs from `data/sample` are shown in the problem statement.
-This can be customized by creating the `data/sample/statement` directory.
+This behavior can be customized by creating the `data/sample/statement` directory.
 If this directory exists, its contents replaces that of `data/sample` for purposes of the problem statement.
-This directory is _required_ for interactive problems.
-
-For each test case, `data/sample/statement` must contain one of the following for default or custom validation (non-interactive) problems:
-- An `.in` and `.ans` file, which are both shown.
-- An `.in`, `.ans`, and `.out` file, of which the `.in` and `.out` file are shown, and the `.ans` file is used for validation.
+Interactive problems with sample test cases should almost always make use of this directory, since the `.in` and `.ans` pairs do not meaningfully capture how users are expected to interact with the output validator in such problems.
 
 ##### Interactive Problems
 
-For interactive problems, `data/sample/statement` must contain one of:
-- An `.in` and `.out` file, which are both shown, neither of which is validated.
+For interactive problems, sample data in `data/sample/statement` must have one of the following forms:
+- An `.in` and `.out` file, which are both shown, and neither of which is validated.
 - An `.interaction` file that contains lines starting with `<` and `>`,
   containing an interaction log with output from the output validator starting with `<` and output from the submission starting with `>`.
-    - An `.interaction` file **may** be specified without a corresponding test case —
-      any such interaction log will be displayed as usual.
 
-If you want to provide files related to interactive problems (such as testing tools or input files), you can use [data/sample/download](#samples-available-for-download).
+Note that an `.interaction` file does **not** need to be accompanied by corresponding `.in` and `.out` files. If both an `.interaction` file and corresponding `.in` and `.out` files are present, only the interaction log in the `.interaction` file is shown in the problem statement.
 
-#### Samples available for download
+If you want to make testing tools available for download, you can use [attachments](#attachments).
+
+##### Other Problem Types
+
+For each test case, `data/sample/statement` must contain one of the following:
+- An `.in` and `.ans` file, which are both shown.
+- An `.in`, `.ans`, and `.out` file, of which the `.in` and `.out` file are shown, and the `.ans` file is used for validation (see below).
+
+
+#### Samples Available for Download
 
 When `data/sample/download` exists, the files in there are available for download.
 If this directory does not exist, the contents of `data/sample/statement` are available for download,
-with any `.out` files renamed to `.ans` (replacing existing `.ans` files).
+with any `.out` files renamed to `.ans` (replacing existing `.ans` files) and all `.interaction` files removed.
 When that also does not exist, the test cases in `data/sample` are available for download.
-`data/sample/download` is required for interactive problems.
 
-The `data/sample/download` directory may contain anything, including, e.g., a testing tool for interactive problems, and is not validated.
+The `data/sample/download` directory may contain anything, and none of its contents are validated. We recommend placing test-case-specific files in this directory, and using [attachments](#attachments) for testing tools and other downloads not associated with any particular test case.
 Testing tools may warn when test cases that are listed in `data/sample` or `data/sample/statement` do not appear in `data/sample/download`,
 or when `data/sample/download` contains test cases that do not appear in one of the two other locations.
 
 #### Validation
 
-All `data/sample/*.in` files are input-validated.
+All `data/sample/*.in` files are input-validated. For problems that are not interactive, all `data/sample/statement/*.in` are also input-validated.
 
-When `data/sample/statement/` does not exist, `data/sample/*.ans` is output validated (i.e., must be accepted).
-
-When `data/sample/statement/` does exists, and the problem is not interactive,
-all `data/sample/statement/*.in` are input validated, and `data/sample/statement/*.out` (with fallback to `data/sample/statement/*.ans`) are output validated (i.e., must be accepted).
+When `data/sample/statement/` does not exist, all test cases in `data/sample` must be accepted by the output validator, with the `.ans` file used as both the `answer_file` and `team_output`.
+ 
+When `data/sample/statement/` does exists, and the problem is not interactive, all test cases in `data/sample/statement/` must be accepted by the output validator. The `.ans` is used as the `answer_file`. The `.out` file is used as the `team_output`, if it exists. Otherwise the `.ans` is used as the `team_output`. 
 
 Files in `data/sample/download` are never validated, 
 although tooling may warn when there are inconsistencies between `download`, `statement`, and `data/sample/` itself.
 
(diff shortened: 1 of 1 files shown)


spec/2023-07-draft.md
```diff
@@ -596,70 +596,70 @@ The directory `invalid` can be organized into a tree-like structure similar to `
 ### Samples
 
 Sample test cases can be used in three places:
-- All submissions are run on the samples, and possibly feedback is provided to teams.
-- The samples are shown in the problem statement.
-- The samples are available for download.
+- As test cases for team submissions (with feedback possibly provided to the teams).
+- As sample input and output displayed in the problem statement.
+- As sample input and output files available for download.
```
niemela, line 601: How about:
```suggestion
- As sample input and output files available for download, or otherwise made available.
```

spec/2023-07-draft.md
```diff
@@ -596,70 +596,70 @@ The directory `invalid` can be organized into a tree-like structure similar to `
...
-We allow customizing this in subdirectories `data/sample/statement` and `data/sample/download`.
+We allow customizing which data is used for each purpose using the subdirectories `data/sample/statement` and `data/sample/download`.
 In this case it is recommended to symlink identical files from these subdirectories to those in `data/sample`.
 
-#### Samples used by the judge
+#### Samples For Judging Team Submissions
 
 The `data/sample` directory contains test cases similar to those in `data/secret`.
-Every submission is run against these.
+Every submission is run on these test cases (but sample test cases do not contribute to the problem score for [scoring problems](#scoring-problems)).
```
niemela, line 611:
```suggestion
Every submission is run on these test cases 
(but sample test cases do not contribute to the problem score for [scoring problems](#scoring-problems)).
```

spec/2023-07-draft.md
```diff
@@ -596,70 +596,70 @@ The directory `invalid` can be organized into a tree-like structure similar to `
...
 
-`data/sample` must not contain test groups.
-Also, it does not _have to_ contain any test cases, for example when the samples shown in the PDF are not actually valid cases.
+`data/sample` must not contain test groups. It may be missing (for problems with no samples) or empty.
```
niemela, line 613:
```suggestion
`data/sample` must not contain test groups. 
It may be missing (for problems with no samples) or empty.
```

spec/2023-07-draft.md
```diff
@@ -596,70 +596,70 @@ The directory `invalid` can be organized into a tree-like structure similar to `
...
-If you want to provide files related to interactive problems (such as testing tools or input files), you can use [data/sample/download](#samples-available-for-download).
+Note that an `.interaction` file does **not** need to be accompanied by corresponding `.in` and `.out` files. If both an `.interaction` file and corresponding `.in` and `.out` files are present, only the interaction log in the `.interaction` file is shown in the problem statement.
+
+If you want to make testing tools available for download, you can use [attachments](#attachments).
+
+##### Other Problem Types
+
+For each test case, `data/sample/statement` must contain one of the following:
+- An `.in` and `.ans` file, which are both shown.
+- An `.in`, `.ans`, and `.out` file, of which the `.in` and `.out` file are shown, and the `.ans` file is used for validation (see below).
```
niemela, line 637: I'm a little bit confused by this. Is the presence of the `.ans` file what triggers validation of the `.in` and `.out` pair? Why do we want do validate these? More specifically, what is the use case for having a different set of `.in` and `.out` files than what is run, but that still validates?

spec/2023-07-draft.md
```diff
@@ -596,70 +596,70 @@ The directory `invalid` can be organized into a tree-like structure similar to `
...
 
 When `data/sample/download` exists, the files in there are available for download.
 If this directory does not exist, the contents of `data/sample/statement` are available for download,
-with any `.out` files renamed to `.ans` (replacing existing `.ans` files).
+with any `.out` files renamed to `.ans` (replacing existing `.ans` files) and all `.interaction` files removed.
 When that also does not exist, the test cases in `data/sample` are available for download.
-`data/sample/download` is required for interactive problems.
 
-The `data/sample/download` directory may contain anything, including, e.g., a testing tool for interactive problems, and is not validated.
+The `data/sample/download` directory may contain anything, and none of its contents are validated. We recommend placing test-case-specific files in this directory, and using [attachments](#attachments) for testing tools and other downloads not associated with any particular test case.
```
niemela, line 647:
```suggestion
The `data/sample/download` directory may contain anything, and none of its contents are validated. 
We recommend placing test-case-specific files in this directory, and using [attachments](#attachments) for testing tools and other downloads not associated with any particular test case.
```

niemela: Maybe even "strongly recommend"?

spec/2023-07-draft.md
```diff
@@ -596,70 +596,70 @@ The directory `invalid` can be organized into a tree-like structure similar to `
...
-All `data/sample/*.in` files are input-validated.
-
-When `data/sample/statement/` does not exist, `data/sample/*.ans` is output validated (i.e., must be accepted).
+All `data/sample/*.in` files are input-validated. For problems that are not interactive, all `data/sample/statement/*.in` are also input-validated.
 
-When `data/sample/statement/` does exists, and the problem is not interactive,
-all `data/sample/statement/*.in` are input validated, and `data/sample/statement/*.out` (with fallback to `data/sample/statement/*.ans`) are output validated (i.e., must be accepted).
+When `data/sample/statement/` does not exist, all test cases in `data/sample` must be accepted by the output validator, with the `.ans` file used as both the `answer_file` and `team_output`.
+ 
+When `data/sample/statement/` does exists, and the problem is not interactive, all test cases in `data/sample/statement/` must be accepted by the output validator. The `.ans` is used as the `answer_file`. The `.out` file is used as the `team_output`, if it exists. Otherwise the `.ans` is used as the `team_output`.
```
niemela, line 657:
```suggestion
When `data/sample/statement/` does exists, and the problem is not interactive, all test cases in `data/sample/statement/` must be accepted by the output validator. 
The `.ans` is used as the `answer_file`. 
The `.out` file is used as the `team_output`, if it exists. 
Otherwise the `.ans` is used as the `team_output`. 
```

spec/2023-07-draft.md
```diff
@@ -596,70 +596,70 @@ The directory `invalid` can be organized into a tree-like structure similar to `
...
+When `data/sample/statement/` does not exist, all test cases in `data/sample` must be accepted by the output validator, with the `.ans` file used as both the `answer_file` and `team_output`.
+ 
+When `data/sample/statement/` does exists, and the problem is not interactive, all test cases in `data/sample/statement/` must be accepted by the output validator. The `.ans` is used as the `answer_file`. The `.out` file is used as the `team_output`, if it exists. Otherwise the `.ans` is used as the `team_output`. 
 
 Files in `data/sample/download` are never validated, 
 although tooling may warn when there are inconsistencies between `download`, `statement`, and `data/sample/` itself.
 
-Validation can be customized by specifying `input_validator_args` and `output_validator_args` in `data/sample/testdata.yaml`.
+Validation can be customized by specifying `input_validator_args` and `output_validator_args` in `data/sample/testdata.yaml`. These arguments are used when validating test cases in both `/data/sample` and `/data/sample/statement`.
```
niemela, line 662:
```suggestion
Validation can be customized by specifying `input_validator_args` and `output_validator_args` in `data/sample/testdata.yaml`. 
These arguments are used when validating test cases in both `/data/sample` and `/data/sample/statement`.
```

niemela:
This is a *really* good cleanup. Some fairly minor comments above.

spec/2023-07-draft.md
evouga: Both cases require an `.ans`, so it is presence of the `.out` file that triggers the second case.

Consider for instance a problem with a custom output validator, where the `.ans` contains some certificate used to check correctness of the team output (rather than the "expected" team output). We don't want to show the certificate on the problem statement, so we put the expected output in `.out`, the certificate in `.ans`, and pass all three to the output validator.

In this case the `.in` and `.ans` files should be the same in both `data/sample` and `/data/sample/statement`. I can't remember the arguments for doing this rather than allowing "partial overriding" of test cases by e.g. providing just an `.out` file in `data/sample/statement`. But I think it was decided that keeping each folder self-contained (while encouraging symlinks) had less potential for mistakes.

niemela: > Both cases require an `.ans`, so it is presence of the `.out` file that triggers the second case.

Right, that's what I should have said :)

> In this case the `.in` and `.ans` files should be the same in both `data/sample` and `/data/sample/statement`. 

Exactly, so then they would be validated in `data/sample` already.

> I can't remember the arguments for doing this rather than allowing "partial overriding" of test cases by e.g. providing just an `.out` file in `data/sample/statement`. But I think it was decided that keeping each folder self-contained (while encouraging symlinks) had less potential for mistakes.

The problem with partial overriding is that you can't remove files that way.

While I was writing a paragraph here about how I still don't understand the use case for validating the `/statement` files, I remembered that this is how we would imeplement *not* running some samples, right? If I have some (possibly valid) samples that I don't want to run, I would put them in `data/sample/statement` and not in `data/sample/`, and that would be nice to be able to validate. That's it, right?

evouga: > The problem with partial overriding is that you can't remove files that way.

Yes, great point.

> If I have some (possibly valid) samples that I don't want to run, I would put them in data/sample/statement and not in data/sample/, and that would be nice to be able to validate. That's it, right?

Right.

evouga:
One use case that we discussed, which is not currently well-supported, are problems where we want to display truncated or too-small sample inputs in the problem statement which (intentionally) do not pass validation.

Do we still care about this case?

If so, then in `sample/statement` maybe we want to allow:
- `.in` (required, input-validated, available for download by default), the *actual* input for the test case
- `.displayed_in` (optional, not input-validated), the input shown in the problem statement
- `.ans` (required), the `answer_file` passed to output validation
- `.displayed_ans` (optional, available for download by default), what is shown in the problem statement and passed to the output validator as `team_output` (currently called `.out`)

spec/2023-07-draft.md
RagnarGrootKoerkamp: somehow mention that `made available` is for example: provided on the team computers in the homedir.

spec/2023-07-draft.md
```diff
@@ -596,70 +596,77 @@ The directory `invalid` can be organized into a tree-like structure similar to `
...
+We recommend placing test-case-specific files in this directory, and using [attachments](#attachments) for testing tools and other downloads not associated with any particular test case.
 Testing tools may warn when test cases that are listed in `data/sample` or `data/sample/statement` do not appear in `data/sample/download`,
 or when `data/sample/download` contains test cases that do not appear in one of the two other locations.
 
 #### Validation
 
-All `data/sample/*.in` files are input-validated.
-
-When `data/sample/statement/` does not exist, `data/sample/*.ans` is output validated (i.e., must be accepted).
+All `data/sample/*.in` files are input-validated. For problems that are not interactive, all `data/sample/statement/*.in` are also input-validated.
```
RagnarGrootKoerkamp, line 656: What's the reason for not validating `data/sample/statement/*.in` for interactive problems? We normally also have (or at least support) input validators for interactive problems.

RagnarGrootKoerkamp (approved):
Reads much better, thanks :)

@mzuenni, wdyt? Does this address #338? In particular, is/was your only concern the testing tool itself, and are you happy/OK with the `data/sample/{.,statement,download}` structure?

mzuenni:
> @mzuenni, wdyt? Does this address https://github.com/Kattis/problem-package-format/issues/338? In particular, is/was your only concern the testing tool itself, and are you happy/OK with the data/sample/{.,statement,download} structure?

looks good, the structure is fine :)

spec/2023-07-draft.md
evouga: I am not sure (the exception was in the original draft, so I kept it per Chesterton's Fence)---how do people usually use `.in`/`.out` in the interactive problem samples? Is the idea that people use these files to fake an interaction log?

RagnarGrootKoerkamp: Oh I see, I missed that it was already there. I don't quite remember the usecase for not validating these though :(

I suspect it's for problems with on-the-fly generated input, where we have an input validator for what we feed to the interactor (=generator), but we don't have a validator for it's output. And specifically `data/sample/statement/*.in` are _outputs_ of the interactor/generator.

niemela (approved):

Issue #354, opened 2024-11-28
Restrict directory names


thorehusfeldt:
In `2023-07-draft`, we have

> All file names for files included in the package must match the regexp
> `^[a-zA-Z0-9][a-zA-Z0-9_.-]{0,253}[a-zA-Z0-9]$`
> i.e., they must be of length at least 2, at most 255, consist solely of lower- or uppercase letters a–z, A–Z, digits 0–9, period, dash, or underscore, but must not begin or end with a period, dash, or underscore.

This allows `1.in`, `aa` and `foo.bar.baz_baz-baz` but disallows `1_`. Fine so far.

For directories, there is no restriction, as far as I can tell. Thus the following are all valid:
* the testgroup `secret/1/`,
* the testcase input file `secret/1/1.in`
* the input validator `input_validators/v` (containing `input_validators/v/validate.cpp` and `input_validators/v/validate`.

and so on. I think all of these are _fine_. 

What is also currently allowed, as far as I can tell are
* the testgroup `secret/_/`,
* the testgroup `secret/huge.connected`, 
* the testcase input file `secret/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/1.in`,
* the input validator `input_validators/[﷽`,
* the submission directory `submissions/😊/th.py`,

as well as several other references to validators, testgroups, testcases, visualisers or submissions that use “whatever the operating system supports”.

I propose to tighten the specification as follows:

> All names for directories included in the package must match the regexp
> `^([a-zA-Z0-9][a-zA-Z0-9_-]{0,253}[a-zA-Z0-9]|[a-zA-Z0-9])$`
> i.e., they must be of length at least 1, at most 255, consist solely of lower- or uppercase letters a–z, A–Z, digits 0–9, dash, or underscore (no period), but must not begin or end with a period, dash, or underscore.

Moreover, I would like to clarify that 

[...]

PR #357, opened 2025-02-21
Limit names of directories and test data groups

### LINKED ISSUE #299: Test cases can no longer be named `testdata`
Since we have introduces the `.yaml` file for test cases, it is no longer allowed to have a test case called `testdata`.

I don't think this is a problem(?), but we should either fix it somehow (I suggest we don't), *or* it needs to be mentioned (I suggest we do).

niemela:
Closes #299 and #354.

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -24,6 +24,12 @@ used for distributing and sharing problems for algorithmic programming contests 
   ```
   i.e., they must be of length at least 2, at most 255, consist solely of lower- or uppercase letters a–z, A–Z, digits 0–9, period, dash, or underscore,
   but must not begin or end with a period, dash, or underscore.
+- All directory names inside the package must match the regexp
+  ```regex
+  ^[a-zA-Z0-9]([a-zA-Z0-9_-]{0,253}[a-zA-Z0-9])?$
+  ```
+  i.e., they must be of length at least 1, at most 255, consist solely of lower- or uppercase letters a–z, A–Z, digits 0–9, dash, or underscore,
+  but must not begin or end with a dash, or underscore.
 - All text files for a problem must be UTF-8 encoded and not have a byte-order mark (BOM).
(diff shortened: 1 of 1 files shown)


eldering (approved):

spec/2023-07-draft.md
```diff
@@ -563,6 +570,8 @@ The `secret` directory must exist and the `secret` test group, or one of its des
 
 Test cases and groups will be used in lexicographical order on file base name.
 If a specific order is desired, a numbered prefix such as `00`, `01`, `02`, `03`, and so on, can be used.
+A subgroup must not have the same name as a test case in the same directory.
+E.g. if the file `data/secret/huge.in` exists then the directory `data/secret/huge/` must not, and vice versa.
```
Tagl, line 574:
```suggestion
For example, if the file `data/secret/huge.in` exists then the directory `data/secret/huge/` must not, and vice versa.
```

spec/2023-07-draft.md
```diff
@@ -24,6 +24,12 @@ used for distributing and sharing problems for algorithmic programming contests
   ```
   i.e., they must be of length at least 2, at most 255, consist solely of lower- or uppercase letters a–z, A–Z, digits 0–9, period, dash, or underscore,
   but must not begin or end with a period, dash, or underscore.
+- All directory names inside the package must match the regexp
+  ```regex
+  ^[a-zA-Z0-9]([a-zA-Z0-9_-]{0,253}[a-zA-Z0-9])?$
+  ```
+  i.e., they must be of length at least 1, at most 255, consist solely of lower- or uppercase letters a–z, A–Z, digits 0–9, dash, or underscore,
```
Tagl, line 31:
```suggestion
that is, they must be of length at least 1, at most 255, consist solely of lower- or uppercase letters a–z, A–Z, digits 0–9, dash, or underscore,
```

Issue #298, opened 2024-08-18
Make allowed compilers consistent


pehrsoderman:
At the moment, we allow c, c++ and Python 3 for "other programs", while we allow c++ and python 3 for Input Validators, output validators and Output Visualizers. This should be made consistent.

PR #359, opened 2025-02-22
Consistently allow C/C++ and Python 3 for programs

eldering:
Also fix a bug: `c++` is the standard name for a C++ compiler; `cpp` is the C preprocessor.

Closes: #298

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -110,7 +110,7 @@ If at least one of these two files is included:
 2. Then, the `run` file (which now exists, and is an executable binary or POSIX-compliant shell script)
    will be used as the validator or visualizer program.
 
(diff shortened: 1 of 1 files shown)


niemela (commented):
This feels to me to make it less clear that input validators must be written in C/C++ or Python 3, and it's not inconsistent with visualizers.

I'm not saying this is bad, but if this is good, shouldn't we do the same change for visualizers?

eldering:
> This feels to me to make it less clear that input validators must be written in C/C++ or Python 3, and it's not inconsistent with visualizers.
> 
> I'm not saying this is bad, but if this is good, shouldn't we do the same change for visualizers?

Indeed, I'd say we also modify it there. I can make it a bit more explicit that the "program" references adds requirements. I think it's better to not explicitly reiterate these (like written in C/C++ or Python 3) elsewhere, because that easily creates inconsistency like we currently had.

PR #360, opened 2025-02-22
Output validator tweaks

eldering:
Small rewording and layout fixes, but also a few slightly bigger changes:

- require that the validator can write to new/existing files in feedback_dir
- replace explicit reference to program languages by reference to "programs" section
- list all possible files under `feedback_dir`

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -970,52 +970,50 @@ If no floating-point tolerance has been set, floating-point tokens are treated j
 
 ### Invocation
 
-When invoked, the output validator will be passed at least three command line parameters and the output stream to validate on stdin.
-
-The validator should be possible to use as follows on the command line:
+The output validator must be invoked and must support being invoked as:
 ```sh
-<output_validator_program> input answer_file feedback_dir [additional_arguments] < team_output [ > team_input ]
+<output_validator_program> input_file answer_file feedback_dir [additional_arguments] < team_output [ > team_input ]
 ```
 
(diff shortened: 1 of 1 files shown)


spec/2023-07-draft.md
```diff
@@ -928,27 +928,18 @@ it is recommended to include the input visualizer source code in the directory `
...
+and provide it as the directory `output_validator/`.
+If no custom output validator is specified, the default validator is used.
 
-1.  The validator must give a judgement (see [Reporting a judgement](#reporting-a-judgement)).
-2.  The validator may give additional feedback,
-    e.g., an explanation of the judgement to humans (see [Reporting Additional Feedback](#reporting-additional-feedback)).
-
-A custom output validator is used if the problem requires more complicated output validation than what is provided by the default diff variant described below.
-It must be provided as the directory `output_validator/`.
-It must either be written in C++ or Python 3,
```
niemela, line 946: This requirement seems to have been dropped?

niemela (changes requested):
Looks very good!

Could we add back the C++ or Python 3 requirement?

spec/2023-07-draft.md
eldering: It's part of the requirements of a "program" referenced to, so I removed the duplicate (but slightly different, because C is missing) requirements here.

eldering: And rebased this PR on top.

niemela (approved):

PR #361, opened 2025-02-22
add valid_output

### LINKED ISSUE #355: `data/valid_outputs`
We have added `data/invalid_outputs` to check if the output validator correctly rejects some input. I think it would also be nice to easily ensure that it accepts some specific outputs.

For example consider https://open.kattis.com/problems/higherarithmetic (we are given some numbers and can add parenthesis and `+`,`-`,`*` and want to maximize the value of the expression).
Here it would be nice to ensure that these two outputs work interchangeable:
```
in: 2 2
ans: 2*2
out: 2+2
```
and
```
in: 2 2
ans: 2+2
out: 2*2
```
The same can more or less already be achieved by adding a specific submission to ensure this behavior. However, that is a lot more work, less expressive, and adds unnecessary overhead since those submission will be run on all test cases.

Some more generic examples could be:
- ensuring that the checker is not space sensitive
- ensuring that the checker is not case sensitive

---
The  semantics would be the same as for `data/invalid_outputs`, i.e., requiring a `.in`, `.ans`, and `.out` file. However, now we want to get AC as verdict.

We could also introduce `data/valid_inputs` and `data/valid_answers` in the same way. However, i think they are not really useful. We don't need to check that a test case that does not exists would be accepted.

mzuenni:
Closes #355 

One think i noticed: 
> The test cases in `(in)valid_output` describe invalid outputs for **_non-interactive_** problems.

Maybe we should also exclude multi-pass problems?

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -606,8 +606,8 @@ Each `tc.in` under `invalid_input` must be rejected by at least one input valida
 
 The test cases in `invalid_output` describe invalid outputs for non-interactive problems.
 They consist of three files.
-The input file `tc.in` must _pass_ input validation.
-The output file `tc.out` must _fail_ output validation for the given answer file `tc.ans`.
+The input file `tc.in`, which must contain valid input.
+The output file `tc.out` must fail output validation for the given answer file `tc.ans`.
 
 In particular, for an existing feedback directory `dir`,
 ```bash
@@ -615,7 +615,7 @@ In particular, for an existing feedback directory `dir`,
 <output_validator_program> tc.in tc.ans dir [arguments] < tc.out # MUST FAIL
 ```
 
-The directory `invalid_output` can be organized into a tree-like structure similar to `secret` and contain arguments in `testdata.yaml` files that are passed to the validators.
+The directory `invalid_output` must be organized into a tree-like structure similar to `secret` and may contain arguments in `testdata.yaml` files that are passed to the validators.
 
 ### Valid Output
 
@@ -623,8 +623,8 @@ The `data` directory may contain a directory of test cases that must pass valida
 Their goal is to ensure the integrity and quality of validation programs.
 The test cases in `valid_output` describe valid outputs for non-interactive problems.
 They consist of three files.
-The input file `tc.in` must _pass_ input validation.
-The output file `tc.out` must _pass_ output validation for the given answer file `tc.ans`.
+The input file `tc.in`, which must contain valid input.
+The output file `tc.out` must pass output validation for the given answer file `tc.ans`.
 
 In particular, for an existing feedback directory `dir`,
 ```bash
@@ -632,7 +632,7 @@ In particular, for an existing feedback directory `dir`,
 <output_validator_program> tc.in tc.ans dir [arguments] < tc.out # MUST PASS
 ```
 
-The directory `valid_output` can be organized into a tree-like structure similar to `secret` and contain arguments in `testdata.yaml` files that are passed to the validators.
+The directory `invalid_output` must be organized into a tree-like structure similar to `secret` and may contain arguments in `testdata.yaml` files that are passed to the validators.
 
 ### Samples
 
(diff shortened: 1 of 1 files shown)


niemela (commented):
This looks good.

That said... didn't we kinda conclude it's not needed? I'm not *against* adding it, but I'm not at all convinced that it's very useful either...

mzuenni:
We concluded that `valid_input` is pointless? `valid_output` is very useful to test the output validator.

niemela:
> We concluded that `valid_input` is pointless? `valid_output` is very useful to test the output validator.

Yes, of course. I'm sorry, I'm very tired (and a little bit stressed).

niemela (approved):

Tagl:
What about output validator flags? How does this work with those?

Tagl:
Oh I see it now, it has the same structure as the data folder.

Tagl (approved):

mpsijm:
Can this PR be linked to #355 ?

spec/2023-07-draft.md
```diff
@@ -614,7 +615,24 @@ In particular, for an existing feedback directory `dir`,
...
-The directory `invalid` can be organized into a tree-like structure similar to `secret` and contain arguments in `testdata.yaml` files that are passed to the validators.
+The directory `invalid_output` can be organized into a tree-like structure similar to `secret` and contain arguments in `testdata.yaml` files that are passed to the validators.
+
+### Valid Output
+
+The `data` directory may contain a directory of test cases that must pass validation.
+Their goal is to ensure the integrity and quality of validation programs.
+The test cases in `valid_output` describe valid outputs for non-interactive problems.
+They consist of three files.
+The input file `tc.in` must _pass_ input validation.
```
eldering, line 626: [I've not followed this issue in detail, so I may overlook something.]
Doesn't this input file and a statement that it must pass *input* validation contradict the section title?

spec/2023-07-draft.md
```diff
@@ -614,7 +615,24 @@ In particular, for an existing feedback directory `dir`,
...
+The output file `tc.out` must _pass_ output validation for the given answer file `tc.ans`.
+
+In particular, for an existing feedback directory `dir`,
+```bash
+<output_validator_program> tc.in tc.ans dir [arguments] < tc.ans # MUST PASS
+<output_validator_program> tc.in tc.ans dir [arguments] < tc.out # MUST PASS
+```
+
+The directory `valid_output` can be organized into a tree-like structure similar to `secret` and contain arguments in `testdata.yaml` files that are passed to the validators.
```
eldering, line 635:
```suggestion
The directory `valid_output` must be organized into a tree-like structure similar to `secret` and may contain arguments in `testdata.yaml` files that are passed to the validators.
```
Of course this implies that subdirectories are not required, just that if you add them, they have to follow the testdata directory structure.

spec/2023-07-draft.md
```diff
@@ -614,7 +615,24 @@ In particular, for an existing feedback directory `dir`,
...
+The directory `invalid_output` can be organized into a tree-like structure similar to `secret` and contain arguments in `testdata.yaml` files that are passed to the validators.
+
+### Valid Output
+
+The `data` directory may contain a directory of test cases that must pass validation.
+Their goal is to ensure the integrity and quality of validation programs.
+The test cases in `valid_output` describe valid outputs for non-interactive problems.
+They consist of three files.
+The input file `tc.in` must _pass_ input validation.
+The output file `tc.out` must _pass_ output validation for the given answer file `tc.ans`.
```
eldering, line 627: Nitpick: is there a reason to make "pass" italic?

spec/2023-07-draft.md
mzuenni: No. Calling the output validator with an invalid input file is just undefined behaviour.

spec/2023-07-draft.md
mzuenni: this is the same sentence as in section "Invalid Output", if we change this it should be changed in both

spec/2023-07-draft.md
mzuenni: this is the same format as in section "Invalid Output", if we change this it should be changed in both

spec/2023-07-draft.md
eldering: Ok, but then I think it would be better to rephrase it as
```suggestion
The input file `tc.in`, which must contain valid input.
```
so that it is clear that no active validation is assumed to be performed on it here.

spec/2023-07-draft.md
eldering: Sure, I'd say: let's change it in both places then. Feel free to do now. Otherwise, I'm planning to make a pass over the whole document, so will then change it.

PR #362, opened 2025-02-23
Add flag to configure whether submissions can only read or both read …

### LINKED ISSUE #333: Comparing output to files (as opposed to stdout)
Let's say you wanted to define the following problem (and I would argue that something like this would be useful in an educational setting):
> Read several numbers from `numbers.txt`, one number per line. Write their sum to `sum.txt` and their product to `product.txt`.
 
We can define this problem in the current version. You would put the `numbers.txt`-files as `<testcase>.files/numbers.txt`. For validating you would have to use a custom validator. You would put the sum and product in `<testcase>.ans` and the validator would get `.ans` and usual, and would have to know how to parse it and what should go in what file. This is not very different from any validator that gets some kind of secret from the `.ans` file.

The implication is that you will *always* have to use a custom validator when you want the submission to write output to a file.

If we don't want this we could introduce something like an `.ansfiles` directory. The default validator would then for each file in `.ansfiles` verify that this exists and is equal in the working directory of the submission after running it.

My sense is that requiring a custom validator whenever you use file output is acceptable, and that adding something like `.ansfiles` is more complication than it's worth. So, I suggest that we don't do anything. I think @Tagl also agrees?

Thoughts?

---

A related question is allowing writing at all.

I would argue that we do want the kind of problems suggested above, and then we *have* to allow writing to file (that is in fact the point). That said, in the normal stdin/stdout case it would be nice if the system is a at least allowed to forbid writing to files. Should there be a setting that you are required (as a problem setter) to use to explicitly allow file writes? This would also allow a system to know at install time that a problem will require writing to file, and thus be able to refuse the problem at install time, rather than failing at run time.

Tagl:
…and write to files in their working directory.

This makes changes based on the discussion in #333 
Closes #333

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -155,7 +155,9 @@ Key                                               | Type                        
 [limits](#limits)                                 | Map with keys as defined below                | No        | See below
 [keywords](#keywords)                             | Sequence of strings                           | No        |
 [languages](#languages)                           | String or sequence of strings                 | No        | `all`
+[allow_file_writing](#allow-file-writing)         | Boolean                                       | No        | `false`
 [constants](#constants)                           | Map of strings to int, float, or string       | No        |
+
 
 ### Problem format version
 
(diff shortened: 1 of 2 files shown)


Tagl:
Closes #333

niemela (commented):
Does "write files" include "create files"?

Regardless of what the answer is, I think it should be clarified.

Tagl:
It does

Tagl:
I have clarified it

niemela (approved):

Issue #348, opened 2024-10-10
Clarify the behavior of the default output validator and the semantics of the default output validator flags


evouga:
The behavior of the default output validator is not well specified.

1. What output encoding does it require? UTF-8 without BOM? How is the output tokenized? How is the "type" of the token (floating-point vs. string, I guess? Maybe integer too, see below?) determined?

2. Are integers compared as strings? (With leading zeros, hexadecimal encoding, etc. being judged as incorrect) or is there some special handling for "integer-type" tokens?

3. `case_sensitive`: notoriously [underspecified](https://www.b-list.org/weblog/2018/nov/26/case/). How exactly is case-insensitive comparison performed? I assume that this flag only applies to tokens of "string type" and not floating-point?

4. `space_change_sensitive`: The default behavior is mentioned in a parenthetical remark `(the default is that any sequence of 1 or more whitespace characters are equivalent)` but this is incomplete since it does not mention leading or trailing whitespace (which I understand the current default validator ignores by default). Moreover `indicates that changes in the amount of whitespace should be rejected` is imprecise: do changes in the *type* of whitespace (newlines to spaces e.g.) cause submissions to be rejected? What is a whitespace anyway? The 25 whitespace Unicode code points?

Tagl:
I agree this needs a better explanation.

> 1. What output encoding does it require? UTF-8 without BOM? How is the output tokenized? How is the "type" of the token (floating-point vs. string, I guess? Maybe integer too, see below?) determined?

The current behavior is: if it can be parsed as a float, and is not nan or inf, then it is a float. This is obviously not great because integers can be parsed as floats and although you might want to allow precision issues with `3.1416` you might not want it for `3`.

> 2. Are integers compared as strings? (With leading zeros, hexadecimal encoding, etc. being judged as incorrect) or is there some special handling for "integer-type" tokens?

If float tolerance is set, then the answer for 1 applies, but otherwise string comparison is used. I'd argue that we want to change the behavior of 1 so integers use string comparison always. Do you oppose string comparison for integers?

> 3\. `case_sensitive`: notoriously [underspecified](https://www.b-list.org/weblog/2018/nov/26/case/). How exactly is case-insensitive comparison performed? I assume that this flag only applies to tokens of "string type" and not floating-point?

[...]

PR #368, opened 2025-02-25
Clarify behavior of the default output validator

evouga:
Addresses https://github.com/Kattis/problem-package-format/issues/348

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -993,7 +993,8 @@ If a floating-point tolerance has been set, the default output validator will at
 For each success, the token is compared to the submission output token using the following floating-point semantics. 
 (If no floating-point tolerance has been set, floating-point tokens are [compared as strings](#comparing-string-tokens).)
 
-If the submission output token cannot be parsed as floating-point, the validator rejects the submission as incorrect. Otherwise, if `s` is the submission output floating-point value and `a` is the answer file value:
+If the submission output token cannot be parsed as floating-point, the validator rejects the submission as incorrect. 
+Otherwise, if `s` is the submission output floating-point value and `a` is the answer file value:
 - if an absolute tolerance `ε` has been set, the token is accepted if `|s-a| ≤ ε`;
 - if a relative tolerance `ε` has been set, the token is accepted if `|s-a| ≤ ε|a|`;
 - when supplying both a relative and an absolute tolerance, the token is accepted if it is within either of the two tolerances.
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -1012,7 +1012,9 @@ Otherwise, the submitted output token is accepted if it matches the answer file 
 
 #### Whitespace Sensitivity
 
-By default, whitespace in the submission output is only used for tokenization and is otherwise ignored. As a consequence, the default output validator is lenient with regard to leading and trailing whitespace, and it treats any sequence of one or more whitespace characters in between tokens as equivalent to any other such sequence.
+By default, whitespace in the submission output is only used for tokenization and is otherwise ignored. 
+As a consequence, the default output validator is lenient with regard to leading and trailing whitespace,
+and it treats any sequence of one or more whitespace characters in between tokens as equivalent to any other such sequence.
 
 If the `space_change_sensitive` argument is set, the default output validator will instead reject any submission output whose whitespace sequences (including leading and trailing) differ from those in the answer file in type or amount.
 
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -990,8 +990,6 @@ In its default mode, the default output validator then ignores the whitespace an
 (If there is a disagreement in the number of tokens in the two files, the validator rejects the submission instead.)
 The validator may also reject any submission output that is unreasonably large (including due to containing unreasonable amounts of whitespace).
 
-A contest system may support a superset of ASCII, including multi-byte encodings such as UTF-8, so long as the default output validator conforms to this specification whenever both the submission output and answer file consist only of the 128 standard ASCII characters.
-
 #### Comparing Floating-Point Tokens
 
 If a floating-point tolerance has been set, the default output validator will attempt to parse each answer file token as floating-point (see [general requirements](#general-requirements) for details). 
@@ -1013,7 +1011,7 @@ For problems containing a mix of integer and floating-point output, a custom out
 
 If `case_sensitive` is specified, the two tokens must match exactly, i.e. consist of exactly the same character sequence.
 
-Otherwise, the submitted output token is accepted if it matches the answer file token up to case. The default output validator must treat uppercase ASCII letters `A`--`Z` as equivalent to their lowercase counterparts whenever both the submission ouput and answer file consist only of the 128 standard ASCII characters. A contest system that supports a superset of ASCII may implement more lenient, locale-specific case-insensitive string comparison.
+Otherwise, the submitted output token is accepted if it matches the answer file token up to case. The default output validator treats uppercase ASCII letters `A`--`Z` as equivalent to their lowercase counterparts.
 
 #### Whitespace Sensitivity
 
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -1001,7 +1001,8 @@ Otherwise, if `s` is the submission output floating-point value and `a` is the a
 
 It is an error to provide any of the `float_relative_tolerance`, `float_absolute_tolerance`, or `float_tolerance` arguments more than once, or to provide a `float_tolerance` alongside `float_relative_tolerance` and/or `float_absolute_tolerance`.
 
-Note that if a floating-point tolerance has been set, the default output validator will parse exact integers as floating-point and apply the above semantics to them. For problems containing a mix of integer and floating-point output, a custom output validator must be used if exact comparison of the integer tokens is required.
+Note that if a floating-point tolerance has been set, the default output validator will parse exact integers as floating-point and apply the above semantics to them. 
+For problems containing a mix of integer and floating-point output, a custom output validator must be used if exact comparison of the integer tokens is required.
 
 #### Comparing String Tokens
 
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -30,7 +30,7 @@ used for distributing and sharing problems for algorithmic programming contests 
   ```
 that is, they must be of length at least 1, at most 255, consist solely of lower- or uppercase letters a–z, A–Z, digits 0–9, dash, or underscore,
   but must not begin or end with a dash, or underscore.
-  - All text files for a problem must be UTF-8 encoded and not have a byte-order mark (BOM).
+- All text files for a problem must be UTF-8 encoded and not have a byte-order mark (BOM).
 - All text files must have Unix-style line endings (newline/LF byte only).
   Note that LF is line-ending and not line-separating in POSIX, which means that all non-empty text files must end with a newline.
 - Natural language (for example, in the [problem statement](#problem-statements) filename) must be specified as 2-letter [ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) code if it exists, otherwise as a 3-letter code from ISO 639.
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -1006,7 +1006,7 @@ Otherwise, the submitted output token is accepted if it matches the answer file 
 
 #### Whitespace Sensitivity
 
-By default, whitespace in the submission output is ignored by the default output validator after tokenization. As a consequence, the validator is lenient with regard to leading and trailing whitespace, and it treats any sequence of one more whitespace characters in between tokens as equivalent to any other such sequence.
+By default, whitespace in the submission output is ignored by the default output validator after tokenization. As a consequence, the validator is lenient with regard to leading and trailing whitespace, and it treats any sequence of one or more whitespace characters in between tokens as equivalent to any other such sequence.
 
 If the `space_change_sensitive` argument is set, the default output validator will instead reject any submission output whose whitespace sequences (including leading and trailing) differ from those in the answer file in type or amount.
 
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -992,7 +992,7 @@ The validator may also reject any submission output that is unreasonably large (
 
 #### Comparing Floating-Point Tokens
 
-If a floating-point tolerance has been set, the default output validator will attempt to parse each answer file token as floating-point (see [general requirements](#general-requirements) for details). 
+If a floating-point tolerance has been set, the default output validator will attempt to parse each answer file token as a floating-point number (see [general requirements](#general-requirements) for details). 
 For each success, the token is compared to the submission output token using the following floating-point semantics. 
 (If no floating-point tolerance has been set, floating-point tokens are [compared as strings](#comparing-string-tokens).)
 
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -977,7 +977,7 @@ The default output validator supports the following command-line arguments:
 Arguments                    | Description
 ---------------------------- | -----------
 `case_sensitive`             | indicates that comparisons should be case-sensitive (see below for details).
-`space_change_sensitive`     | indicates that differences in type or amount of of whitespace should be rejected (see below for details).
+`space_change_sensitive`     | indicates that differences in type or amount of whitespace should be rejected (see below for details).
 `float_relative_tolerance ε` | indicates that floating-point tokens should be accepted if they are within relative error ≤ ε (see below for details).
 `float_absolute_tolerance ε` | indicates that floating-point tokens should be accepted if they are within absolute error ≤ ε (see below for details).
 `float_tolerance ε`          | short-hand for applying ε as both relative and absolute tolerance.
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -973,7 +973,7 @@ and optionally [report additional feedback](#reporting-additional-feedback).
 
 The default output validator is essentially a beefed-up diff that can be used in the common case where the output validator needs only compare the output of a submitted program against a trusted judge reference solution.
 It parses the submission output and answer files as UTF-8 (without byte-order mark) and tokenizes the files by splitting on sequences of one or more consecutive Unicode whitespace code points. 
-In its default mode, the default output validator then ignores the whitespace and compares the submission output and answer files token by token. If there is a disagreement in the number of tokens in the two files, the validator rejects the submission with a Wrong Answer verdict. The validator may also reject as Wrong Answer any submission output that is unreasonable large (including due to containing unreasonable amounts of whitespace).
+In its default mode, the default output validator then ignores the whitespace and compares the submission output and answer files token by token. If there is a disagreement in the number of tokens in the two files, the validator rejects the submission. The validator may also reject any submission output that is unreasonably large (including due to containing unreasonable amounts of whitespace).
 
 The default output validator supports the following command-line arguments to control the above process:
 
@@ -989,7 +989,7 @@ Arguments                    | Description
 
 If a floating-point tolerance has been set, the default output validator will attempt to parse each answer file token as floating-point (see [general requirements](#general-requirements) for details). For each success, the token is compared to the submission output token using the following floating-point semantics. (If no floating-point tolerance has been set, floating-point tokens are [compared as strings](#comparing-string-tokens).)
 
-If the submission output token cannot be parsed as floating-point, the validator returns a Wrong Answer verdict. Otherwise, if `s` is the submission output floating-point value and `a` is the answer file value:
+If the submission output token cannot be parsed as floating-point, the validator rejects the submission as incorrect. Otherwise, if `s` is the submission output floating-point value and `a` is the answer file value:
 - if an absolute tolerance `ε` has been set, the token is accepted if `|s-a| ≤ ε`;
 - if a relative tolerance `ε` has been set, the token is accepted if `s` and `a` are both exactly zero, or if `|s-a|/|a| ≤ ε`;
 - when supplying both a relative and an absolute tolerance, the token is accepted if it is within either of the two tolerances.
@@ -1000,7 +1000,7 @@ Note that if a floating-point tolerance has been set, the default output validat
 
 #### Comparing String Tokens
 
-If `case_sensitive` is specified, the two tokens must match exactly, i.e. consist of exactly the same code point sequences. 
+If `case_sensitive` is specified, the two tokens must match exactly, i.e. consist of exactly the same code point sequence. 
 
 Otherwise, the submitted output token is accepted if it matches the answer file token up to case. The default output validator must treat uppercase ASCII letters `A`--`Z` as equivalent to their lowercase counterparts. A contest system may additionally implement more lenient, locale-specific case-insensitive string comparison.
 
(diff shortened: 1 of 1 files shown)


spec/2023-07-draft.md
```diff
@@ -972,24 +972,43 @@ and optionally [report additional feedback](#reporting-additional-feedback).
...
-When a floating-point tolerance has been set, any valid formatting of floating-point numbers is accepted for floating-point tokens.
-So, for instance, if a token in the answer file says `0.0314`, a token of `3.14000000e-2` in the output file would be accepted.
+#### Comparing Floating-Point Tokens
+
+If a floating-point tolerance has been set, the default output validator will attempt to parse each answer file token as floating-point (see [general requirements](#general-requirements) for details). For each success, the token is compared to the submission output token using the following floating-point semantics. (If no floating-point tolerance has been set, floating-point tokens are [compared as strings](#comparing-string-tokens).)
+
+If the submission output token cannot be parsed as floating-point, the validator rejects the submission as incorrect. Otherwise, if `s` is the submission output floating-point value and `a` is the answer file value:
+- if an absolute tolerance `ε` has been set, the token is accepted if `|s-a| ≤ ε`;
+- if a relative tolerance `ε` has been set, the token is accepted if `s` and `a` are both exactly zero, or if `|s-a|/|a| ≤ ε`;
```
eldering, line 994: This leads to division by zero if `a=0` and `s!=0`, although that implicitly means that the condition is not true.

I was comparing with our (old?) copy of the reference implementation and there I don't see the check that `s` and `a` are both exactly zero.

spec/2023-07-draft.md
```diff
@@ -972,24 +972,43 @@ and optionally [report additional feedback](#reporting-additional-feedback).
...
+- when supplying both a relative and an absolute tolerance, the token is accepted if it is within either of the two tolerances.
 
 It is an error to provide any of the `float_relative_tolerance`, `float_absolute_tolerance`, or `float_tolerance` arguments more than once, or to provide a `float_tolerance` alongside `float_relative_tolerance` and/or `float_absolute_tolerance`.
 
-If no floating-point tolerance has been set, floating-point tokens are treated just like any other token and must match exactly.
+Note that if a floating-point tolerance has been set, the default output validator will parse exact integers as floating-point and apply the above semantics to them. For problems containing a mix of integer and floating-point output, a custom output validator must be used if exact comparison of the integer tokens are required.
```
eldering, line 999:
```suggestion
Note that if a floating-point tolerance has been set, the default output validator will parse exact integers as floating-point and apply the above semantics to them. For problems containing a mix of integer and floating-point output, a custom output validator must be used if exact comparison of the integer tokens is required.
```

spec/2023-07-draft.md
```diff
@@ -972,24 +972,43 @@ and optionally [report additional feedback](#reporting-additional-feedback).
...
+
+#### Comparing String Tokens
+
+If `case_sensitive` is specified, the two tokens must match exactly, i.e. consist of exactly the same code point sequence. 
+
+Otherwise, the submitted output token is accepted if it matches the answer file token up to case. The default output validator must treat uppercase ASCII letters `A`--`Z` as equivalent to their lowercase counterparts. A contest system may additionally implement more lenient, locale-specific case-insensitive string comparison.
+
+#### Whitespace Sensitivity
+
+By default, whitespace in the submission output is ignored by the default output validator after tokenization. As a consequence, the validator is lenient with regard to leading and trailing whitespace, and it treats any sequence of one or more whitespace characters in between tokens as equivalent to any other such sequence.
```
eldering, line 1009: I don't understand the first sentence: after tokenization there's no way anymore to (not) ignore whitespace, I'd say? A suggestion:
```suggestion
By default, the validator treats any sequence of one or more whitespace characters in between tokens as equivalent to any other such sequence. Since this includes newlines, the default validator is also lenient with regard to leading and trailing whitespace.
```

evouga: I don't think that quite covers it, since there could be zero whitespaces before the first token vs. one or more whitespaces, and that would still be accepted.

spec/2023-07-draft.md
evouga: Yes, good point. So when `s==a==0`, should that count as failing the relative error tolerance?

spec/2023-07-draft.md
niemela: I agree with @evouga . What @eldering says, is how we have usually explained it (and it is the logic that name `space_change_sensitive` is based on), but what we *actually* do is what @evouga is explaining, i.e. we tokenize on whitespace, which is the same as "amount of whitespace does not matter" **but also** "leading and trailing whitespace does not matter".

spec/2023-07-draft.md
RagnarGrootKoerkamp: Both 0 should pass. It would if you write it using a multiplication instead: |s-a|<=eps*a, so maybe we should just do that?

evouga:
@RagnarGrootKoerkamp @eldering I've pushed some changes to address your comments; please take a look.

eldering (approved):

spec/2023-07-draft.md
```diff
@@ -972,24 +972,43 @@ and optionally [report additional feedback](#reporting-additional-feedback).
 ### Default Output Validator Specification
 
 The default output validator is essentially a beefed-up diff that can be used in the common case where the output validator needs only compare the output of a submitted program against a trusted judge reference solution.
-In its default mode, it tokenizes the output and answer files and compares them token by token.
-It supports the following command-line arguments to control how tokens are compared.
+It parses the submission output and answer files as UTF-8 (without byte-order mark) and tokenizes the files by splitting on sequences of one or more consecutive Unicode whitespace code points.
```
niemela, line 975: Is "Unicode whitespace code points" well defined? There are "space separator characters", and there is the "Whitespace property", which I believe is had by all "space separator characters" and some "control characters".

The only (seemingly correct) list I can find is this:
https://en.wikipedia.org/wiki/Whitespace_character#Unicode

Also, do we know that the reference implementation does this?

spec/2023-07-draft.md
```diff
@@ -972,24 +972,43 @@ and optionally [report additional feedback](#reporting-additional-feedback).
 ### Default Output Validator Specification
 
 The default output validator is essentially a beefed-up diff that can be used in the common case where the output validator needs only compare the output of a submitted program against a trusted judge reference solution.
-In its default mode, it tokenizes the output and answer files and compares them token by token.
-It supports the following command-line arguments to control how tokens are compared.
+It parses the submission output and answer files as UTF-8 (without byte-order mark) and tokenizes the files by splitting on sequences of one or more consecutive Unicode whitespace code points. 
+In its default mode, the default output validator then ignores the whitespace and compares the submission output and answer files token by token. If there is a disagreement in the number of tokens in the two files, the validator rejects the submission. The validator may also reject any submission output that is unreasonably large (including due to containing unreasonable amounts of whitespace).
```
niemela, line 976:
```suggestion
In its default mode, the default output validator then ignores the whitespace and compares the submission output and answer files token by token. 
If there is a disagreement in the number of tokens in the two files, the validator rejects the submission. 
The validator may also reject any submission output that is unreasonably large (including due to containing unreasonable amounts of whitespace).
```

niemela, line 976: This makes it sounds like we only care about the **number** of tokens, which is obviously not the case.

spec/2023-07-draft.md
```diff
@@ -972,24 +972,43 @@ and optionally [report additional feedback](#reporting-additional-feedback).
...
 `float_absolute_tolerance ε` | indicates that floating-point tokens should be accepted if they are within absolute error ≤ ε (see below for details).
 `float_tolerance ε`          | short-hand for applying ε as both relative and absolute tolerance.
 
-When supplying both a relative and an absolute tolerance, the semantics are that a token is accepted if it is within either of the two tolerances.
-When a floating-point tolerance has been set, any valid formatting of floating-point numbers is accepted for floating-point tokens.
-So, for instance, if a token in the answer file says `0.0314`, a token of `3.14000000e-2` in the output file would be accepted.
+#### Comparing Floating-Point Tokens
+
+If a floating-point tolerance has been set, the default output validator will attempt to parse each answer file token as floating-point (see [general requirements](#general-requirements) for details). For each success, the token is compared to the submission output token using the following floating-point semantics. (If no floating-point tolerance has been set, floating-point tokens are [compared as strings](#comparing-string-tokens).)
```
niemela, line 990:
```suggestion
If a floating-point tolerance has been set, the default output validator will attempt to parse each answer file token as floating-point (see [general requirements](#general-requirements) for details). 
For each success, the token is compared to the submission output token using the following floating-point semantics. 
(If no floating-point tolerance has been set, floating-point tokens are [compared as strings](#comparing-string-tokens).)
```

spec/2023-07-draft.md
```diff
@@ -972,24 +972,43 @@ and optionally [report additional feedback](#reporting-additional-feedback).
...
+
+If the submission output token cannot be parsed as floating-point, the validator rejects the submission as incorrect. Otherwise, if `s` is the submission output floating-point value and `a` is the answer file value:
```
niemela, line 992:
```suggestion
If the submission output token cannot be parsed as floating-point, the validator rejects the submission as incorrect. 
Otherwise, if `s` is the submission output floating-point value and `a` is the answer file value:
```

spec/2023-07-draft.md
```diff
@@ -972,24 +972,43 @@ and optionally [report additional feedback](#reporting-additional-feedback).
...
+- if an absolute tolerance `ε` has been set, the token is accepted if `|s-a| ≤ ε`;
+- if a relative tolerance `ε` has been set, the token is accepted if `|s-a| ≤ ε|a|`;
+- when supplying both a relative and an absolute tolerance, the token is accepted if it is within either of the two tolerances.
 
 It is an error to provide any of the `float_relative_tolerance`, `float_absolute_tolerance`, or `float_tolerance` arguments more than once, or to provide a `float_tolerance` alongside `float_relative_tolerance` and/or `float_absolute_tolerance`.
 
-If no floating-point tolerance has been set, floating-point tokens are treated just like any other token and must match exactly.
+Note that if a floating-point tolerance has been set, the default output validator will parse exact integers as floating-point and apply the above semantics to them. For problems containing a mix of integer and floating-point output, a custom output validator must be used if exact comparison of the integer tokens is required.
```
niemela, line 999:
```suggestion
Note that if a floating-point tolerance has been set, the default output validator will parse exact integers as floating-point and apply the above semantics to them. 
For problems containing a mix of integer and floating-point output, a custom output validator must be used if exact comparison of the integer tokens is required.
```

spec/2023-07-draft.md
```diff
@@ -972,24 +972,43 @@ and optionally [report additional feedback](#reporting-additional-feedback).
...
+
+#### Comparing String Tokens
+
+If `case_sensitive` is specified, the two tokens must match exactly, i.e. consist of exactly the same code point sequence. 
+
+Otherwise, the submitted output token is accepted if it matches the answer file token up to case. The default output validator must treat uppercase ASCII letters `A`--`Z` as equivalent to their lowercase counterparts. A contest system may additionally implement more lenient, locale-specific case-insensitive string comparison.
```
niemela, line 1005:
```suggestion
Otherwise, the submitted output token is accepted if it matches the answer file token up to case. 
The default output validator must treat uppercase ASCII letters `A`--`Z` as equivalent to their lowercase counterparts. 
A contest system may additionally implement more lenient, locale-specific case-insensitive string comparison.
```

niemela: I'm not sure about "A contest system may additionally implement more lenient, locale-specific case-insensitive string comparison"?

I think I would rather say that systems may defined additional options for the default validator?

spec/2023-07-draft.md
```diff
@@ -972,24 +972,43 @@ and optionally [report additional feedback](#reporting-additional-feedback).
...
+
+#### Whitespace Sensitivity
+
+By default, whitespace in the submission output is only used for tokenization and is otherwise ignored. As a consequence, the default output validator is lenient with regard to leading and trailing whitespace, and it treats any sequence of one or more whitespace characters in between tokens as equivalent to any other such sequence.
```
niemela, line 1009:
```suggestion
By default, whitespace in the submission output is only used for tokenization and is otherwise ignored. 
As a consequence, the default output validator is lenient with regard to leading and trailing whitespace,
and it treats any sequence of one or more whitespace characters in between tokens as equivalent to any other such sequence.
```

niemela (changes requested):

spec/2023-07-draft.md
evouga: In a perfect world, I agree. However, in the linked issue there was a desire for the default (case-insensitive) behavior to support e.g. Icelandic and Swedish accented characters.

spec/2023-07-draft.md
evouga: The whitespace property is probably what we want, though there are some unusual corner cases (i.e. the zero-width space U+200B and Braille pattern blank U+2800 will not count as whitespace).

evouga: > Also, do we know that the reference implementation does this?

No, but I suspect that the reference solution's support of Unicode in general is suspect. Is it OK to require adding support?

niemela: > No, but I suspect that the reference solution's support of Unicode in general is questionable.

Why do you think that? Proper casing with Unicode is complicated (and as we have discussed locale-dependent), but plain diffing should work fine?

> Is it OK to require adding support?

I think so, yes. We should require that it's (in some reasonable sense) backwards compatible, and we should understand exactly what we are adding. I.e. it should be well defined **and** we should know what the current implementation does.

spec/2023-07-draft.md
niemela: Yes, but I don't think that should be done by saying that not `case_sensitive` can optionally mean something very different on some systems. We should either expand the meaning of not `case_sensitive` so that we get nice casing for alphabets outside of ASCII, or add some other setting for locale specific casing. I think the former is impossible, because (I think) the set off all locale rules are contradictory, and I'm not sure the latter is wort it, but maybe?

spec/2023-07-draft.md
simonlindholm: I don't understand why we feel the need to support anything outside of ASCII here. Feels like it's just making things more complicated for no reason?

evouga: It is already current practice to include non-ASCII characters in ICPC problems. See the linked issue where there are reports of using Swedish and Icelandic accented characters in problem output. I also remember seeing a problem that asked contestants to print emojis.

Requiring anyone who wants to use more than the 26 letters of the English alphabet to write a custom validator seems *really* silly to me in 2025.

evouga: Or are you talking specifically about whitespaces? If we allow Unicode in problem output, it seems sensible to treat all Unicode whitespaces (such as a non-breaking space, perhaps copy-pasted from the contest web site) as whitespace?

Tagl: I don't think the default validator **needs** to support more. It would certainly be very nice if it did though, but I do not think it is a priority as is.

simonlindholm: I'm talking specifically about whitespace. I find it hard to believe that contestants would expect non-ASCII whitespace to be interchangable with ASCII whitespace.

> such as a non-breaking space, perhaps copy-pasted from the contest web site

AFAIK browsers convert non-breaking spaces to normal ones upon copy.

spec/2023-07-draft.md
evouga: I'm not opposed to a `locale_dependent` option. Are people willing to make the default ASCII-case-insensitive-only?

spec/2023-07-draft.md
Tagl: > Or are you talking specifically about whitespaces? If we allow Unicode in problem output, it seems sensible to treat all Unicode whitespaces (such as a non-breaking space, perhaps copy-pasted from the contest web site) as whitespace?

Agreed, but this is a change to the behavior of `isspace` as far as I'm aware. It might break an old problem to change it, but it is unlikely.

evouga: Ok. So we need consensus on:

1. The character set supported by the default output validator. Options include:
- the 128 standard ASCII code points only;
- ASCII, plus some unspecified additional set of characters left up to the contest system (the status quo);
- UTF-8 Unicode;
- something else.

2. The characters recognized as whitespace. Options include:
- a specific set of ASCII whitespace characters only;
- an unspecified set of characters left up to the contest system (the status quo);
- Unicode whitespace characters;
- something else.

evouga: > I find it hard to believe that contestants would expect non-ASCII whitespace to be interchangable with ASCII whitespace.

Honestly, I don't think contestants think too much about character encodings at all.

I'm not going to die on this hill; but my point is that non-ASCII spaces *do* show up in places visible to contestants (including on web sites and in PDFs of the problem statement) and I do think they would expect a copy-pasted "space" to be treated as a space.

niemela: These are all the Unicode "whitespace" characters: https://en.wikipedia.org/wiki/Whitespace_character#Unicode

We currently support exactly the once in the "Basic Latin" block, i.e. Tab, Line feed, Vertical tab, Form feed, Carriage return, and space. (Support as in "they are considered whitespace, and thus used for separating tokens", you can **use** all the others perfectly fine, but they will not. be "ignored" as whitespace).

Looking at the other whitespace characters, I don't really feel that we need to "support" them. I would prefer to properly specify the exact current behavior and let that be the spec.

And this doesn't mean that we don't support UTF-8 or Unicode. We have problems that use much more than ASCII for input and output, and that works perfectly fine.

evouga: It "works fine" because the implementation uses functions whose behavior is unspecified outside the POSIX locale, where the unspecified behavior happens to have done the "right thing" for your specific contest. That's not a good specification :(

I happy to say that the default output validator must support ASCII, and may support any contest-system-specified superset of ASCII, since that seems to be the way the wind is blowing.

spec/2023-07-draft.md
evouga: I've reworded to make it clearer that this is the exceptional case.

evouga:
Now requires only ASCII support; gives contest systems permission to allow a superset of ASCII (with contest-system-specified behavior).

evouga:
Per a phone conversation with @niemela , I've made the specification less permissive about what the contest system can do.
- `case_sensitive` ignores the case of exactly the ASCII letters, and nothing else.  
- I've removed all mention of UTF-8 now; the validator compares tokens byte-by-byte. Valid UTF-8 should be parsed and compared correctly; invalid UTF-8 must be accepted by the validator and compared to the answer file as raw single-character byte sequences.

spec/2023-07-draft.md
```diff
@@ -972,24 +972,54 @@ and optionally [report additional feedback](#reporting-additional-feedback).
...
-In its default mode, it tokenizes the output and answer files and compares them token by token.
-It supports the following command-line arguments to control how tokens are compared.
+The default output validator supports the following command-line arguments:
 
 Arguments                    | Description
 ---------------------------- | -----------
-`case_sensitive`             | indicates that comparisons should be case-sensitive.
-`space_change_sensitive`     | indicates that changes in the amount of whitespace should be rejected (the default is that any sequence of 1 or more whitespace characters are equivalent).
+`case_sensitive`             | indicates that comparisons should be case-sensitive (see below for details).
+`space_change_sensitive`     | indicates that differences in type or amount of of whitespace should be rejected (see below for details).
```
niemela, line 980:
```suggestion
`space_change_sensitive`     | indicates that differences in type or amount of whitespace should be rejected (see below for details).
```

niemela (changes requested):

niemela (approved):

spec/2023-07-draft.md
```diff
@@ -972,24 +972,54 @@ and optionally [report additional feedback](#reporting-additional-feedback).
...
 
-When supplying both a relative and an absolute tolerance, the semantics are that a token is accepted if it is within either of the two tolerances.
-When a floating-point tolerance has been set, any valid formatting of floating-point numbers is accepted for floating-point tokens.
-So, for instance, if a token in the answer file says `0.0314`, a token of `3.14000000e-2` in the output file would be accepted.
+#### Output Parsing
+
+The default output validator parses the submission output and answer files as a string of single-byte characters and tokenizes the files by splitting on sequences of one or more consecutive whitespaces.
+Whitespace characters are: space (`0x20`), form feed (`0x0c`), line feed (`0x0a`), carriage return (`0x0d`), horizontal tab (`0x09`), and vertical tab (`0x0b`).
+In its default mode, the default output validator then ignores the whitespace and compares the submission output and answer files token by token. 
+(If there is a disagreement in the number of tokens in the two files, the validator rejects the submission instead.)
```
eldering, line 990: Nitpick: 
```suggestion
If there is a mismatch in the number of tokens in the two files, the validator rejects the submission.
```
I the parentheses should not be there because this is a relevant specification. Also small wording changes.

spec/2023-07-draft.md
```diff
@@ -972,24 +972,54 @@ and optionally [report additional feedback](#reporting-additional-feedback).
...
+The validator may also reject any submission output that is unreasonably large (including due to containing unreasonable amounts of whitespace).
+
+#### Comparing Floating-Point Tokens
+
+If a floating-point tolerance has been set, the default output validator will attempt to parse each answer file token as floating-point (see [general requirements](#general-requirements) for details).
```
eldering, line 995:
```suggestion
If a floating-point tolerance has been set, the default output validator will attempt to parse each answer file token as a floating-point number (see [general requirements](#general-requirements) for details). 
```
I'd say that this is more correct than using just "floating point"; it would require some more changes below.

spec/2023-07-draft.md
```diff
@@ -972,24 +972,54 @@ and optionally [report additional feedback](#reporting-additional-feedback).
...
 
 It is an error to provide any of the `float_relative_tolerance`, `float_absolute_tolerance`, or `float_tolerance` arguments more than once, or to provide a `float_tolerance` alongside `float_relative_tolerance` and/or `float_absolute_tolerance`.
 
-If no floating-point tolerance has been set, floating-point tokens are treated just like any other token and must match exactly.
+Note that if a floating-point tolerance has been set, the default output validator will parse exact integers as floating-point and apply the above semantics to them. 
+For problems containing a mix of integer and floating-point output, a custom output validator must be used if exact comparison of the integer tokens is required.
+
+#### Comparing String Tokens
+
+If `case_sensitive` is specified, the two tokens must match exactly, i.e. consist of exactly the same character sequence.
```
eldering, line 1012: The text doesn't say anything about character sets. Wouldn't it be better to say "byte sequence" here? I'm not an expert on UTF-8, but doesn't it have different byte sequences repesenting the same character (sequence)?

simonlindholm: No, UTF-8 has a unique encoding of every character. There is however the question of how to deal with bytes sequences that are not valid UTF-8, which would be resolved by switching to comparing raw bytes instead.

spec/2023-07-draft.md
evouga: Let's remove the parentheses but keep the "instead"? Otherwise niemela had a concern earlier that the sentence might be read to imply the inverse.

spec/2023-07-draft.md
evouga: Yes, let's say byte sequence.

UTF-8 has overlongs, which either need to be treated as invalid or aliased to their shortest code points. But let's not deal with this---if students submit UTF-8 output that is properly encoded, comparing byte sequences should work fine.

Issue #344, opened 2024-09-18
rethink constants


mzuenni:
Right now we replace constants at the following places:

1. problem statements
2. input and output validators
3. included code
4. example submissions
5. `testdata.yaml`

There are a few things that I want to mention.

1. "Replacement in example submissions and included code" What is the use case here and why do we differentiate between example submissions and other submissions? (Note that I don't want to argue that we should do replacement for all submissions, rather I think that we should not touch any code at all).

2. In most cases text replacement is just inconvenient. We already restrict validators to c++ and python in both cases there are nicer ways to automagically introduce constants. For example in c++ we could add `constexpr <std::string_view,long long,double> <constant name> = <constant value>` and something similar in python. In latex, we could also introduce the constants as macros.
I know that this makes constants more complicated to implement on the tooling side but IMO it is much more convenient for the user. (For markdown and `testdata.yaml` i have no better idea...)

RagnarGrootKoerkamp:
Text replacement badly messes up editor highlighting and autoformatting and such. Wouldn't it be nicer in your example if we did a `#define MAX_N 1000` in the background? And similarly we could `\newcommand{\maxn}{1000}` for latex.

evouga:
Is there concrete evidence that the proposed balanced-braces “badly” messes up anything in practice?

I think it is good to make it obvious which variables are externally defined problem constants and which are statically scoped, and to make it more or less impossible to accidentally shadow the problem constants

mzuenni:
regarding the change in judge submissions: I am not sure if this really makes things much easier since you still need to change all the test data anyway (note that we don't define generators). And distinguishing between user submissions and jury submissions makes stuff more complicated than necessary. Further, we don't don't allow a constant to write solutions like "do random stuff until you are out of time" for which automatic updates would be useful.

regarding messing stuff up: my problem is more that this messes up standard build processes. For c++ I can't compile stuff anymore but if it's done with macros or as constant I can simply tell my compiler what macros exist. Same for latex. The curly bracket thing breaks latexmk rebuild whereas with macros this can not happen.

Tagl:
Yeah I think defines and environment variables might be better approaches here.

evouga:
For LaTeX, don't you already need a script that does preprocessing before being able to compile the problem statement? You need to wrap the problem statement in a `document` environment, define all of the problem-package-specific macros, etc.

[...]

PR #370, opened 2025-03-01
add constants command for latex

mzuenni:
Adresses #344.
I am not sure if we want to keep constants for includes and submission.

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -392,13 +392,15 @@ where `name` is one of the names defined in `constants`.
 Tags {% raw %}`{{xyz}}`{% endraw %} containing a name that is not defined are not modified but may be warned for.
 
 All constant sequences in the following files will be replaced by the value of the corresponding constant:
+- Markdown problem statements
 - input and output validators
 - included code
 - example submissions
 - `testdata.yaml`
 
+Note that constants are also available in LaTeX problem statements via the dedicated command `\constant{name}`.
+
 Constant sequences are **not** replaced in test data files or in `problem.yaml` itself.
-Note that constants are also available in the problem statements via the dedicated command `\constant{name}`.
 
 ## Problem Statements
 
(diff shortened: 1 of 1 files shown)


spec/2023-07-draft.md
```diff
@@ -392,13 +392,13 @@ where `name` is one of the names defined in `constants`.
 Tags {% raw %}`{{xyz}}`{% endraw %} containing a name that is not defined are not modified but may be warned for.
 
 All constant sequences in the following files will be replaced by the value of the corresponding constant:
-- problem statements
```
niemela, line 395: We still do the replace for **Markdown** statements, the special casing is **only** for LaTeX problem statements.

spec/2023-07-draft.md
```diff
@@ -392,13 +392,13 @@ where `name` is one of the names defined in `constants`.
...
 - input and output validators
 - included code
 - example submissions
 - `testdata.yaml`
 
 Constant sequences are **not** replaced in test data files or in `problem.yaml` itself.
+Note that constants are also available in the problem statements via the dedicated command `\constant{name}`.
```
niemela, line 401: Do we want to do replacement (for simplicity and consistency) **and** dedicated commands for those that prefer that, or is that a terrible idea?

niemela (changes requested):

spec/2023-07-draft.md
mzuenni: No. As mentioned before, I think that replacement is not grate and a language specific injection would make fewer problems... Unfortunately, there is stuff like Markdown where there is no language specific option.
Anyway, constant replacement breaks a few things during interactive latex compilation and also `{{xxx}}` is perfectly fine latex code that also appears very commonly in macros.

spec/2023-07-draft.md
mzuenni: fixed

niemela (approved):

mzuenni:
can this be merged?

Issue #308, opened 2024-08-19
Description of `.interaction`-files are spread out


niemela:
Currently `.interaction`-files are described/specified (slightly differently) in
- section "Test Data" / "Interactive Problems"
- section "Test Data" / "Samples" / "Samples shown in the problem statement"
- section "Output Validator" / "Reporting Additional Feedback" / "Multi-pass validation"

We should consider defining it in one place and refer to that.

Issue #373, opened 2025-03-06
directory vs extension for samples


mzuenni:
While implementing the spec for the new sample directories (`data/sample/statement/` and `data/sample/download/`) for BAPCtools I noticed a few quirks and potential pitfalls.
For example the existence of `data/sample/download/` shadows all testcases (not only those for which a overwrite file is provided).
Additionally the directory structure makes it more likely the directories get out of sync, i.e. a sample is only added/removed from one of the samples.
Also somehow the point of `data/sample/statement/` is missed? IMO this is only useful to present non validated input files (else it could just be a normal testcase?) so i explicitly want it to be **not** validated? (see the attached PR)

I would propose to instead introduce the extensions `.statement` and `.download` for this.
For example `data/sample/x.in` would than be replaced by `data/sample/x.in.statement` for the problem statement.

This approach has a few benefits:
- all testcases related files are in the same directory
- files get shadowed on a per file basis not in generell 
- all files belonging to the testcase `data/sample/<x>` match the glob `data/sample/<x>.*`

RagnarGrootKoerkamp:
I'm neutral on this. But if playing around with it seems better, yeah I get that and sounds reasonable.

evouga:
>  For example the existence of data/sample/download/ shadows all testcases (not only those for which a overwrite file is provided).

This is intentional as otherwise there is no way to remove test cases (such as those in `/data/sample/statement`) that should not be included.

> IMO this is only useful to present non validated input files (else it could just be a normal testcase?) 

Often you want to show different sample output than what's in the `.ans` file (for example, because the `.ans` file contains way more precision than is needed by the problem statement; or the `.ans` file contains some certificate used for output validation that is unrelated to the output expected from the contestant submission). In this case you want to use the `samples` folder and you want the data to be validated (with the `.ans` and `.out` files treated as described in the spec).

mzuenni:
>>For example the existence of data/sample/download/ shadows all testcases (not only those for which a overwrite file is provided).
>
>This is intentional as otherwise there is no way to remove test cases (such as those in /data/sample/statement) that should not be included.

Having a `statement` test case without a `.in` would already signal this?

>>IMO this is only useful to present non validated input files (else it could just be a normal testcase?)
>
[...]

PR #375, opened 2025-03-06
rewrote statement sample section

mzuenni:
Adresses #373

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -697,7 +697,7 @@ Note that if you are using a `.interaction` file you must not provide a `.statem
 #### Samples Available for Download
 
 By default, the `.in` and `.ans` files in `data/sample` are available for download.
-Note that the content of `.in.statement` replaces that of `.in` and that the content of `.out` and `.ans.statement` replace that of `.ans` for the download.
+Note that the content of `.in.statement` replaces that of `.in` and that the content of `.out` or `.ans.statement` replace that of `.ans` for the download.
 This behavior can be further customized by providing files with the extension `in.download` or `ans.download`.
 If one of these files exists, its contents replaces that of the file with the same name -- except the `.download` extension -- for the problem download.
 
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -664,7 +664,7 @@ It may be missing (for problems with no samples) or empty.
 #### Samples Shown in the Problem Statement
 
 By default, the `.in` and `.ans` pairs from `data/sample` are shown in the problem statement.
-If a `.out` file exists the `.out` file is shown instead of the `ans` file in the problem statement.
+If a `.out` file exists the `.out` file is shown instead of the `.ans` file in the problem statement.
 This behavior can be customized by creating files with extension `.in.statement` and `.ans.statement`.
 If one of these files exists, its contents replaces that of the file with the same name -- except the `.statement` extension -- for purposes of the problem statement.
 Note that it is an error to provide a `.out` and a `.ans.statement` file.
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -680,7 +680,7 @@ For interactive problems displaying two files is typically not meaningfully to c
 Therefore, it is advised to instead provide samples for the problem statement in the form of a `.interaction` file.
 This file contains lines starting with `<` and `>`, containing an interaction log with output from the output validator starting with `<` and output from the submission starting with `>`.
 
-Note that if you are using a `.interaction` file you must not provide a `.statement` or `.out` file.
+Note that if you are using a `.interaction` file you must not provide a `.in.statement`, `.ans.statement`, or `.out` file.
 
 ##### Multi-Pass Problems
 
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -700,6 +700,7 @@ By default, the `.in` and `.ans` files in `data/sample` are available for downlo
 Note that the content of `.in.statement` replaces that of `.in` and that the content of `.out` or `.ans.statement` replace that of `.ans` for the download.
 This behavior can be further customized by providing files with the extension `in.download` or `ans.download`.
 If one of these files exists, its contents replaces that of the file with the same name -- except the `.download` extension -- for the problem download.
+Additionally, any other file or directory with the extension `.download` is also available for download (without the `.download` extension).
 
 If you want to make other files -- like testing tools -- available for download, you can use [attachments](#attachments).
 
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -664,9 +664,10 @@ It may be missing (for problems with no samples) or empty.
 #### Samples Shown in the Problem Statement
 
 By default, the `.in` and `.ans` pairs from `data/sample` are shown in the problem statement.
+If a `.out` file exists the `.out` file is shown instead of the `ans` file in the problem statement.
 This behavior can be customized by creating files with extension `.in.statement` and `.ans.statement`.
 If one of these files exists, its contents replaces that of the file with the same name -- except the `.statement` extension -- for purposes of the problem statement.
-Note that `.statement` files are not validated.
+Note that it is an error to provide a `.out` and a `.ans.statement` file.
 
 ##### Interactive Problems
 
@@ -675,44 +676,41 @@ The validator gets access to the `.in` and `.ans` files for each test case and c
 Standard in of the output validator corresponds to standard out of the submission, and standard out of the output validator corresponds to standard in of the submission.
 Therefore, the output validator can control what information from the `.in` and `.ans` files is provided to the submission.
 
-For interactive problems the `.in` and `.ans` pairs typically do not meaningfully capture how users are expected to interact with the output validator in such problems.
-Therefore, it is advised to provide samples for the problem statement in one of the following two forms:
-- An `.in.statement` and `.ans.statement` file as explained before.
-- An `.interaction` file that contains lines starting with `<` and `>`,
-  containing an interaction log with output from the output validator starting with `<` and output from the submission starting with `>`.
+For interactive problems displaying two files is typically not meaningfully to capture how users are expected to interact with the output validator.
+Therefore, it is advised to instead provide samples for the problem statement in the form of a `.interaction` file.
+This file contains lines starting with `<` and `>`, containing an interaction log with output from the output validator starting with `<` and output from the submission starting with `>`.
 
-Note that you can not combine `.statement` and `.interaction` files.
+Note that if you are using a `.interaction` file you must not provide a `.statement` or `.out` file.
 
 ##### Multi-Pass Problems
 
 Multi-pass problems require a custom output validator, which interacts with the submission see [multi-pass](#multi-pass-validation).
 
-For multi-pass problems the `.in` and `.ans` pairs typically do not meaningfully capture how users are expected to interact with the output validator in such problems.
-Therefore, it is advised to provide samples for the problem statement in one of the following two forms:
-- An `.in.statement` and `.ans.statement` file as explained before.
-- An `.interaction` file that contains lines starting with `<` and `>`, like for interactive problems.
-  Passes are separated by a line containing `---` (three dashes).
-  When the problem is not interactive, simply start each pass by a number of lines starting with `<`, containing the sample input, followed by some lines starting with `>`, containing the sample answer.
+For multi-pass problems displaying two files is typically not meaningfully to capture how users are expected to interact with the output validator.
+Therefore, it is advised to instead provide samples for the problem statement in the form of a `.interaction` file.
+This file contains lines starting with `<` and `>`, like for interactive problems.
+Passes are separated by a line containing `---` (three dashes).
+When the problem is not interactive, simply start each pass by a number of lines starting with `<`, containing the sample input, followed by some lines starting with `>`, containing the sample output.
 
-Note that you can not combine `.statement` and `.interaction` files.
+Note that if you are using a `.interaction` file you must not provide a `.statement` or `.out` file.
 
 #### Samples Available for Download
 
 By default, the `.in` and `.ans` files in `data/sample` are available for download.
-Note that, files with `.statement` replace those without the additional extension, e.g., `.in.statement` is renamed to `.in` and replaces it for the download.
-This behavior can be further customized by creating files with the additional extension `.download` like `.in.download`.
+Note that the content of `.in.statement` replaces that of `.in` and that the content of `.out` and `.ans.statement` replace that of `.ans` for the download.
+This behavior can be further customized by providing files with the extension `in.download` or `ans.download`.
 If one of these files exists, its contents replaces that of the file with the same name -- except the `.download` extension -- for the problem download.
-Note that `.download` files are not validated.
 
 If you want to make other files -- like testing tools -- available for download, you can use [attachments](#attachments).
 
 #### Validation
 
 All `data/sample/*.in` files are input-validated.
-For non interactive and non multi-pass problems all test cases in `data/sample` must be accepted by the output validator, with the `.ans` file used as both the `answer_file` and `team_output`.
-If the problem is neither interactive nor multi-pass and `.in.statement` does not exits then `.ans.statement` is assumed to be a valid submission output for `.in`.
-Thus it has to pass output validation `output_validator .in .ans < .ans.statement`.
+For non interactive and non multi-pass problems the `.out` files must pass the output validator.
 All other files are not validated in any way.
+
+Note that `.ans.statement` and `.out` can both be used to change what is shown in the statement.
+However, since only the `.out` files are validated it is advised to use these if possible.
 
 Validation can be customized by specifying `input_validator_args` and `output_validator_args` in `data/sample/testdata.yaml`.
 
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -707,6 +707,7 @@ If you want to make other files -- like testing tools -- available for download,
 
 All `data/sample/*.in` files are input-validated.
 For non interactive and non multi-pass problems the `.out` files must pass the output validator.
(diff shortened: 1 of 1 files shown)


mzuenni:
@niemela any feedback on this?

spec/2023-07-draft.md
```diff
@@ -676,50 +675,48 @@ The validator gets access to the `.in` and `.ans` files for each test case and c
...
 
 If you want to make testing tools available for download, you can use [attachments](#attachments).
 
-##### Other Problem Types
+##### Multi-Pass Problems
 
-For each test case, `data/sample/statement` must contain one of the following:
-- An `.in` and `.ans` file, which are both shown.
-- An `.in`, `.ans`, and `.out` file, of which the `.in` and `.out` file are shown, and the `.ans` file is used for validation (see below).
+Multi-pass problems require a custom output validator, which interacts with the submission see [multi-pass](#multi-pass-validation).
```
niemela, line 690: So does interactive, why is this only written here?

spec/2023-07-draft.md
```diff
@@ -676,50 +675,48 @@ The validator gets access to the `.in` and `.ans` files for each test case and c
...
 
+For multi-pass problems the `.in` and `.ans` pairs typically do not meaningfully capture how users are expected to interact with the output validator in such problems.
+Therefore, it is advised to provide samples for the problem statement in one of the following two forms:
+- An `.in.statement` and `.ans.statement` file as explained before.
+- An `.interaction` file that contains lines starting with `<` and `>`, like for interactive problems.
+  Passes are separated by a line containing `---` (three dashes).
+  When the problem is not interactive, simply start each pass by a number of lines starting with `<`, containing the sample input, followed by some lines starting with `>`, containing the sample answer.
```
niemela, line 697: I feel that the "Interactive Problems" and "Multi-Pass Problems" sections could definitely be merged, and probably also into the parent section?

spec/2023-07-draft.md
```diff
@@ -676,50 +675,48 @@ The validator gets access to the `.in` and `.ans` files for each test case and c
...
-If this directory does not exist, the contents of `data/sample/statement` are available for download,
-with any `.out` files renamed to `.ans` (replacing existing `.ans` files) and all `.interaction` files removed.
-When that also does not exist, the test cases in `data/sample` are available for download.
+Note that you should choose one of these two options and not both.
 
-The `data/sample/download` directory may contain anything, and none of its contents are validated. 
-We recommend placing test-case-specific files in this directory, and using [attachments](#attachments) for testing tools and other downloads not associated with any particular test case.
-Testing tools may warn when test cases that are listed in `data/sample` or `data/sample/statement` do not appear in `data/sample/download`,
-or when `data/sample/download` contains test cases that do not appear in one of the two other locations.
+If you want to make testing tools available for download, you can use [attachments](#attachments).
```
niemela, line 701: Shouldn't this be in the "Samples Available for Download" section?

spec/2023-07-draft.md
```diff
@@ -676,50 +675,48 @@ The validator gets access to the `.in` and `.ans` files for each test case and c
...
+  Passes are separated by a line containing `---` (three dashes).
+  When the problem is not interactive, simply start each pass by a number of lines starting with `<`, containing the sample input, followed by some lines starting with `>`, containing the sample answer.
 
-#### Samples Available for Download
-
-When `data/sample/download` exists, the files in there are available for download.
-If this directory does not exist, the contents of `data/sample/statement` are available for download,
-with any `.out` files renamed to `.ans` (replacing existing `.ans` files) and all `.interaction` files removed.
-When that also does not exist, the test cases in `data/sample` are available for download.
+Note that you should choose one of these two options and not both.
```
niemela, line 699: Same as above.

spec/2023-07-draft.md
```diff
@@ -676,50 +675,48 @@ The validator gets access to the `.in` and `.ans` files for each test case and c
...
-Testing tools may warn when test cases that are listed in `data/sample` or `data/sample/statement` do not appear in `data/sample/download`,
-or when `data/sample/download` contains test cases that do not appear in one of the two other locations.
+If you want to make testing tools available for download, you can use [attachments](#attachments).
 
-#### Validation
+#### Samples Available for Download
 
-All `data/sample/*.in` files are input-validated. For problems that are not interactive, all `data/sample/statement/*.in` are also input-validated.
+By default, the files in `data/sample` are available for download, except for `.interaction` files.
+Note that, files with `.statement` replace those without the additional extension, e.g., `.in.statement` is renamed to `.in` and replaces it for the download.
```
niemela, line 706: Do we allow any other `.statement` files than `.in.statement` and `.and.statement`? I would say that the "Samples Shown in the Problem Statement" section above tells me "no", and this line implies "maybe yes?".

spec/2023-07-draft.md
```diff
@@ -676,50 +675,48 @@ The validator gets access to the `.in` and `.ans` files for each test case and c
...
-We recommend placing test-case-specific files in this directory, and using [attachments](#attachments) for testing tools and other downloads not associated with any particular test case.
-Testing tools may warn when test cases that are listed in `data/sample` or `data/sample/statement` do not appear in `data/sample/download`,
-or when `data/sample/download` contains test cases that do not appear in one of the two other locations.
+If you want to make testing tools available for download, you can use [attachments](#attachments).
 
-#### Validation
+#### Samples Available for Download
 
-All `data/sample/*.in` files are input-validated. For problems that are not interactive, all `data/sample/statement/*.in` are also input-validated.
+By default, the files in `data/sample` are available for download, except for `.interaction` files.
```
niemela, line 705: Not all. Definitely not `.png`, `.jpg`, `.jpeg`, `.svg`, I'm pretty sure not `.yaml` either.

spec/2023-07-draft.md
```diff
@@ -676,50 +675,48 @@ The validator gets access to the `.in` and `.ans` files for each test case and c
...
+Note that, files with `.statement` replace those without the additional extension, e.g., `.in.statement` is renamed to `.in` and replaces it for the download.
+This behavior can be further customized by creating files with the additional extension `.download` like `.in.download`.
```
niemela, line 707: What `tc.*.download` files are allowed? Previously we said "any", but now we don't, and I would **not** interpret that as "any".

I would prefer if we could limit to only `.in` and `.ans`, but maybe that's not enough?

spec/2023-07-draft.md
```diff
@@ -676,50 +675,48 @@ The validator gets access to the `.in` and `.ans` files for each test case and c
...
-When `data/sample/statement/` does exists, and the problem is not interactive, all test cases in `data/sample/statement/` must be accepted by the output validator. 
-The `.ans` is used as the `answer_file`. 
-The `.out` file is used as the `team_output`, if it exists. 
-Otherwise the `.ans` is used as the `team_output`. 
+#### Validation
 
-Files in `data/sample/download` are never validated, 
-although tooling may warn when there are inconsistencies between `download`, `statement`, and `data/sample/` itself.
+All `data/sample/*.in` files are input-validated.
+For non interavtive and non multi-pass problems all test cases in `data/sample` must be accepted by the output validator, with the `.ans` file used as both the `answer_file` and `team_output`.
```
niemela, line 714:
```suggestion
For non interactive and non multi-pass problems all test cases in `data/sample` must be accepted by the output validator, with the `.ans` file used as both the `answer_file` and `team_output`.
```

spec/2023-07-draft.md
```diff
@@ -676,50 +675,48 @@ The validator gets access to the `.in` and `.ans` files for each test case and c
...
+If the preoblem is neither interactive nor multipass and `.in.statement` does not exits then `.ans.statement` is assumed to be a valid submission output for `.in`.
```
niemela, line 715:
```suggestion
If the problem is neither interactive nor multi-pass and `.in.statement` does not exits then `.ans.statement` is assumed to be a valid submission output for `.in`.
```

niemela: I think it would be better to day that `out` is always assumed to be valid submission output for `.in`. I.e. regardless of whether `.in.statement` exists. Otherwise there is no way to have a validated output when you want a statement input.

Then `.ans.statement` could be never validated. (And system can warn when you use this)

spec/2023-07-draft.md
```diff
@@ -676,50 +675,48 @@ The validator gets access to the `.in` and `.ans` files for each test case and c
...
+Thus it has to pass output validation `output_validator .in .ans < .ans.statement`.
+All other files are not validated in any way.
```
niemela, line 717: I mean.... true... but we only ever validate `.in`, `.and`, `.out`, so it feels strange to say it.

niemela (changes requested):

spec/2023-07-draft.md
```diff
@@ -665,9 +664,9 @@ It may be missing (for problems with no samples) or empty.
 #### Samples Shown in the Problem Statement
 
 By default, the `.in` and `.ans` pairs from `data/sample` are shown in the problem statement.
-This behavior can be customized by creating the `data/sample/statement` directory.
-If this directory exists, its contents replaces that of `data/sample` for purposes of the problem statement.
-Interactive problems with sample test cases should almost always make use of this directory, since the `.in` and `.ans` pairs do not meaningfully capture how users are expected to interact with the output validator in such problems.
+This behavior can be customized by creating files with extension `.in.statement` and `.ans.statement`.
+If one of these files exists, its contents replaces that of the file with the same name -- except the `.statement` extension -- for purposes of the problem statement.
+Note that `.statement` files are not validated.
```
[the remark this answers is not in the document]
mzuenni: right now `.ans.statement` are validated so you can see them as `.out`

spec/2023-07-draft.md
mzuenni: it is already written for interactive, i just added multi-pass which was missing

spec/2023-07-draft.md
mzuenni: I see no use case for this, but I also don't care if they exists?

spec/2023-07-draft.md
mzuenni: updated this

spec/2023-07-draft.md
mzuenni: maybe but i am unsure how ^^'

niemela: I certainly does not have to be in this PR, in fact I think it make more sense to try to do that in a follow up. There is enough happening here...

spec/2023-07-draft.md
```diff
@@ -676,50 +676,43 @@ The validator gets access to the `.in` and `.ans` files for each test case and c
...
-For each test case, `data/sample/statement` must contain one of the following:
-- An `.in` and `.ans` file, which are both shown.
-- An `.in`, `.ans`, and `.out` file, of which the `.in` and `.out` file are shown, and the `.ans` file is used for validation (see below).
+For multi-pass problems displaying two files is typically not meaningfully to capture how users are expected to interact with the output validator.
+Therefore, it is advised to instead provide samples for the problem statement in the form of a `.interaction` file.
+This file contains lines starting with `<` and `>`, like for interactive problems.
+Passes are separated by a line containing `---` (three dashes).
+When the problem is not interactive, simply start each pass by a number of lines starting with `<`, containing the sample input, followed by some lines starting with `>`, containing the sample output.
 
+Note that if you are using a `.interaction` file you must not provide a `.statement` or `.out` file.
```
niemela, line 695:
```suggestion
Note that if you are using a `.interaction` file you must not provide a `.in.statement`, `.ans.statement`, or `.out` file.
```

spec/2023-07-draft.md
niemela: Thinking a bit more about this. If we have problems with command line of file input, those would be given in `.yaml` and `.files` respectively, so they could need to be overridden for the statement.

spec/2023-07-draft.md
```diff
@@ -676,50 +676,43 @@ The validator gets access to the `.in` and `.ans` files for each test case and c
...
 
 #### Samples Available for Download
 
-When `data/sample/download` exists, the files in there are available for download.
-If this directory does not exist, the contents of `data/sample/statement` are available for download,
-with any `.out` files renamed to `.ans` (replacing existing `.ans` files) and all `.interaction` files removed.
-When that also does not exist, the test cases in `data/sample` are available for download.
+By default, the `.in` and `.ans` files in `data/sample` are available for download.
```
niemela, line 699:
```suggestion
By default, the `.in`, `.ans`, and `.files` files in `data/sample` are available for download.
```

Should we special case empty `.in` files? I.e. if there is no input, you currently must have an empty `.in` file (as opposed to a missing `.in` file, and I don't think we should change that). Should the empty `.in` file still be downloadable?

spec/2023-07-draft.md
```diff
@@ -676,50 +676,43 @@ The validator gets access to the `.in` and `.ans` files for each test case and c
...
+Note that the content of `.in.statement` replaces that of `.in` and that the content of `.out` and `.ans.statement` replace that of `.ans` for the download.
```
niemela, line 700:
```suggestion
Note that the content of `.in.statement` replaces that of `.in` and that the content of `.out` or `.ans.statement` replace that of `.ans` for the download.
```

spec/2023-07-draft.md
```diff
@@ -676,50 +676,43 @@ The validator gets access to the `.in` and `.ans` files for each test case and c
...
+This behavior can be further customized by providing files with the extension `in.download` or `ans.download`.
+If one of these files exists, its contents replaces that of the file with the same name -- except the `.download` extension -- for the problem download.
```
niemela, line 702: Should we mention `.files.download`? I think "yes".

spec/2023-07-draft.md
niemela: Sorry, it should probably be "...and `.files`", right?

spec/2023-07-draft.md
```diff
@@ -676,50 +676,43 @@ The validator gets access to the `.in` and `.ans` files for each test case and c
...
-We recommend placing test-case-specific files in this directory, and using [attachments](#attachments) for testing tools and other downloads not associated with any particular test case.
-Testing tools may warn when test cases that are listed in `data/sample` or `data/sample/statement` do not appear in `data/sample/download`,
-or when `data/sample/download` contains test cases that do not appear in one of the two other locations.
+If you want to make other files -- like testing tools -- available for download, you can use [attachments](#attachments).
 
 #### Validation
 
-All `data/sample/*.in` files are input-validated. For problems that are not interactive, all `data/sample/statement/*.in` are also input-validated.
+All `data/sample/*.in` files are input-validated.
+For non interactive and non multi-pass problems the `.out` files must pass the output validator.
```
niemela, line 709: Is it clear what that means?

spec/2023-07-draft.md
```diff
@@ -676,50 +676,43 @@ The validator gets access to the `.in` and `.ans` files for each test case and c
...
+All other files are not validated in any way.
```
niemela, line 710: If there is no `.out` file (or `.interaction` or `.ans.statement`, so if the `.ans` is not overridden for the statement), then the `.ans` file must pass the output validator.

niemela (changes requested):

spec/2023-07-draft.md
```diff
@@ -665,9 +664,10 @@ It may be missing (for problems with no samples) or empty.
 #### Samples Shown in the Problem Statement
 
 By default, the `.in` and `.ans` pairs from `data/sample` are shown in the problem statement.
-This behavior can be customized by creating the `data/sample/statement` directory.
-If this directory exists, its contents replaces that of `data/sample` for purposes of the problem statement.
-Interactive problems with sample test cases should almost always make use of this directory, since the `.in` and `.ans` pairs do not meaningfully capture how users are expected to interact with the output validator in such problems.
+If a `.out` file exists the `.out` file is shown instead of the `ans` file in the problem statement.
```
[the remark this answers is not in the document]
mzuenni: i am not sure ^^'

spec/2023-07-draft.md
mzuenni: I would say yes.

spec/2023-07-draft.md
mzuenni: both would be fine for me

niemela:
This PR should resolve #263.

The thing still remaining is allowing (and special handling) `...` in `.interaction` (and maybe in `.ans.statement`). See also [my comment](https://github.com/Kattis/problem-package-format/issues/263#issuecomment-2710773917) in that thread.

niemela:
@mzuenni Maybe this PR could also fix #308? Feel free to say no.

mzuenni:
isn't #308 already fixed with this? I only see `.interaction` in "Test Data" / "Interactive Problems" and "Test Data" / "Multi-Pass Problems"

niemela:
> isn't #308 already fixed with this?

Yes, you're right.

niemela:
@mzuenni This needs to be rebased, then I think we're ready to merege. (And if there is any still open issues we can always break them out into new issues).

spec/2023-07-draft.md
niemela: @mzuenni How do you feel about adding mu suggestion here?

The other question could be broken out into a new issue.

niemela (approved):

mzuenni:
sorry I lost track somewhere, can you repeat the suggestion or link it ^^'

niemela:
> sorry I lost track somewhere, can you repeat the suggestion or link it ^^'

Ah, sorry, it showed up clearly in my GitHub view. [This change](https://github.com/Kattis/problem-package-format/pull/375#discussion_r1987077157).

spec/2023-07-draft.md
mzuenni: should this only be for empty `.in` files or do we just want to say that empty files are not included for downloads?

niemela: It's only needed for empty `.in` files or empty `.*.statement` and `.*.download` files. Any other file could be either not there, or overridden by an empty `.*.statement` or `.*.download`. We could even limit it so empty `.*.statement` and `.*.download` files, but then you would have to actually override empty `.in` files with empty `.in.statement` or `.in.download` files.

Issue #329, opened 2024-08-20
Static validator needs some clarification


niemela:
Comments from @gkreitz:
> * There is no specification how the validator renders a judgment (I assume exit 42/43?).
> * There is no way I see for a static validator to indicate what languages it supports (I guess you can sort of do something default for non-supported languages, but.. yuck...).

Thoughts?

niemela:
> There is no way I see for a static validator to indicate what languages it supports (I guess you can sort of do something default for non-supported languages, but.. yuck...).

IMO, the static validator *must* handle all languages allowed by `languages` in `problem.yaml`. Maybe we should say that explicitly?

Note that "handle| *could* mean, "if C++ just accept".

Tagl:
> There is no specification how the validator renders a judgment (I assume exit 42/43?).

This is indeed missing. It should report in the same way as the output validator, with exit codes 42 and 43, using judgemessage.txt, teammessage.txt and score.txt.

Tagl:
I'll try to provide one for an old problem we had at RU as soon as I can.
This is all from educational view, the programming language was Python in all cases and the teachers wanted to fail for using certain things in most cases, such as preventing use of while, for, recursion, or certain builtin functions.
In Python you can kind of do this already with `include`.
Almost always your static validator would be focused on a specific language, not multiple.
Having the compiled binaries and artifacts I'm not sure is useful or necessary (it might be), but having the source code is.

gkreitz:
> I'll try to provide one for an old problem we had at RU as soon as I can.

Thank you!

> Having the compiled binaries and artifacts I'm not sure is useful or necessary (it might be), but having the source code is.

I didn't phrase myself very well. My point was basically that if I wrote a static analyzer, I would _much_ prefer to have its input only be exactly the files submitted by the user, instead of those plus whatever compilation artifacts may have been produced by the compilation process used by a judging system (which an average problem author has little insight into, and which may easily change).

Tagl:
A few things I found from Fall 2023, these are 4 different examples:

Teacher wanted students to specifically use a for loop to find two digit numbers less than $x$ satisfying a specific property.
[...]

PR #377, opened 2025-03-07
Clean up the details of static validators

evouga:
Addresses https://github.com/Kattis/problem-package-format/issues/329

I've also cleaned up the scoring semantics of the static validator to align with the output validator: the default score is computed via maximum score inference from the parent test group, and this score can be overridden using `score.txt` or `score_multiplier.txt` as normal.

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -1240,15 +1240,12 @@ A static validator may be provided under the `static_validator` directory, simil
 Each test group may define a static validation test case.
 It is an error to define static validation test cases without providing a static validator.
 A static validation test case is defined within a group's `testdata.yaml` file by specifying the key `static_validation`.
-If a map is specified, it may have two keys `args`, and in the case of scoring test groups, `score`.
-The key `args` maps to a string which represents the additional arguments passed to the static validator in this group's static validation test case.
-The key `score` maps to a float which represents both the maximum score achievable for the static validation test case and the default score assigned in case the static validator accepts the submission for that test case.
-The static validator can override this score by outputting a value to `score.txt` in the feedback directory, the same as an output validator.
-It can have the value of `false` meaning there is no static validation, `true` meaning that static validation is enabled with no score defined and no arguments.
-Aggregation is then applied to the test case in the same manner as other test cases.
-It is an error to assign a score in a `pass-fail` test group.
-It is an error to not assign a score in a `min` or `sum` test group.
-It is also an error to provide a static validator for `submit-answer` type problems.
+If a map is specified, its only allowed key is `args`, which maps to a string which represents the additional arguments passed to the static validator in this group's static validation test case.
+The `static_validation` key can also have the value of `false` meaning there is no static validation, or `true` meaning that static validation is enabled with no additional arguments.
+
+For scoring problems, a test group's static validation test case, if it exists, is treated as a normal test case for the purposes of maximum score inference; see [Scoring Problems](#scoring-problems) for details. 
+
+It is an error to provide a static validator for `submit-answer` type problems.
 
 ### Invocation
 
@@ -1262,7 +1259,7 @@ The validator should be possible to use as follows on the command line:
 The meaning of the parameters listed above are:
 
 - language:
-  a string specifying the code of the language of the submission as shown in the [languages table](languages.md).
+  a string specifying the code of the language of the submission as shown in the [languages table](languages.md). A static validator must handle all of the programming languages specified in the `languages` key of `problem.yaml`.
 
 - entry_point:
   a string specifying the entry point, that is a filename, class name, or some other identifier, which the static validator should know how to use depending on the language of the submission.
@@ -1274,3 +1271,5 @@ The meaning of the parameters listed above are:
 
 - additional_arguments:
   in case the static validation test case specifies additional args, these are passed as additional arguments to the validator on the command line.
+
+The static validator follows the semantics of an output validator for [reporting a judgment](#reporting-a-judgement) and [overriding the validation test case's default score](#scoring-test-cases).


spec/2023-07-draft.md
```diff
@@ -1240,15 +1240,12 @@ A static validator may be provided under the `static_validator` directory, simil
...
-It can have the value of `false` meaning there is no static validation, `true` meaning that static validation is enabled with no score defined and no arguments.
-Aggregation is then applied to the test case in the same manner as other test cases.
-It is an error to assign a score in a `pass-fail` test group.
-It is an error to not assign a score in a `min` or `sum` test group.
-It is also an error to provide a static validator for `submit-answer` type problems.
+If a map is specified, its only allowed key is `args`, which maps to a string which represents the additional arguments passed to the static validator in this group's static validation test case.
+The `static_validation` key can also have the value of `false` meaning there is no static validation, or `true` meaning that static validation is enabled with no additional arguments.
+
+For scoring problems, a test group's static validation test case, if it exists, is treated as a normal test case for the purposes of maximum score inference; see [Scoring Problems](#scoring-problems) for details.
```
niemela, line 1246: Why did we drop the `score` key? Now, there is no way to set the (maximum) score for the static validation, and you typically want to do that. Note that, with regards to scoring, static validation works more like a group than a case, and groups can set their score in their `testcase.yaml`. The `score` key here is the equivalent to that.

spec/2023-07-draft.md
```diff
@@ -1262,7 +1259,7 @@ The validator should be possible to use as follows on the command line:
 The meaning of the parameters listed above are:
 
 - language:
-  a string specifying the code of the language of the submission as shown in the [languages table](languages.md).
+  a string specifying the code of the language of the submission as shown in the [languages table](languages.md). A static validator must handle all of the programming languages specified in the `languages` key of `problem.yaml`.
```
niemela, line 1262:
```suggestion
  a string specifying the code of the language of the submission as shown in the [languages table](languages.md). 
  A static validator must handle all of the programming languages specified in the `languages` key of `problem.yaml`.
```

Is it clear that if `languages: all` then it must handle all? I think so?

niemela (changes requested):

spec/2023-07-draft.md
evouga: That is possible, but a bit subtle.

If a test group has a maximum score of 100 points, 5 test cases, and a static validator with a `score` of 20 points, how much is each test case worth? Pretty clearly 16 points, right? That can be accounted for by changing the score inference formulas to include static validator maximum scores.

If a test group has a maximum score of 100 points, 5 test cases, and a static validator without a `score` key, how much is each test case worth? I can't see how the answer is anything but 20, but that can't be the intended behavior either.

Should static validators be required to include a `score` key, except within `unbounded` groups?

niemela: > If a test group has a maximum score of 100 points, 5 test cases, and a static validator with a `score` of 20 points, how much is each test case worth? Pretty clearly 16 points, right? 

Yes.

> That can be accounted for by changing the score inference formulas to include static validator maximum scores.

Exactly.

> If a test group has a maximum score of 100 points, 5 test cases, and a static validator without a `score` key, how much is each test case worth? I can't see how the answer is anything but 20, but that can't be the intended behavior either.

No, it should clearly be 16.6.... The static validator should be treated as a  "node" just like subgroups or test cases.

> Should static validators be required to include a `score` key, except within `unbounded` groups?

I don't think it needs to be required (and thus I don't think it *should* be), but I think you would almost always want it in practice.

niemela:
The "Static Validators" section really should not be after "Verdict/Score Aggregation". The logical places would be before "Input Validators" or after "Output Visualizer". I think I prefer the former.

@evouga Does it makes sense to make that change in this PR (I think so)?

spec/2023-07-draft.md
evouga: @niemela see new version below.

spec/2023-07-draft.md
```diff
@@ -74,8 +74,8 @@ File or Folder         | Required? | Described In
 `input_validators/`    | Yes       | [Input Validators](#input-validators)         | Programs that verifies correctness of the test data inputs
 `input_visualizer/`    | No        | [Input Visualizer](#input-visualizer)         | Scripts and documentation about how test case illustrations were generated
 `output_validator/`    | No        | [Output Validator](#output-validator)         | Custom program for judging solutions
-`output_visualizer/`   | No        | [Output Visualizer](#output-visualizer)       | Program to generate images illustrating submission output
 `static_validator/`    | No        | [Static Validator](#static-validator)         | Custom program for judging solutions with source files as input
+`output_visualizer/`   | No        | [Output Visualizer](#output-visualizer)       | Program to generate images illustrating submission output
```
niemela, line 78: The order here seems strange:

```
input_validators
input_visualizer
output_validator
static_validator
output_visualizer
```

I would suggest:
```
static_validator
input_validators
input_visualizer
output_validator
output_visualizer
```

or possibly:
```
input_validators
input_visualizer
output_validator
output_visualizer
static_validator
```

(But I think the former is better)

spec/2023-07-draft.md
```diff
@@ -1133,6 +1133,55 @@ It must not write to `score.txt`, `teammessage.txt`, or any other files in the f
...
+### Static Validation Test Cases
+
+Each test group may define a static validation test case.
+It is an error to define static validation test cases without providing a static validator.
+A static validation test case is defined within a group's `testdata.yaml` file by specifying the key `static_validation`.
+If a map is specified, its allowed key are:
+- `args`, which maps to a string which represents the additional arguments passed to the static validator in this group's static validation test case;
+- `score`, the maximum score of the static validation test case (see [Scoring Problems](#scoring-problems) for details).
+
+The `static_validation` key can also have the value of `false` meaning there is no static validation, or `true` meaning that static validation is enabled with no additional arguments and default maximum score.
```
niemela, line 1154: I think we should clarify that the default here is as specified in the "Maximum Score Inference" section.

niemela (changes requested):

spec/2023-07-draft.md
evouga: The static validator is much closer in function to the output validator (which judges contestant submissions) than to the input validator (which checks the judge data only).

evouga: I can see an argument for putting the static validator right before the output validator, but I prefer the current ordering as it lets us say that the static validator shares the judgement semantics of the output validator.

Tagl (approved):

spec/2023-07-draft.md
thorehusfeldt: Here’s a categorisation by problem life cycle:

Output validation (statically or not) belongs to submission-time. It is performed when a solver submission is validated.

Input validation belongs to installation-time of a problem. It is performed when a problem is installed on a judge system.

Visualisation belongs to problem’s construction-time. It is performed when the problem is designed. (Like input _generation_.)

niemela:
@evouga I think this is ready to merge, but it needs to be rebased.

Issue #372, opened 2025-03-04
Specifying `problem.yaml`, small changes suggested


thorehusfeldt:
Here’s a CUE schema for `problem.yaml`:

```cue
package problemformat

import "list"
import "time"

#ProgrammingLanguage: "ada" | "algol68" | "apl" | "bash" | "c" | "cgmp" | "cobol" | "cpp" | "cppgmp" | "crystal" | "csharp" | "d" | "dart" | "elixir" | "erlang" | "forth" | "fortran" | "fsharp" | "gerbil" | "go" | "haskell" | "java" | "javaalgs4" | "javascript" | "julia" | "kotlin" | "lisp" | "lua" | "modula2" | "nim" | "objectivec" | "ocaml" | "octave" | "odin" | "pascal" | "perl" | "php" | "prolog" | "python2" | "python3" | "python3numpy" | "racket" | "ruby" | "rust" | "scala" | "simula" | "smalltalk" | "snobol" | "swift" | "typescript" | "visualbasic" | "zig"
#LanguageCode: =~ "^[a-z]{2,3}(-[A-Z]{2})?$"
#Type: "pass-fail" | "scoring" | "multi-pass" | "interactive" | "submit-answer"

#Source: string | {
	name!: string
	url?:  string
}

#Person: string
#Persons: #Person | [#Person, ...#Person]

#Problem: {
	problem_format_version!: =~"^[0-9]{4}-[0-9]{2}(-draft)?$" | "draft" | "legacy" | "legacy-icpc"

	type?: *"pass-fail" | #Type | [#Type, ...#Type]
	if (type & [...]) != _|_ {
		_policy: true &
			!(list.Contains(type, "scoring") && list.Contains(type, "pass-fail")) &&
			!(list.Contains(type, "multi-pass") && list.Contains(type, "submit-answer")) &&
			!(list.Contains(type, "interactive") && list.Contains(type, "submit-answer"))
	}

	name!: string | {[#LanguageCode]: string}

	uuid!:    string

	version?: string

	credits?: #Person | {
		authors!:                                                        #Persons // TODO: Made authors mandatory
		["contributors" | "testers" | "packagers" | "acknowledgements"]: #Persons
		translators?: [#LanguageCode]: #Persons
	}

	source?: #Source | [#Source, ...#Source]
	license: *"unknown" | "public domain" | "cc0" | "cc by" | "cc by-sa" | "educational" | "permission"
	rights_owner?: _
[...]

PR #382, opened 2025-03-11
Clarify some types for data in problem.yaml

mpsijm:
Fixes C1, C2, C3, and C5 from #372.

Please check whether I didn't miss any "non-empty" anywhere :smile: Note that I left out "non-empty" for `keywords`, because I don't think it's an error to provide zero keywords.

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -144,7 +144,7 @@ Any unknown keys should be treated as an error.
 Key                                               | Type                                          | Required  | Default
 ------------------------------------------------- | --------------------------------------------- | --------- | -------
 [problem_format_version](#problem-format-version) | String                                        | Yes       |
-[type](#type)                                     | String or sequence of strings                 | No        | `pass-fail`
+[type](#type)                                     | String or non-empty sequence of strings       | No        | `pass-fail`
 [name](#name)                                     | String or map of strings                      | Yes       |
 [uuid](#uuid-and-version)                         | String                                        | Yes       |
 [version](#uuid-and-version)                      | String                                        | No        |
@@ -155,7 +155,7 @@ Key                                               | Type                        
 [embargo-until](#problem-publication-embargo)     | Date                                          | No        |
 [limits](#limits)                                 | Map with keys as defined below                | No        | See below
 [keywords](#keywords)                             | Sequence of strings                           | No        |
-[languages](#languages)                           | String or sequence of strings                 | No        | `all`
+[languages](#languages)                           | String or non-empty sequence of strings       | No        | `all`
 [allow_file_writing](#allow-file-writing)         | Boolean                                       | No        | `false`
 [constants](#constants)                           | Map of strings to int, float, or string       | No        |
 
@@ -172,7 +172,7 @@ Documentation for version `<version>` is available at `https://www.kattis.com/pr
 ### Type
 
 Type of problem.
-Must be either a single string or a sequence of strings, from the table below, with no repetition.
+Must be either a single string or a non-empty sequence of strings, from the table below, with no repetition.
 Two values listed as incompatible must not both be in the sequence.
 
 Value           | Incompatible with           | Comments
@@ -187,7 +187,7 @@ Value           | Incompatible with           | Comments
 
 The name of the problem in each language for which a problem statement exists.
 If there are statements in more than one language, the `name` field must be a map with the language codes as keys and the problem names as values.
-The set of languages for which `name` is given must **exactly** match the set of languages for which a problem statement exists.
+The set of languages for which `name` is given must **exactly** match the set of languages for which a [problem statement](#problem-statements) exists.
 
 A deliberately complex and construed example:
 ```yaml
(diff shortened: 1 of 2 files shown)


spec/2023-07-draft.md
```diff
@@ -316,28 +317,28 @@ Values other than `unknown` or `public domain` require `rights_owner` to have a
...
+`time_resolution`    | optional float > 0, in seconds | 1.0            |
+`memory`             | optional int > 0, in MiB       | system default | 2048
+`output`             | optional int > 0, in MiB       | system default | 8
+`code`               | optional int > 0, in KiB       | system default | 128
+`compilation_time`   | optional int > 0, in seconds   | system default | 60
+`compilation_memory` | optional int > 0, in MiB       | system default | 2048
+`validation_time`    | optional int > 0, in seconds   | system default | 60
+`validation_memory`  | optional int > 0, in MiB       | system default | 2048
+`validation_output`  | optional int > 0, in MiB       | system default | 8
+`validation_passes`  | optional int >= 1              | 2              |
```
niemela, line 341: I don't think the inequalities are needed here, but I guess it doesn't hurt?

Why `int > 0` vs `int >= 1`? Those mean the same.

spec/2023-07-draft.md
```diff
@@ -347,10 +348,10 @@ Only specify limits when the problem needs a specific limit, but do specify limi
...
 `time_multipliers` is a map with the following keys:
 
-Key                 | Comments | Default
-------------------- | -------- | -------
-`ac_to_time_limit`  | float    | 2.0
-`time_limit_to_tle` | float    | 1.5
+Key                 | Comments   | Default
+------------------- | ---------- | -------
+`ac_to_time_limit`  | float >= 1 | 2.0
+`time_limit_to_tle` | float >= 1 | 1.5
```
niemela, line 354: Should this be `>= 1`? We have problems in the legacy format where the multiplier is < 1.

spec/2023-07-draft.md
thorehusfeldt: If you have a preference, put your foot down and we’ll just make it so. The argument from parsimony is that `int & >0` is one letter less letter than `int & >=1`; the argument from cognitive load is that _my_ brain reads `int & >0` as the qualitative constraint “strictly positive integer” and `int & >=1` as the quantitative constraint “integer that is at least 1”. Others will have a different mental model of numerical types.

mpsijm: > Why `int > 0` vs `int >= 1`? Those mean the same.

For me, it was basically accidental :joy: I think my initial thought was that all the other limits are "strictly positive numbers, probably on the larger side", whereas `validation_passes` is "probably some small number, but at least 1". And I literally translated those thoughts :stuck_out_tongue: 

However, we should probably make `validation_passes` at least 2, the same as Thore also has in his schema. I've also added that this can only be given for multi-pass problems :slightly_smiling_face:

spec/2023-07-draft.md
mpsijm: If we would allow this, can you explain what that means? Does that mean that submissions in `submissions/time_limit_exceeded/` are allowed to be faster than the `time_limit`? :thinking: 

Also, that is the legacy format, so we can choose to be stricter about this in this new format.

spec/2023-07-draft.md
niemela: > If you have a preference, put your foot down and we’ll just make it so. The argument from parsimony is that `int & >0` is one letter less letter than `int & >=1`; the argument from cognitive load is that _my_ brain reads `int & >0` as the qualitative constraint “strictly positive integer” and `int & >=1` as the quantitative constraint “integer that is at least 1”. Others will have a different mental model of numerical types.

I don't have a strong opinion, the point was that *both* are used, which, since they mean the same thing, is confusing. I think I would prefer `> 0`.

> For me, it was basically accidental 😂 I think my initial thought was that all the other limits are "strictly positive numbers, probably on the larger side", whereas `validation_passes` is "probably some small number, but at least 1". And I literally translated those thoughts 😛

Yeah, I was initially thinking the same. Also, all the other are amounts, although technically integers, it's natural to think about them as `> 0`, while `validation_passes` is a count, so more naturally `>= 1`. But, again, saying the exact same thing differently is confusing.

> However, we should probably make `validation_passes` at least 2, the same as Thore also has in his schema. I've also added that this can only be given for multi-pass problems 🙂

Agreed, and that would also solve (or avoid?) the "problem".

mpsijm: Thinking of these other limits as "amounts" and `validation_passes` as "count" definitely makes sense! Then I'll consider this solved (or rather, "avoided" :smile:)

spec/2023-07-draft.md
mpsijm: > Having `ac_to_time_limit` < 1 would mean that the time limit is lower than the fastest submissions/accepted/ solution.

I think you mean "the time limit is lower than the _slowest_ accepted submission", but yes, that interpretation makes sense. And I agree that we don't need to keep this :stuck_out_tongue: We now allow [example submissions](https://icpc.io/problem-package-format/spec/2023-07-draft.html#example-submissions) to be in some custom directory that allows them to be both `AC` and `TLE` (we typically call it `mixed/`), so I think that should be used now rather than having a `ac_to_time_limit` multiplier < 1 :slightly_smiling_face:

niemela: > I think you mean "the time limit is lower than the _slowest_ accepted submission"

Indeed. Apparently I can't communicate today.

> And I agree that we don't need to keep this 😛 We now allow [example submissions](https://icpc.io/problem-package-format/spec/2023-07-draft.html#example-submissions) to be in some custom directory that allows them to be both `AC` and `TLE` (we typically call it `mixed/`), so I think that should be used now rather than having an `ac_to_time_limit` multiplier < 1 🙂

Yes, but that will not work. You are assuming that there are *other* `AC` that *are* fast enough, that is typically not (never?) the case when this was used. If the problem was that only one or some of many `AC` where a bit too slow, you would probably just drop them, this is for when *all* (or the only one) is a bit too slow.

This is also somewhat questionable, for sure, but it's not solved by your fix above.

mpsijm: That was my assumption indeed, that sounds very questionable :stuck_out_tongue: I can't imagine in which scenario the time limit should be set lower than _all_ accepted jury submissions, but apparently this has happened in the past :joy: 

Another way to "work around" this problem in the new format, would be to explicitly set a `time_limit` (which was not possible in the legacy format). That will give errors when all AC submissions are slower than that, though.

Even when a system secretly supports `ac_to_time_limit < 1`, we probably don't want to have this documented, because it should™ not be used. So I guess the current `>= 1` can stay, then? :slightly_smiling_face:

PR #386, opened 2025-03-13
Markdown first draft

Matistjati:
A first draft of the Markdown specification.

Not allowing arbitrary HTML might be controversial, but securely implementing arbitrary HTML is a massive headache. The current view is basically that if you need more advanced control, use LaTeX instead.

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -418,7 +418,7 @@ so using this format should be avoided if at all possible.
 
 Auxiliary files needed by the problem statement files must all be in `statement/`.
 `problem.<language>.<filetype>` should reference auxiliary files as if the working directory is `statement/`.
-Image file formats supported are `.png`, `.jpg`, `.jpeg`, and `.pdf`.
+All statement types support the image file formats `.png`, `.jpg`, `.jpeg`.
 
 ### Sample Data
 
@@ -435,6 +435,8 @@ Image file formats supported are `.png`, `.jpg`, `.jpeg`, and `.pdf`.
 Problem statements provided in LaTeX must consist only of the problem statement body
 (i.e., the content that would be placed within a `document` environment).
 It is the judging system's responsibility to wrap this text in an appropriate LaTeX class.
+
+LaTeX statements support `.pdf` images in addition to the common image extensions.
 
 The LaTeX class shall provide the convenience environments `Input`, `Output`, and `Interaction` for delineating sections of the problem statement.
 It shall also provide the following commands:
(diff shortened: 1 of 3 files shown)


spec/2023-07-draft.md
```diff
@@ -463,6 +465,30 @@ However, judging systems must make a best effort to correctly render **at minimu
 - `\includegraphics` from the package `graphicx`, including the Polygon-style workaround for scaling the image using `\def \htmlPixelsInCm`.
 - The miscellaneous commands `\url`, `\href`, `\section`, `\subsection`, and `\epigraph`.
 
+### Markdown Environment and Supported Features
+
+Problem statements in Markdown must not include the problem name, as the judging system will automatically prepend it.
```
niemela, line 470: It seems that the current thinking is that we will require `\problemname` for LaTeX. If so, should we require a `{{problemname}}` for consistency?

spec/2023-07-draft.md
```diff
@@ -418,7 +418,7 @@ so using this format should be avoided if at all possible.
 
 Auxiliary files needed by the problem statement files must all be in `statement/`.
 `problem.<language>.<filetype>` should reference auxiliary files as if the working directory is `statement/`.
-Image file formats supported are `.png`, `.jpg`, `.jpeg`, and `.pdf`.
+All statement types support the image file formats `.png`, `.jpg`, `.jpeg`.
```
niemela, line 421: I would suggest listing all images supported by some format here in the same place.
```suggestion
All statement types support the image file formats `.png`, `.jpg`, `.jpeg`.
LaTeX statements also support `.pdf`.
Markdown statements also support `.svg`.
```

spec/2023-07-draft.md
```diff
@@ -436,6 +436,8 @@ Problem statements provided in LaTeX must consist only of the problem statement
 (i.e., the content that would be placed within a `document` environment).
 It is the judging system's responsibility to wrap this text in an appropriate LaTeX class.
 
+LaTeX statements support `.pdf` images in addition to the common image extensions.
```
niemela, line 439:
```suggestion
```

spec/2023-07-draft.md
```diff
@@ -463,6 +465,30 @@ However, judging systems must make a best effort to correctly render **at minimu
...
+
+The judging system shall provide the following commands:
+
+- {% raw %}`{{nextsample}}`{% endraw %} tells the judge system to include the next sample test case here.
+  It is an error to use {% raw %}`{{nextsample}}`{% endraw %} when there are no remaining sample test cases.
+- {% raw %}`{{remainingsamples}}`{% endraw %} tells the judge system to include all sample test cases that have not previously been included by {% raw %}`{{nextsample}}`{% endraw %}.
+  It is allowed to use {% raw %}`{{remainingsamples}}`{% endraw %} even if there are no remaining sample test cases, which will simply include nothing.
+- {% raw %}`{{name}}`{% endraw %} evaluates to the value of the corresponding constant, see [constants](#constants).
+
+The judging system shall support [CommonMark](https://commonmark.org/) features. However, full compliance with CommonMark is not required.
```
niemela, line 484: What does this mean?

niemela (approved):
I think this is a very good first draft. Should we merge it and iterate... or are there some obvious immediate improvements? (I suggested some changes).

spec/2023-07-draft.md
Matistjati: Full compliance with commonmark means passing a bunch of testcases. https://spec.commonmark.org/0.31.2/spec.json.
Many Markdown implementations don't pass all of these. It would be strange to require full compliance, since disabling HTML parsing probably makes the implementation uncompliant anyway.

spec/2023-07-draft.md
Matistjati: Added in c8dd8a74c373cdb76b78f14c647f21cb7022b105

spec/2023-07-draft.md
Matistjati: Added in c8dd8a74c373cdb76b78f14c647f21cb7022b105

spec/2023-07-draft.md
niemela: I agree (after you explanation) that we should not require full compliance. That said, it still feels very strange to say "shall support CommonMark ... full compliance ... is not required"

Matistjati:
Update: it seems that in practice, we're going to use an HTML/pdf sanitizer on the output anyway. Therefore, it might be nice to instead specify "unsafe HTML such as scripting is not allowed. Expect the output to be ran through a sanitizer". This would of course greatly improve the expressive power, but portability is decreased if some implementation uses a draconian HTML sanitizer. Any opinions on this?

spec/2023-07-draft.md
Matistjati: Summary of discussion with Fredrik: adding this does nothing except make it "more consisent". Personally, I believe that forcing something that does nothing isn't a good idea.

Matistjati:
@niemela I have made the changes as we discussed. Is this ready for a merge, or do we want more opinions?

niemela:
> Is this ready for a merge,

I think this is a very good first draft (better than just a draft). I find it easier to merge and then iterate, instead of having multiple open PRs.

Merging.

niemela (approved):

niemela (approved):

PR #388, opened 2025-03-14
Clarify usage of `\problemname`

### LINKED ISSUE #350: Clarify difference between `\problemname` and `\problemname{}`
If I understand LaTeX correctly, (and our spec), `\problemname` should mean "Insert problem name here", and `\problemname{}` should mean "set name to empty string (and insert it here)".

Do we want that? An empty name should be an error anyway? Would it be better is `\problemname{}` was specified to mean `\problemname` or to be an error?

niemela:
Closes #350 .

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -439,7 +439,10 @@ It is the judging system's responsibility to wrap this text in an appropriate La
 The LaTeX class shall provide the convenience environments `Input`, `Output`, and `Interaction` for delineating sections of the problem statement.
 It shall also provide the following commands:
 
-- `\problemname`, which must be the first line of the problem statement and places the `name` value matching the problem statement's language from `problem.yaml` into the problem statement header.
+- `\problemname{name}`, which should typically be the first line of the problem statement and places the problem name into the problem statement header.
+  The argument `name` is optional.
+  If it is missing, the `name` value from `problem.yaml` matching the problem statement's language is used.
+  If it is present, it is used instead, and must be a LaTeX-formatted version of the `name` value from `problem.yaml` matching the problem statement's language.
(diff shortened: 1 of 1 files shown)


spec/2023-07-draft.md
```diff
@@ -439,7 +439,10 @@ It is the judging system's responsibility to wrap this text in an appropriate La
 The LaTeX class shall provide the convenience environments `Input`, `Output`, and `Interaction` for delineating sections of the problem statement.
 It shall also provide the following commands:
 
-- `\problemname`, which must be the first line of the problem statement and places the `name` value matching the problem statement's language from `problem.yaml` into the problem statement header.
+- `\problemname{name}`, which should typically be the first line of the problem statement and places the problem name into the problem statement header.
+  The argument `name` is optional.
+  If it is missing, the `name` value from `problem.yaml` matching the problem statement's language is used.
+  If it is present, it is used instead, and must be a LaTeX-formatted version of the `name` value from `problem.yaml` matching the problem statement's language.
```
RagnarGrootKoerkamp, line 445: What would 'latex formatting' of the name in the yaml do?
I think all one can do is to simply copy that string as a black box into this.

niemela: The name in `problem.yaml` does not support math mode. The name in LaTeX clarly does. So, for example, `name` in `problem.yaml` could be `"H2O"`, while we use `\problemname{H$_2$O}` in LaTeX. A system that wish to be "LaTeX-aware" is free to pick up the LaTeX-version from here, but systems that does not suport LaTeX can use `name` in `problem.yaml` as is.

RagnarGrootKoerkamp: Oh I misread as the yaml string being formatted.
Yes all good then.

PR #396, opened 2025-03-17
Sync requirements for name key and statment filename

### LINKED ISSUE #389: Somewhat contradictory requirements for `name` field in `problem.yaml`
Currently we say:

> If only a single problem statement exists, this may be a string with the name of the problem in that language (but a map with a single key is allowed).
> 
> So the following example implies that only an English language problem statement exists:
> 
> ```yaml
> name: Hello World!
> ```

I would argue that the definition says that example given should imply that only a single language problem statement exists, but that that could be in any language.

I prefer the semantics given by the example (i.e. the language must be given unless there is only one language **and** that language is English) over what the definition says. 

Thoughts?

### LINKED ISSUE #390: Required and allowed languages for `name` in `problem.yaml`
Currently we require that the set of languages provided in `problem.yaml` **exactly** match the set of languages for which some statement is given.

It makes perfect sense to strictly require that any language used of a statement **must** be given in `problem.yaml`, but why should we require that additional languages must **not** be given?

The only reason I can think of is that you could use the `name` key to figure out what languages a problem is available in. My counterargument to that is that it's just as east and natural to check for the presence of `statement/problem.<lang>.*`. Also, since a given system might not support all of LaTeX, Markdown, and PDF, that could give a more relevant result.

A reason to not add it is that a workaround is trivial, just add an empty "translation" for the language you want to provide a title for, but that breaks both ways of figuring out what languages a problem is available in.

If we think this is something one might reasonably want to do, I think we should allow `name` in `problem.yaml` to be a superset of the languages for which some statement is given. If we don't, we don't have to do anything.

Thoughts?

niemela:
Closes #390: by doing nothing, which was the consensus.
Closes #389: by making language code *always* required for statement files, but defaulting to English in `problem.yaml`.

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -187,7 +187,7 @@ Value           | Incompatible with           | Comments
 
 The name of the problem in each language for which a problem statement exists.
 The `name` field is a map with the language codes as keys and the problem names as values.
-If there is only one langauge **and** that language is English, the `name` field can simplt be the problem name instead.
+If there is only one langauge **and** that language is English, the `name` field can simply be the problem name instead.
 The set of languages for which `name` is given must **exactly** match the set of languages for which a [problem statement](#problem-statements) exists.
 
 A deliberately complex and construed example:
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -199,7 +199,7 @@ name:
   fil: Kumusta mundo!
 ```
 
-The simplest example, which imples that the only provided language is `en`;
+The simplest example, which implies that the only provided language is `en`;
 ```yaml
 name: Hello World!
 ```
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -186,7 +186,7 @@ Value           | Incompatible with           | Comments
 ### Name
 
 The name of the problem in each language for which a problem statement exists.
-If there are statements in more than one language, the `name` field must be a map with the language codes as keys and the problem names as values.
+The `name` field must be a map with the language codes as keys and the problem names as values.
 The set of languages for which `name` is given must **exactly** match the set of languages for which a [problem statement](#problem-statements) exists.
 
 A deliberately complex and construed example:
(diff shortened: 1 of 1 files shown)


spec/2023-07-draft.md
```diff
@@ -186,7 +186,8 @@ Value           | Incompatible with           | Comments
 ### Name
 
 The name of the problem in each language for which a problem statement exists.
-The `name` field must be a map with the language codes as keys and the problem names as values.
+The `name` field is a map with the language codes as keys and the problem names as values.
+If there is only one langauge **and** that language is English, the `name` field can simplt be the problem name instead.
```
Tagl, line 190:
```suggestion
If there is only one langauge **and** that language is English, the `name` field can simply be the problem name instead.
```

niemela:
Updated based on @evouga's [comment](https://github.com/Kattis/problem-package-format/issues/389#issuecomment-2727871314).

spec/2023-07-draft.md
```diff
@@ -198,9 +199,7 @@ name:
   fil: Kumusta mundo!
 ```
 
-If only a single problem statement exists, this may be a string with the name of the problem in that language (but a map with a single key is allowed).
-
-So the following example implies that only an English language problem statement exists:
+The simplest example, which imples that the only provided language is `en`;
```
Tagl, line 202:
```suggestion
The simplest example, which implies that the only provided language is `en`;
```

spec/2023-07-draft.md
```diff
@@ -186,7 +186,8 @@ Value           | Incompatible with           | Comments
 ### Name
 
 The name of the problem in each language for which a problem statement exists.
-If there are statements in more than one language, the `name` field must be a map with the language codes as keys and the problem names as values.
+The `name` field is a map with the language codes as keys and the problem names as values.
+If there is only one langauge **and** that language is English, the `name` field can simply be the problem name instead.
```
Tagl, line 190:
```suggestion
If there is only one language **and** that language is English, the `name` field can simply be the problem name instead.
```

PR #397, opened 2025-03-20
Tweaks and typo fixes, and comments/questions on current text

eldering:


spec/2023-07-draft.md
```diff
@@ -532,14 +532,12 @@ possibly overwriting the compiled submission file or included data in the case o
 
 ### Illustrations
 
-One illustration file may be provided per test case.
+An illustration provides a visualization of the associated test case.
+At most one illustration file may be provided per test case.
 The file must share the base name of the associated test case.
 Illustration files are meant to be privileged information.
```
eldering, line 538: "privileged" here seems to imply "only available to the jury", but above for `test.files` it says "contains privileged data files available to the submission via file I/O.".

I think that should be made consistent and/or more clearly indicated what "privileged" exactly means.

On a second reading, I realize that both are *not* available to the team, so it makes sense. But that leaves me wondering (naively) whether `test.in` and other things are not privileged?

eldering (commented):
I was hoping to also be able to comment on text outside the diff of this PR, but I can't unfortunately...

spec/2023-07-draft.md
```diff
@@ -648,7 +646,7 @@ In particular, for an existing feedback directory `dir`,
 <output_validator_program> tc.in tc.ans dir [arguments] < tc.out # MUST PASS
```
eldering, line 646: A few lines above (and also in the section "Invalid Output") the text "for an existing feedback directory `dir`" looks broken to me as neither "feedback directory" nor `dir` seem to refer to anything.

I guess it should instead say something like:
```
In particular, for any test case in `valid_output/`, for example `valid_output/tc`,
```

spec/2023-07-draft.md
niemela: Good catch.

We currently use the word "privileged" in three places:
- `test.files`: "privileged data files available to the submission via file I/O"
- Illustrations: "meant to be privileged information"
- `judgeimage.<ext>`: "intent .. to be privileged information used as a debugging aid by contest judges"

These are clearly (?) two slightly different kinds of "privileged". 

The first is privileged in the same sense as any `/secret` data (or rather non-`full_feedback` data). I think we should simply not call this "privileged". It is only "privileged" if it is secret, so *not* when in `/sampe` of `full_feedback`:ed. Also, whatever we call this, we should use the exact same term for `.in` (and `args`).

The second two are privileged in the sense that they are supposed to be available for judges and not contestants/students. (I think it makes sense to just say "privileged" and not say exactly who should have access to it, because the access policy is IMO outside the scope of this spec. The best (and only) example I can think of is what to do with spectators. It feels completely up to the user whether to make privileged information available to spectators).

niemela: @eldering Do you want to fix this in this PR or make it into a new issue? Either sounds perfectly fine with me.

niemela (approved):

spec/2023-07-draft.md
eldering: > @eldering Do you want to fix this in this PR or make it into a new issue? Either sounds perfectly fine with me.

Sure, can do it in this PR. Now going to bed though, so will try to do it tomorrow.

spec/2023-07-draft.md
eldering: I've replaced these.

spec/2023-07-draft.md
eldering: I've replaced these "privileged" references by a central text. Would be good to get some reviews as I rewrote things a bit and added definition of "team" and "jury".

niemela: I think it would be nice if we could keep the terminology use case agnostic. So, not "team" and "jury" (and not "student", and "teacher"), but maybe "problem solver", and "problem setter"?

Am I just complicating things for very little benefit?

Also, I do know that we have already "hardcoded" e.g. `teammessage` and `judgemssage` into the spec, but that's much harder to fix.

Thoughts?

niemela: (I will merge this now, if we want to make the terminology more agnostic we can easily do that in a follow up)

PR #399, opened 2025-03-24
Add Webrick to gemfile

Matistjati:
There are no official instructions on how to run this document locally. It's nice to be able to run it locally to check formatting. Personally, i use `bundle install` followed by `bundle exec jekyll serve`. When I do this, it seems that the gem `webrick` is missing. 

Am I doing something wrong? If so, maybe we should add instructions on how to run it properly. Otherwise, it would be nice to add it to the gemfile.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,4 +20,5 @@ group :jekyll_plugins do
   gem "jekyll-remote-theme"
   gem 'faraday', '< 1'
   gem 'liquid', '>= 4.0.1'
+  gem 'webrick'
 end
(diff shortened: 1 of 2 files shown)


RagnarGrootKoerkamp:
Probably it's just a new dependency of some updated package somewhere. Sadly, this happens all the time with ruby/bundle :/

But yeah that's fine. (Before merging, maybe remove the markdown stuff from this PR.)

niemela (approved):

Issue #398, opened 2025-03-23
Clarify non-latex problem name


Matistjati:
In the legacy edition of the format, it implies that problemname should be rendered as LaTeX. However, most implementations need to display a non-LaTeXed version of the problem name. Kattis' solution is to use the undocumented comment `% plainproblemname <name>` (and if this is not present, the problem name is always rendered as plaintext). This is unnice since it differs from problemtools. My suggestion internally to "fix" this is to:
- always render statement as LaTeX
- if present, use `name` from problem.yaml for plaintext problem name
- otherwise, use the LaTeXed problemname verbatim

Is this something we want to add to the legacy specification? I.e., clarify that if you want a LaTeXed problemname, it would probably be wise to also give a non-latex problem name in the `name` in problem.yaml. Kattis doesn't mind changing the implementation, given that there are so few problems using LaTeX in their problem name.

PR #400, opened 2025-03-24
Specify where to get plaintext problem name

Matistjati:
Closes https://github.com/Kattis/problem-package-format/issues/398.

Should this also be added to `legacy-icpc`?

### DIFF
--- a/spec/legacy.md
+++ b/spec/legacy.md
@@ -240,7 +240,7 @@ Auxiliary files needed by the problem statement files must all be in `problem_st
 Image file formats supported are `.png`, `.jpg`, `.jpeg`, and `.pdf`.
 
 A LaTeX file may include the problem name using the LaTeX command `\problemname` in case LaTeX formatting of the title is wanted.
-If a plaintext version of the problem name is required, the `name` from value from `problem.yaml` shall be used if present.
+If a plaintext version of the problem name is required, the `name` value from `problem.yaml` shall be used if present.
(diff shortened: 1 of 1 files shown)


spec/legacy.md
```diff
@@ -239,7 +239,9 @@ Auxiliary files needed by the problem statement files must all be in `problem_st
 `problem.<language>.<filetype>` should reference auxiliary files as if the working directory is `problem_statement/`.
 Image file formats supported are `.png`, `.jpg`, `.jpeg`, and `.pdf`.
 
-A LaTeX file may include the Problem name using the LaTeX command `\problemname` in case LaTeX formatting of the title is wanted.
+A LaTeX file may include the problem name using the LaTeX command `\problemname` in case LaTeX formatting of the title is wanted.
+If a plaintext version of the problem name is required, the `name` value from `problem.yaml` shall be used if present.
+If not, the name given in `\problemname` is displayed verbatim.
```
niemela, line 244:
```suggestion
If not, the name given in `\problemname` is used verbatim.
```
"displayed" makes it (IMO) sound as if we are talking about "in the statement", but we're not.

niemela (approved):
Suggested one change.

@Matistjati What do you think?

spec/legacy.md
Matistjati: I agree that your wording is clearer.

PR #401, opened 2025-03-25
Remove "and construed" for multi-lingual name example

mpsijm:


gkreitz:
I think the current wording is correct.

mpsijm:
Ah, I hadn't heard the word "construe" before :stuck_out_tongue: It looks like "to construe" means something like "to interpret", but then I still don't think it's fitting :thinking: What about "A deliberately complex and illustrative example"? Then we also use simpler language (or at least, from my perspective, as I hadn't heard of "construe" before) :slightly_smiling_face:

gkreitz:
My understanding is that construed can also be used as a synonym for complex, but I can't find support for that in a dictionary now that I google. At least I guess I'm not completely hallucinating, as I guess whoever wrote the original had the same idea. :)

That said, why not keep things even simpler and just delete it? "A deliberately complex example:" feels clear enough to me.

gkreitz (approved):

PR #409, opened 2025-03-30
Fix languages table link

Matistjati:
Some old links to languages.md give 404, as it is now languages.html instead.


niemela:
> Some old links to languages.md give 404, as it is now languages.html instead.

Is that a jekyll cahnge, or what?...

niemela (approved):

Issue #408, opened 2025-03-30
How do we handle `score` > max score


niemela:
Broken out from #402

> I don't see anything preventing an output validator from writing a score.txt which is larger than the bounded maximum score. It should be clarified if that's intentional or not. (I see some potential use cases for this, but also a risk of bugs being unnoticed - partially, but not fully, mitigated by "It is a judge error if the score of any group or subgroup exceeds its maximum score.")

See also comments [1](https://github.com/Kattis/problem-package-format/issues/402#issuecomment-2751569291) and [2](https://github.com/Kattis/problem-package-format/issues/402#issuecomment-2752013412)

niemela:
There seem to be three options:

1. Consider the max score as a guarantee, this means that if `score` > max is produced (for any group) this is a judge error.
2. Consider the max score as a command, this means that the actual score produced (for any group) is clamped to `[0, max]`.
3. Consider the max score as a suggestion, this means that the max score is just used for the UI to know what "100%" of the score is, but the score can be higher.

It's really hard to test/verify that the `score` can't ever be > max, (one extreme example it the ["Travelling Salesperson 2D" problem](https://open.kattis.com/problems/tsp) where it took ~13 years to get a submission with a score > max. (This was not technically the max as we're discussing here, rather an assumption in the validator, but same principle). It is OTOH, trivial to make sure that the score can't ever be > max, but just adding clamping.

Currently (I guess) we do option 3, I feel that that is broken, but it does have the advantage that it's both visible, and not *catastrophic*. That said, in some sense it's worse than a judge error, since the latter is clear that it should and will be fixed, whereas it could be a hard sell to remove points that have already been awarded (but should have been impossible) in a running contest.

I'm strongly leaning towards option 2.

evouga:
Muddying the waters a bit is that the maximum score could be implied rather than explicit. For example, consider:
- `secret`, with maximum score of 100;
- `secret/small`, containing 8 test cases and no explicit maximum score;
- `secret/big`, containing 2 test cases and no explicit maximum score;
- a custom output validator that scores every problem out of 10.

What do we want to have happen here?

evouga:
Option 2 would yield a problem score of 70 for an AC solution; if tooling warns about exceeding the `secret/small` test group maximum score, that would tell the user they need to go in to `secret/*/test_group.yaml` and explicitly set a maximum score.

niemela:
[...]

PR #417, opened 2025-04-02
Simplify result aggregation

### LINKED ISSUE #402: Ambiguities in Vedict/Score aggregation
I tried understanding the new Verdict/Score aggregation description, and it's ambiguous (IMHO) in a couple of places. Grouping everything into this single issue, as it's all somewhat related. (Oops, this turned out longer than I expected. Keeping it as one issue for now, hopefully most of this is trivial to fix).

There is no default value specified for `require_pass` (which is empty list, I guess?)

There is no default value specified for `score` except for the `secret` group. The `score` key is not inherited. If I read the spec correctly, the only legal options are either to set `score: unbounded` in `secret`, or to explicitly provide a score config in *every* subgroup. Is that correct? If so, perhaps that should be explicitly stated.

The spec then says "`A` subgroups and static validation test cases without a provided score". It's grammatically unclear how "without a provided score" binds, but per my question above, I cannot see a scenario where subgroups can legally be without a provided score. So, is `A` always 0 or 1, depending on if there is a static test with default score?

"It is a judge error if `S > M` for a group with bounded maximum score and `sum` aggregation." - How is this a *judge* error? To me, a judge error is something that happens while judging, but unless I'm missing something, this should be a validation error in the problem package?

I don't see anything preventing an output validator from writing a `score.txt` which is larger than the bounded maximum score. It should be clarified if that's intentional or not. (I see some potential use cases for this, but also a risk of bugs being unnoticed - partially, but not fully, mitigated by "It is a judge error if the score of any group or subgroup exceeds its maximum score.")

`require_pass` semantic definition says "A group may specify that it should only be scored if a submission is accepted for another *test case* or all test cases in another test data group and their dependencies." (my emphasis), but later "Each required group must be either sample or a subgroup of secret with pass-fail aggregation.". Is `require_pass` allowed to list a test case or only a test group?

"The path of a group, relative to the data/ folder, must come later lexicographically than the paths of all dependent test cases and groups." - this reads backwards to me. I think it's clearer to say "The path of a group, relative to the data/ folder, must come later lexicographically than the paths of all test cases and groups it depends on."

You have earlier said "If it has no subgroups or test cases, then its score is 0.". If `require_pass` lists such an empty subgroup, is that considered passed or failed?

The syntax of the strings in `require_pass` should be clarified with an example. E.g., when specifying a test group (which is a directory), do we use Unix slashes as a directory separator? Is a trailing slash allowed? Are any sort of wildcards allowed (hopefully not)? Plus more if test cases are allowed...

niemela:
- Renames "Verdict/Score Aggregation" section to "Result Aggregation".
- Score must be specified for all test data groups (#402)
- A validator can't use either `score.txt` or `score_multiplier.txt`, but not both (#402)
- A score > max score is an error (#408)

Closes #402 and #408

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -1307,7 +1307,6 @@ A custom output validator or static validator may produce a `score.txt` or `scor
   The score of the test case is this number _multiplied_ by the test case maximum score. 
 - for test cases with bounded maximum score, `score.txt`,  if produced, must contain a single single non-negative floating-point number.
   The score of the test case is that number. 
-  The number can be larger than the test case's inferred maximum score.
 - for test cases with bounded maximum score, if no `score_multiplier.txt` or `score.txt` is produced, the test case score is its maximum score.
 - for test cases with unbounded maximum score, `score.txt` must be produced and must contain a non-negative floating-point number.
   The score of the test case is that number.
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -635,7 +635,7 @@ The format of `test_group.yaml` is as follows:
 
 Key                     | Type                                | Default                                 | Inheritance       | Comments
 ----------------------- | ----------------------------------- | --------------------------------------- | ----------------- | --------
-`scoring`               | Map                                 | [See Verdict/Score Aggregation](#verdictscore-aggregation)  | Not inherited  | Description of how the results of the group test cases and subgroups should be aggregated. This key is only permitted for the `secret` group and its subgroups.
+`scoring`               | Map                                 | [See Result Aggregation](#result-aggregation)  | Not inherited  | Description of how the results of the group test cases and subgroups should be aggregated. This key is only permitted for the `secret` group and its subgroups.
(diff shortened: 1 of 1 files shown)


spec/2023-07-draft.md
```diff
@@ -1303,12 +1307,12 @@ A custom output validator or static validator may produce a `score.txt` or `scor
   The score of the test case is this number _multiplied_ by the test case maximum score. 
 - for test cases with bounded maximum score, `score.txt`,  if produced, must contain a single single non-negative floating-point number.
   The score of the test case is that number. 
-  The number can be larger than the test case's inferred maximum score.
```
Matistjati, line 1306: What happens if the number is larger? Judge error? Warning? Clamp + warning? I understand if you prefer to address this in a follow-up PR, but the expected behavior definitely has to be specified.

Matistjati (approved):
Overall, I like the direction of the PR.

spec/2023-07-draft.md
niemela: Producing a score above the maximum, at any point, is a Judge Error.

I agree that this should be clarified (it is not as clear as I felt it was when I wrote it 😄), but I suggest that we do that in followup clean-ups.

PR #427, opened 2025-04-06
Fix language table links

Matistjati:
All links I changed in this PR are currently broken.

This is starting to feel like whac-a-mole.  It might we worth it to run CI checking that all links work. Especially annoying is that some of them work when running locally using jekyll, but not on the website.


niemela:
Any idea what changed to break this?

niemela (approved):

Matistjati:
Nope. I doesn't feel like the answer is going to be interesting enough to make it worth it to get to the bottom of it.

PR #428, opened 2025-04-06
Fix Markdown reference

Matistjati:
Oops, it's not CommonMark.


Matistjati:
To add to this: I don't think there is a good specification of Markdown tables. In practice, what we want is a description of the format of a table that is most likely to work across implementations, which is effectively what Markdown Guide is describing. I think that's the least bad alternative other than simply linking to Pandoc's extended Markdown.

niemela:
> Oops, it's not CommonMark.

Well, CommonMark is also a thing. Why *don't* we us that?

Matistjati:
There are no tables or footnotes https://spec.commonmark.org/0.31.2/. I don't think we can live without them.

niemela:
> There are no tables or footnotes https://spec.commonmark.org/0.31.2/. I don't think we can live without them.

Right, sorry, I failed at reading. We *are* using CommonMark (that's what we say at line 499), with the listed *additions*, and they obviously are not in CommonMark, because then we wouldn't have to add them.

niemela (approved):

PR #429, opened 2025-04-06
`require_pass` only on `pass-fail` groups

### LINKED ISSUE #411: Clarify how `require_pass` works.
Broken out from #402.

> require_pass semantic definition says "A group may specify that it should only be scored if a submission is accepted for another test case or all test cases in another test data group and their dependencies." (my emphasis), but later "Each required group must be either sample or a subgroup of secret with pass-fail aggregation.". Is require_pass allowed to list a test case or only a test group?

> You have earlier said "If it has no subgroups or test cases, then its score is 0.". If require_pass lists such an empty subgroup, is that considered passed or failed?

> The syntax of the strings in require_pass should be clarified with an example. E.g., when specifying a test group (which is a directory), do we use Unix slashes as a directory separator? Is a trailing slash allowed? Are any sort of wildcards allowed (hopefully not)? Plus more if test cases are allowed...

See also comments [1](https://github.com/Kattis/problem-package-format/issues/402#issuecomment-2751605580) and [2](https://github.com/Kattis/problem-package-format/issues/402#issuecomment-2751641826)

niemela:
Closes #411

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -1335,11 +1335,11 @@ It is a judge error if the score of any group or subgroup exceeds its maximum sc
 #### Required Dependent Groups
 
 A group may specify that it depends on some other test data groups.
-Each required group must be either `sample` or a have `pass-fail` aggregation.
-The dependent group will only be run if the group being depended on gets its maximum score, that is receives an accepted verdict for all test cases in the group.
+Each required group must be either `sample` or have `pass-fail` aggregation.
+The dependent group will only be run if the group being depended on receives an accepted verdict for all test cases in the group.
 If the dependent group is not run, the group score is 0.
 
 The paths of these required groups, relative to the `data` folder, are listed under the `require_pass` key.
-A path consists of zero or more directory names followed by a directory or file name, with a `/` character separating consecutive names. Each name must conform to the [general requirements](#general-requirements) on directory and file names and the specified test data group or test case must exist.
+A path consists of zero or more directory names followed by a directory or file name, with a `/` character separating consecutive names. Each name must conform to the [general requirements](#general-requirements) on directory and file names and the specified test data group must exist.
 
 The path of a group, relative to the `data/` folder, must come later lexicographically than the paths of all test cases and groups it depends on.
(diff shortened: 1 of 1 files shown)


spec/2023-07-draft.md
```diff
@@ -1334,12 +1334,12 @@ It is a judge error if the score of any group or subgroup exceeds its maximum sc
 
 #### Required Dependent Groups
 
-A group may specify that it should only be scored if a submission is accepted for another test case or all test cases in another test data group and their dependencies.
-Otherwise, none of the group's test cases are judged and the group score is 0.
+A group may specify that it depends on some other test data groups.
+Each required group must be either `sample` or a have `pass-fail` aggregation.
```
Matistjati, line 1338: "...or a have..." -> "...or have..."

spec/2023-07-draft.md
```diff
@@ -1334,12 +1334,12 @@ It is a judge error if the score of any group or subgroup exceeds its maximum sc
...
+The dependent group will only be run if the group being depended on gets its maximum score, that is receives an accepted verdict for all test cases in the group.
```
Matistjati, line 1339: Maybe reformulate "gets its maximum score" to remove amibuity? Since we require `pass-fail` anyway, how about something like "...will only be run if every test case in the group being depended on receives an accepted verdict".

spec/2023-07-draft.md
```diff
@@ -1334,12 +1334,12 @@ It is a judge error if the score of any group or subgroup exceeds its maximum sc
...
+If the dependent group is not run, the group score is 0.
 
-The paths of these required test cases or groups, relative to the `data` folder, are listed under the `require_pass` key.
+The paths of these required groups, relative to the `data` folder, are listed under the `require_pass` key.
 A path consists of zero or more directory names followed by a directory or file name, with a `/` character separating consecutive names. Each name must conform to the [general requirements](#general-requirements) on directory and file names and the specified test data group or test case must exist.
```
Matistjati, line 1343: Remove test case in "...test data group or test case must exist."

Matistjati (approved):

spec/2023-07-draft.md
niemela: Yes, especially since we already say that in the "that is..."

PR #430, opened 2025-04-07
Clarify right owner for public domain

### LINKED ISSUE #426: specification, rights_holder, public domain
It is not stated that "public domain" forbids rights_holder to be set, however, problemtools does assert this, and it is also logically true.

Currently it is only stated that a license other than "public domain" and "unknown" requires a rights_holder, but not that "public domain" forbids one.

niemela:
Closes #426


Matistjati (approved):
lgtm

Issue #421, opened 2025-04-05
legacy format specification, accept_score, reject_score and range


Zazmuz:
Currently in testdata.yaml, accept_score and reject_score are both defined as type 'String' when they should be 'int' or possibly 'float'.

range is also type 'String' and is a little unclear. Is it that the range is specified as "start end", meaning two numbers space separated, if so, what kind of numbers, ints, floats, both? 

The current code treats range as two space separated python floats, so making the specification clearer on this would be a good idea, even if its usage is very rare...

niemela:
Yes, we have not been great at specifying types. That said, it's not really technically wrong, the value *is* a string (all values you could possibly specify in YAML are), it's just the subset of string that can be interpreted as a number.

Could you write a PR for this?

PR #431, opened 2025-04-07
types for accept_score, reject_score and range

Zazmuz:
Made specification clearer regarding types for these keys in testdata.yaml.


Zazmuz:
Issue regarding PR: https://github.com/Kattis/problem-package-format/issues/421

spec/legacy.md
```diff
@@ -319,9 +319,9 @@ Key                      | Type                                           | Defa
...
-`accept_score`           | String                                         | 1            | Default score for accepted input files. May only be specified for scoring problems.
-`reject_score`           | String                                         | 0            | Default score for rejected input files. May only be specified for scoring problems.
-`range`                  | String                                         | \-inf +inf   | Two numbers A and B ("inf", "-inf", "+inf" are allowed for plus/minus infinity) specifying the range of possible scores. May only be specified for scoring problems.
+`accept_score`           | Int                                            | 1            | Default score for accepted input files. May only be specified for scoring problems.
+`reject_score`           | Int                                            | 0            | Default score for rejected input files. May only be specified for scoring problems.
```
niemela, line 323: *Should* this be an integer and not a decimal? What does the current implementation of problemtools do? If integer, I suggest `Integer` instead of `Int`.

niemela (changes requested):
Also make the same changes to `legacy-icpc`.

spec/legacy.md
Zazmuz: The problemtools implementation does not actively parse the value but it is later assigned to score which is a float, I will switch to floating-point number that and update icpc legacy too

Zazmuz: I checked again and if I am not wront legacy-icpc does not have these fields since they do not have testdata.yaml.

niemela (approved):

PR #439, opened 2025-04-20
Rename input_visualizer to test_case_visualizer; add *_visualizer_args

### LINKED ISSUE #351: Example for output_visualizer is missing
I would expect an example for the output_visualizer in the `maximal` problem. This is currently missing. Can this be added before the draft becomes official?

mpsijm:
Closes #351.

I've tested the example visualizers in the "maximal" problem with https://github.com/RagnarGrootKoerkamp/BAPCtools/tree/draft-visualizer, with the following side-notes:
- The visualizer currently only runs with a proper `generators.yaml`, so I temporarily wrote one:
  <details>
  <summary>Temporary <code>generators.yaml</code></summary>

  ```yaml
  solution: /submissions/accepted/accepted.py
  version: 2025-04

  data:
    sample:
      testdata.yaml:
        test_case_visualizer_args: [sample]
        output_visualizer_args: [sample]
      data:
        "1":
          in: "41"

    secret:
      testdata.yaml:
        test_case_visualizer_args: [secret]
        output_visualizer_args: [secret]
      data:
        "1":
          in: "13"
        "2":
          in: "-100"
        "3":
          in: "99"
        "4":
          in: "0"
  ```

  </details>

- BAPCtools does not support the `include/` (yet?), so it should be invoked as `bt generate --no-solution --no-validators`
- BAPCtools still uses `testdata.yaml`, I renamed this manually to `test_group.yaml` after running `bt generate`

Nonetheless, the "maximal" problem should be spec-compliant (and at some point in the future, BAPCtools should run it without problems as well) :smile:


spec/2023-07-draft.md
```diff
@@ -1244,15 +1250,16 @@ Almost all test cases failed — are you even trying to solve the problem?
 A validator program is allowed to write any kind of debug information to its standard error pipe.
 This information may be displayed to the user upon invocation of the validator.
 
-## Input Visualizer
+## Test Case Visualizer
```
niemela, line 1253: I still don't think that "Test case" is a better term than "Input" here. Could you explain why it's better?

In *some* sense "input" and "test cases" are of course synonyms. The "test case" *is* the "input". In some other sense you could argue (and I guess this is the point?) that the "test case" is the "input" plus some additional metadata. Most importantly (?) the expected output (which clearly is not *part* of the "input", but OTOH, it's *clearly* a function of the input), but also configuration and illustrations.

I prefer the name "Input Visualizer"

spec/2023-07-draft.md
```diff
@@ -1244,15 +1250,16 @@ Almost all test cases failed — are you even trying to solve the problem?
...
 
 If a tool was used to automate creating test case illustration annotations,
-it is recommended to include the input visualizer source code in the directory `input_visualizer/` along with invocation instructions in a file such as `input_visualizer/README.txt`.
+it is recommended to include the test case visualizer source code in the directory `test_case_visualizer/` along with invocation instructions in a file such as `test_case_visualizer/README.txt`.
+Tooling may make use of the `test_case_visualizer_args` key in the [test data configuration](#test-data-configuration) and [test case configuration](#test-case-configuration).
```
niemela, line 1257: How can `test_case_visualizer_args` be used if "invocation instructions" are needed?

It would be great if we could specify invocation, as we do with output visualizers.

niemela (changes requested):

spec/2023-07-draft.md
RagnarGrootKoerkamp: The point is that the visualizer has access to both the input and the answer, so it could, say, draw both a graph (input) and a corresponding shortest path (answer).

niemela: I would argue that visualizing the expected output *is* still visualizing the input, because it is visualizing the expected output **of the input**. (And I don't think that claim is quite as stupid as it sounds).

spec/2023-07-draft.md
```diff
@@ -1247,13 +1253,22 @@ This information may be displayed to the user upon invocation of the validator.
...
+
+Here, `arguments` is the `input_validator_args` key in the [test data configuration](#test-data-configuration) and [test case configuration](#test-case-configuration).
+
+Note that the visualizer will not be invoked by the judging system, the invocation instructions are mainly provided to ensure consistency between problem development tools.
 
 ## Output Visualizer
 
 An output visualizer is an optional [program](#programs) that is run after every invocation of the output validator in order to generate images illustrating the submission output.
-A visualizer program must be an application (executable or interpreted) capable of being invoked with a command line call. It is invoked using the same arguments as the output validator.
+A visualizer program must be an application (executable or interpreted) capable of being invoked with a command line call. It is invoked using the same arguments as the output validator, except that `output_validator_args` is replaced with `output_visualizer_args`.
```
niemela, line 1271: Isn't an input visualizer also a "visualizer program"?

```suggestion
An output visualizer program must be an application (executable or interpreted) capable of being invoked with a command line call. It is invoked using the same arguments as the output validator, except that `output_validator_args` is replaced with `output_visualizer_args`.
```

niemela:
@mpsijm Could you rebase?

niemela (approved):

PR #455, opened 2025-08-23
Clearer \problemname and specify that `name` from `problem.yaml` is displayed verbatim

Matistjati:
The previous version of the `\problemname` section was very repetitive. I've tried to make it clearer.

I have also added that `name` from `problem.yaml` should always be provided without LaTeX formatting. Most judge systems will display the problem name in at least one place where it does not support LaTeX rendering, so I think that adding this meaning is very useful for both online judge developers and problem authors. 

This also implies that markdown statements cannot have statements with LaTeX formatting in them. I personally think that this is fine; our usage of markdown is intended to weaker and simpler than LaTeX.

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -455,12 +455,12 @@ It is the judging system's responsibility to wrap this text in an appropriate La
 The LaTeX class shall provide the convenience environments `Input`, `Output`, and `Interaction` for delineating sections of the problem statement.
 It shall also provide the following commands:
 
-- `\problemname{name}`, which should typically be the first line of the problem statement and places the problem name into the problem statement header.
+- `\problemname{name}`, which should typically be the first line of the problem statement and sets the problem name in the statement header.
   The argument `name` is optional.
   If it is missing, the `name` value from `problem.yaml` matching the problem statement's language is used.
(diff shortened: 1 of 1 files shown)


eldering (approved):
LGTM

spec/2023-07-draft.md
```diff
@@ -191,6 +191,7 @@ Value           | Incompatible with           | Comments
 ### Name
 
 The name of the problem in each language for which a problem statement exists.
+Each name must not contain any LaTeX formatting, as it will be used in places where judge system does not support LaTeX rendering.
```
niemela, line 194: Technically it can absolutely "contain LaTeX formatting", it just wont be interpreted as LaTeX formatting. I don't think ant reasonable person will (mis-)read this sentence like I just did though?

niemela (approved):

spec/2023-07-draft.md
eldering: Agreed, but maybe a reason to use "should"? As in, if it contains LaTeX formatting then it will just render badly.

niemela: Good point, "should" is actually much better here.

niemela:
```suggestion
Each name should not contain any LaTeX formatting, as it will be used in places where judge system does not support LaTeX rendering.
```

PR #458, opened 2025-08-26
Add style guide

### LINKED ISSUE #326: Define formatting/writing style points?
While getting down to the the small details, there's been plenty of fixes and sometimes overriding of previous changes in regards to things like table layouts. What about collecting some points of what we try to adhere to?

Some points to consider:

- Use `_` for italics and `**` for bold in markdown
- How do we format tables? For example, do we use a prefix `|`?
- Spelling issues, for example, do we want to consistently write `e.g.` and `i.e.` as `for example` and `that is`?
- ....

niemela:
Fixes #326

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -1,0 +1,128 @@
+# Problem package format style guide
+
+This document defines the formatting and writing style conventions for the Kattis problem package format specification documents.
+
+## Purpose
+
+This style guide ensures consistency across all specification documents and makes diffs more useful when reviewing changes.
+
+## Markdown formatting
+
+### Text emphasis
+
+- Use `_` for _italics_
+- Use `**` for **bold** text
+
+### Tables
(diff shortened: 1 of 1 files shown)


STYLE.md
```diff
@@ -0,0 +1,128 @@
...
+- Use `**` for **bold** text
+
+### Tables
+
+Tables should be formatted with the following conventions:
+
+- No prefix `|` at the beginning of rows
+- No postfix `|` at the end of rows  
+- Exactly one space between `---` and `|` in the header divider row on each side
+- Match width of columns for all rows (that is, all columns are as wide as the largest content), except when that becomes unreasonable
```
eldering, line 23: I'd also argue that you may not want to reduce the width when removing the widest element in a column to prevent a diff of the whole table.

PR #466, opened 2025-08-26
Cleanup scoring

### LINKED ISSUE #461: Clarify default for `full_feedback`
In the "Test Data Configuration" section we say that `full_feedback` defaults to "`false` in `secret`, `true` in `sample`. We should clarify that it also defaults to `false` in any test group under `secret` (which is where all test groups must be, so alternatively we could say "`false` in `secret` or any test group").

niemela:
Flattens test_group.yaml by removing the `scoring` and `static_validation` maps.

Fixes #461

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -1012,7 +1012,7 @@ Each test group may define a static validation test case.
 It is an error to define static validation test cases without providing a static validator.
 A static validation test case is defined within a group's `test_group.yaml` file by specifying the key `static_validation_score`.
 
-If `static_validation_score` is specified as an non-negaive integer is it the maximum score of the static validation test case (see [Scoring Problems](#scoring-problems) for details).
+If `static_validation_score` is specified as a non-negative integer, then it is the maximum score of the static validation test case (see [Scoring Problems](#scoring-problems) for details).
 If it is specified as `pass-fail` then the test group it is part of must have `pass-fail` aggregation, or the problem be of type `pas-fail`.
 If `static_validator_args` is given it defines arguments passed to the static validator.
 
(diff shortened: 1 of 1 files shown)


spec/2023-07-draft.md
```diff
@@ -1007,17 +1010,16 @@ A static validator may be provided under the `static_validator` directory, simil
...
 It is an error to define static validation test cases without providing a static validator.
-A static validation test case is defined within a group's `test_group.yaml` file by specifying the key `static_validation`.
-If a map is specified, its allowed keys are:
-- `args`, which maps to a string which represents the additional arguments passed to the static validator in this group's static validation test case;
-- `score`, the maximum score of the static validation test case (see [Scoring Problems](#scoring-problems) for details).
+A static validation test case is defined within a group's `test_group.yaml` file by specifying the key `static_validation_score`.
 
-The `static_validation` key can also have the value of `false` meaning there is no static validation, or `true` meaning that static validation is enabled with no additional arguments and unspecified maximum score (to be determined by [maximum score inference](#maximum-score-inference)).
+If `static_validation_score` is specified as an non-negaive integer is it the maximum score of the static validation test case (see [Scoring Problems](#scoring-problems) for details).
```
Tagl, line 1015:
```suggestion
If `static_validation_score` is specified as a non-negative integer, then it is the maximum score of the static validation test case (see [Scoring Problems](#scoring-problems) for details).
```

spec/2023-07-draft.md
```diff
@@ -1007,17 +1010,16 @@ A static validator may be provided under the `static_validator` directory, simil
...
+If it is specified as `pass-fail` then the test group it is part of must have `pass-fail` aggregation, or the problem be of type `pas-fail`.
```
Tagl, line 1016:
```suggestion
If it is specified as `pass-fail`, then the test group it is part of must have `pass-fail` aggregation, or the problem must be of type `pass-fail`.
```

spec/2023-07-draft.md
```diff
@@ -1007,17 +1010,16 @@ A static validator may be provided under the `static_validator` directory, simil
...
+If `static_validator_args` is given it defines arguments passed to the static validator.
```
Tagl, line 1017:
```suggestion
If `static_validator_args` is given, then it defines arguments passed to the static validator.
```

spec/2023-07-draft.md
```diff
@@ -1007,17 +1010,16 @@ A static validator may be provided under the `static_validator` directory, simil
...
 
 It is an error to:
 - provide a static validator for `submit-answer` type problems,
-- specify a `score` in a test group with `pass-fail` aggregation or a problem that does not have the type `scoring`,
-- *not* specify a `score` in a test group that has `sum` or `min` aggregation.
+- specify a `static_validation_score` in a test group with `pass-fail` aggregation or a problem that does not have the type `scoring`,
+- specify a `static_validator_args` without specifying `static_validation_score`.
```
Tagl, line 1022:
```suggestion
- specify `static_validator_args` without specifying `static_validation_score`.
```

PR #479, opened 2025-08-31
Clarify score calculation

### LINKED ISSUE #475: Enforce `max_score`?
I seem to recall that we reached consensus that it would be a judge error to exceed `max_score`. 

One new potential annoyance I thought of if we do this: for groups with `sum` aggregation, every testcase would have to worth the same number of points (see [here](https://www.kattis.com/problem-package-format/spec/2023-07-draft.html#scoring-problems)). This would never have been a problem for any of the problems I have worked on, but my world-view might be slightly narrow.

niemela:
Fixes #475 

As noted in the discussion, most of the issue was a misreading. This minor change fixes the rest...


Matistjati (approved):
I still want to change a formulation to `max_score`. Will submit separate PR since I see no easy way to suggest it in Github

Issue #453, opened 2025-08-19
Consider making file regex more permissive


Matistjati:
After having used the new problemtools, which strictly enforces the regex, it seems to be stricter than what people use in practice. In particular, disallowing files starting with a . or underscore seems to be annoying. Some examples:
- Historically, the Swedish olympiad has renamed generators to `_donotuse_generator.sh` after the data was generated and final
- `.gitignore` files have been used by the Swedish Olympiad on a per-problem basis
- I seem to recall domjudge having some dotfile related to time limit, but I might be remembering wrong

These are the examples I am aware of, but surely there are more. I'm not aware of the rationale behind disallowing this. If there is no strong reason, maybe we should allow it?

niemela:
> So we (used to) support a file `.timelimit` in the problem package.

That was never actually in the spec though. And as the spec is written, they are actually forbidden.

niemela:
> After having used the new problemtools, which strictly enforces the regex, it seems to be stricter than what people use in practice. In particular, disallowing files starting with a . or underscore seems to be annoying. Some examples:
>
> [...]

None of your examples are actually "used" in the package.

The spec currently says `All file names for files included in the package must match the regexp <regexp>`. The strict way to interpret this would be that it's disallowed for any file within the package to not match the regexp, but it could *also* be interpreted as any file in the package that does not match the regexp are not "in the package", i.e. can safely be ignored.

All that said though... I do think the first strict reading makes more sense.

niemela:
> These are the examples I am aware of, but surely there are more. I'm not aware of the rationale behind disallowing this. If there is no strong reason, maybe we should allow it?

I agree with @eldering that there are very good reasons to not allow files starting with `.`. I don't see a good reason to not allow files starting with `_` though?

jsannemo:
I think e.g. .gitignore is a very strong reason to allow . files to exist?

Tagl:
I doubt any of them need to be so I would say no, but they must be allowed
to exist in subdirectories during development.

niemela:
In any case. I'm sensing that the consensus is "make the regexp more permissive". I also feel that the consensus is "still don't allow 'quotes, spaces, and strange stuff". It seems that the only real controversy is "dotfiles are bad because they are invisible" versus "dotfiles are useful because they area often used/needed".

After writing it out I'm actually leaning towards the "allow dotfiles" side.

Thoughts? Do you agree with what I claim we agree on? What do you think about the dotfile question?

[...]

PR #494, opened 2025-09-02
Allow underscores and periods at start of file name (but no period at start of test cases or programs)

evouga:
Fixes https://github.com/Kattis/problem-package-format/issues/453

(Based on my interpretation of the consensus there.)

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -25,10 +25,11 @@ unless it is indicated as meant to be shared with those attempting to solve the 
   Alternatively, the package can be a ZIP-compressed archive of such a directory with identical base name and extension `.kpp` or `.zip`.
 - All file names for files included in the package must match the regexp
   ```regex
-  ^[a-zA-Z0-9][a-zA-Z0-9_.-]{0,253}[a-zA-Z0-9]$
+  ^[a-zA-Z0-9_.][a-zA-Z0-9_.-]{0,253}[a-zA-Z0-9]$
   ```
   i.e., they must be of length at least 2, at most 255, consist solely of lower- or uppercase letters a–z, A–Z, digits 0–9, period, dash, or underscore,
-  but must not begin or end with a period, dash, or underscore.
+  but must not begin with a dash or end with a period, dash, or underscore.
+- Auxiliary files beginning with a period (such as `.gitignore`) are allowed in the problem package for convenience when interoperating with external tools. However, no test case or single-file [program](#program) may begin with a period.
 - All directory names inside the package must match the regexp
   ```regex
   ^[a-zA-Z0-9]([a-zA-Z0-9_-]{0,253}[a-zA-Z0-9])?$


spec/2023-07-draft.md
```diff
@@ -25,10 +25,11 @@ unless it is indicated as meant to be shared with those attempting to solve the
   Alternatively, the package can be a ZIP-compressed archive of such a directory with identical base name and extension `.kpp` or `.zip`.
 - All file names for files included in the package must match the regexp
   ```regex
-  ^[a-zA-Z0-9][a-zA-Z0-9_.-]{0,253}[a-zA-Z0-9]$
+  ^[a-zA-Z0-9_.][a-zA-Z0-9_.-]{0,253}[a-zA-Z0-9]$
   ```
   i.e., they must be of length at least 2, at most 255, consist solely of lower- or uppercase letters a–z, A–Z, digits 0–9, period, dash, or underscore,
-  but must not begin or end with a period, dash, or underscore.
+  but must not begin with a dash or end with a period, dash, or underscore.
```
niemela, line 31: Not beginning with dash is (I think) a very good idea, but why can't we end with dash, or underscore?

spec/2023-07-draft.md
```diff
@@ -25,10 +25,11 @@ unless it is indicated as meant to be shared with those attempting to solve the
...
+- Auxiliary files beginning with a period (such as `.gitignore`) are allowed in the problem package for convenience when interoperating with external tools. However, no test case or single-file [program](#programs) may begin with a period.
```
niemela, line 32:
```suggestion
- Auxiliary files beginning with a period (such as `.gitignore`) are allowed in the problem package for convenience when interoperating with external tools.
However, no test case or single-file [program](#programs) may begin with a period.
```

Are these all the "meaningful" files?

niemela (changes requested):

niemela:
> I like that, but it has the potential of being a slightly circular definition, since this implies that spec ignores all dotfiles, which then implies that dotfiles are automatically not relevant to the package.

How is that circular? And isn't that what we want? Take `.gitignore` for example, that is "not relevant for the package" itself, it's only relevant for tooling that works with the package. Isn't that what we wanted?

spec/2023-07-draft.md
evouga: That's the regex people wanted in the linked issue.

RagnarGrootKoerkamp:
> However, the behavior of the problem package must remain the same if all files starting with a period are removed.

This implies that the problem package ignores all dotfiles, which then implies this statement itself. Thus, we might as well state `any dotfiles in a problem package are ignored`, which is much more direct than the above.

But it seems we might want to special-case a few directories where we really do not wany *any* dotfiles, such as submissions or testcases, which contradicts the above statement.

(Well unless you argue that removing `data/secret/.dotfile` **does** change the meaning of the package, since then the package goes from broken to working, implying that `data/secret/.dotfile` is not allowed in the first place.)

Anyway, more direct wording is better.

-----

Side note: we used to have a bunch of `.gitkeep` files hanging around in `data/{sample,secret}` since we generate the content and we don't like git removing the directories. Would be nice to allow those?

spec/2023-07-draft.md
niemela: Yes I agree that that's what we said.

I just wondered while reviewing, is there a good reason we don't allow that?

niemela:
> What is "the behavior of the problem package"? What does it mean for it to "remain the same"?

I think (?) the wording I use was "meaning" or "semantics", rather than behavior, but that it similarly unclear.

The intended meaning of this is "removing them changes nothing for how the problem is run or evaluated".

> Note for instance that arguably it is possible to add or remove some test cases and submissions without changing "the behavior of the problem package" (the same set of user submissions provably get the same verdict with and without those test cases, etc.) Do we really want to allow dotfile test cases with this property?

We **absolutely** don't want that, but I don't think you can do that. Removing or adding a test case clearly (?) changes how a problem is judged.

evouga:
Pushed a revised attempt: now `._-` is allowed anywhere in file and folder names (except for leading dashes), but the problem package shall behave as if anything with a leading `.` were removed.

spec/2025-09.md
```diff
@@ -23,18 +23,13 @@ unless it is indicated as meant to be shared with those attempting to solve the
 - The package must consist of a single directory containing files as described below.
   The directory name must consist solely of lowercase letters a–z and digits 0–9.
   Alternatively, the package can be a ZIP-compressed archive of such a directory with identical base name and extension `.kpp` or `.zip`.
-- All file names for files included in the package must match the regexp
+- All file names and directories included in the package must match the regexp
```
eldering, line 26: Nitpick:
```suggestion
- All file and directory names included in the package must match the regexp
```

eldering:
> $ (shows up in Java .class filenames)

Nitpick counter-argument: this could potentially lead to issues when these filenames are processed in shell code that doesn't escape things properly and evaluates things as variable names. I guess this is not a strong enough argument though.

> ~ (emacs, old Windows)

Those are backup files, so should never affect the actual problem package, so I'd argue can be excluded with the same reasoning as a leading dot.

RagnarGrootKoerkamp:
I feel like I said this before but I don't see it in this thread here, but I still feel that we're conflating things:
1. the tree during development
2. the final package as uploaded to contest systems

With 1 I'm really fine with whatever, but with 2 it makes sense to be strict.
I don't think there's any need to upload files with `~` and `$`, and they should not be allowed in a final package, but that does not mean tools have to error on the development working directory.

RagnarGrootKoerkamp:
Do people typically just zip the directory and upload that?

With BAPCtools we pretty much always `bt zip` to get a `problem.zip`, and I think that already filters the top-level directory to only include known elements. (But does include junk inside eg `submissions/`.)
For testcases it's already quite precise and finds all `*.in` files and then from those finds corresponding `ans`/`out`/... files.


Does the fact that some metadata files out of our control have `~` and `$` in them imply that we'll accept testcases/submissions/programs that have those in their name? That seems silly.

spec/2025-09.md
```diff
@@ -23,18 +23,15 @@ unless it is indicated as meant to be shared with those attempting to solve the
 - The package must consist of a single directory containing files as described below.
   The directory name must consist solely of lowercase letters a–z and digits 0–9.
   Alternatively, the package can be a ZIP-compressed archive of such a directory with identical base name and extension `.kpp` or `.zip`.
-- All file names for files included in the package must match the regexp
+- All file and directory names included in the package must match the regexp
```
Matistjati: > - All file and directory names included in the package must match the regexp

We are contradicting ourselves slightly if "included" means "present in the folder structure" (which is a very reasonable interpretation IMO), since we later say that they are allowed to be present. How about

```suggestion
- All file and directory names included in the package (except for a few exceptions listed below) must match the regexp
```

Matistjati (approved):
I would be fine with merging this.

niemela (approved):

Issue #469, opened 2025-08-26
Could input validators support args and files


niemela:
This is basically a follow up on #393 and #416.

#416 fixed it for *output* validators, but we never actually did for *input* validators.

Matistjati:
W.r.t. to this, Fredrik asked me "what is the input validator version of what we're doing in the output validator? Please make a PR with that". But I simply don't think that that's possible.

What would feel better to me would be to implement what BAPCtools does for input validators (and the input validator has to simply figure it out by reading the files), and then generalize it to output validators. More specifically, fill the working directory with the input files.

...Annoyingly, we have that the working directory for output validators should contain "... the submitted files, any compiled binaries of the submitted files, as well as any included files". The most reasonable solution is that we copy the input files, and then potentially overwrite contestant files, as is specified in [input](https://www.kattis.com/problem-package-format/spec/2023-07-draft.html#input). It should be an error to specify testdata files that would overwrite validator files.

So to summarize my rambling, I suggest:
For both input, output and static validators and output visualizers, we copy `tc.in`, `tc.ans`, `tc.args`, and the contents of the `.files` directory, in a manner which could possibly overwite submitted files. 
What do you think @niemela ? If you like it, I could make a PR

niemela:
> Tangentially related: I could realistically see a problem with shortname `validator`, and then it has a static validator named `valitor.cpp`. Kattis names your submission `shortname.language`, so the static validator would overwrite any submission submission made in C++. So perhaps we should standardize it by creating subfolders `submissions/<submission_files>` and `data/<data_files>`.

That is a good point, but...

> Although this like optimizing for a case that "never happens".

Yes.

@gkreitz Thoughts?

niemela:
@gkreitz:

> AFAICT, you're talking about static validators

No, we were (mainly) talking about input validators.

> Can you explain the problem you're asking about a bit more clearly?

Input validators validate .in files, but for problems using files and/or args, those are also part of the input, and should be validated.

@Matistjati:

Similarly to this, we should make sure that files and args are supported in the invalid input/output and valid output sections.

Matistjati:
@gkreitz 
> So it would be explicitly against the spec to copy a validator.cpp into the working directory when its run, if I'm reading that correctly. Can you explain the problem you're asking about a bit more clearly?

I sidetracked a little. My point was to construct a not-unreasonable scenario where the current semantics are problematic.

PR #495, opened 2025-09-02
Extend input validators to nextpass.in, .files and args

### LINKED ISSUE #486: Can input validators be run on nextpass.in files
I see no reason why not (in fact, I think we *should* do that). But an implementer may not consider it if we don't explicitly specify that it should happen.

Matistjati:
Closes #486 and #469. I did both in the same PR, since splitting up would become messy.

It might be slightly controversial that an input validator failing to validate a `nextpass.in` is a judge error, but I really think that this is a good idea: if the output validator produces an invalid `nextpass.in`, that *is* a judge error.


jsannemo:
-- typically only means further arguments should only interpreted as positional arguments rather than options", no? At least I've never seen it used to distinguish between semantically different types of positional arguments

RagnarGrootKoerkamp:
Hmmm indeed, in the `cargo` example above it's options on the left and anything (options or positional) on the right.

I don't think that's an unreasonable interpretation here though: the input validator args feel more like options to the program, while the arguments passed to the submission are more of a black box (ie positional).

jsannemo:
Maybe, but I don't think we should think of it as a separator. Instead, the normal rationale of wanting all the tc-args being positional rather than interpreted as options is good enough (and the correct thing, IMO).

However, this runs into what I think is the 'real' issue: the decision to pass the default input validator flags as positional arguments instead of options. I don't know if there is a good reason for this. --case-insensitive is not harder to parse for arbitrary programs, but easier for any program using a stdlib/third-party argument parsing library. Maybe @niemela remembers why this was chosen?

The possibility of the non-option arguments to the default input validator and tc-args colliding is unfortunate, and means that either it is specified that the above only applies to a custom input validator, or that -- is indeed semantically this separator.

Is it too late to actually use proper options for the default validator args? :-)

Somewhat unrelated: the current phrasing around being able to run e.g. `validator [arguments] < input` seems wildly underspecified. The intent is that "the string [arguments] is passed to the program as separated arguments according to 'normal' shell parsing rules", which tbh is not a very convoluted alternative phrasing.

RagnarGrootKoerkamp:
> Maybe, but I don't think we should think of it as a separator. Instead, the normal rationale of wanting all the tc-args being positional rather than interpreted as options is good enough (and the correct thing, IMO).

The point of `--` is usually to say that everything after should be passed through untouched, regardless of whether they are options or positional arguments? Also here it really doesn't matter what the test case args look like right?

> However, this runs into what I think is the 'real' issue: the decision to pass the default input validator flags as positional arguments instead of options. I don't know if there is a good reason for this. --case-insensitive is not harder to parse for arbitrary programs, but easier for any program using a stdlib/third-party argument parsing library. Maybe @niemela remembers why this was chosen?

I'd say that is orthogonal, but I agree: it would be nice to have then as proper `--options` instead. Validator libraries can simply be updated to test for both `case-insensitive` and `--case-insensitive`.

> The possibility of the non-option arguments to the default input validator and tc-args colliding is unfortunate, and means that either it is specified that the above only applies to a custom input validator, or that -- is indeed semantically this separator.

But I still don't really see why it's bad to have `input_validator --flag arg1 arg2 --  --tc_flag tc_arg1 tc_arg2`

> Somewhat unrelated: the current phrasing around being able to run e.g. `validator [arguments] < input` seems wildly underspecified. The intent is that "the string [arguments] is passed to the program as separated arguments according to 'normal' shell parsing rules", which tbh is not a very convoluted alternative phrasing.

works for me

RagnarGrootKoerkamp:
Ah right; you want to specifically support use existing argument parsing frameworks for this. Yeah, then it's definitely nicer to follow the convention by default and have `--case-insensive` as an option by default.

Then users can still do custom non-option input validator arguments and pair it with a custom argument parser of course.

niemela:
> Is it too late to actually use proper options for the default validator args? :-)

As per my previous comment, that's not really related to *this* PR, since there is no default **input** validator.

spec/2023-07-draft.md
```diff
@@ -131,7 +131,7 @@ The language of files is inferred from their extension as listed in the [languag
 
 Each program must be run in a working directory with the following contents and **nothing else**:
 
-- For input validators: the files in the program directory of the input validator in question.
+- For input validators: the files in the program directory of the input validator in question, and the contents of the `.files` directory of the test case being validated (if this directory exists). It is an error for file names from `.files` and the input validator to collide.
```
niemela, line 134:
```suggestion
- For input validators: the files in the program directory of the input validator in question, and the contents of the `.files` directory of the test case being validated (if this directory exists). 
  It is an error for file names from `.files` and the input validator to collide.
```

spec/2023-07-draft.md
```diff
@@ -963,28 +963,27 @@ It is required that at least one submission is used to lower bound the time limi
 ## Input Validators
 
 Input Validators, verifying the correctness of the input files, are provided in `input_validators/`.
-Input validators can be specified as [VIVA](http://viva.vanb.org/)-files (with file ending `.viva`),
-[Checktestdata](https://github.com/DOMjudge/checktestdata)-file (with file ending `.ctd`),
-or as a program (as specified [above](#programs)).
+Input validators can be written in three formats: as a program (as specified [above](#programs)), a [VIVA](http://viva.vanb.org/)-file (with file ending `.viva`), or a [Checktestdata](https://github.com/DOMjudge/checktestdata)-file (with file ending `.ctd`).
```
niemela, line 966:
```suggestion
Input validators can be written in three formats: as a program (as specified [above](#programs)), 
a [VIVA](http://viva.vanb.org/)-file (with file ending `.viva`), 
or a [Checktestdata](https://github.com/DOMjudge/checktestdata)-file (with file ending `.ctd`).
```

spec/2023-07-draft.md
```diff
@@ -963,28 +963,27 @@ It is required that at least one submission is used to lower bound the time limi
...
 
-All input validators provided will be run on every input file.
-Validation fails if any validator fails.
+Every input validator will be run on every test case defined in `data`, as well as on any `nextpass.in` files
+that are dynamically generated during the execution of a `multi-pass` problem.
+Validation fails if any validator fails. It is a judge error for validation of a `nextpass.in` file to fail.
```
niemela, line 970:
```suggestion
Every input validator will be run on every test case defined in `data`, as well as on any `nextpass.in` files that are dynamically generated during the execution of a `multi-pass` problem.
Validation fails if any validator fails. It is a judge error for validation of a `nextpass.in` file to fail.
```

spec/2023-07-draft.md
```diff
@@ -963,28 +963,27 @@ It is required that at least one submission is used to lower bound the time limi
...
-When invoked, the input validator will get the input file on stdin.
-
 The validator should be possible to use as follows on the command line:
 ```sh
 <input_validator_program> [arguments] < inputfile
 ```
 
-Here, `arguments` is the `input_validator_args`.
+- `inputfile` is either a `.in` file defined in the `data` directory, or a `nextpass.in` file generated by an output validator.
+- `arguments` are the `input_validator_args` concatenated with the `args` of the testcase being validated. They must be chosen such that the input validator can distinguish which arguments belong to `input_validator_args`, and which belong to `args`.
```
niemela, line 982: Is this `input_validator_args` || `args` or `args` || `input_validator_args`? It's not super clear, but if anything I would interpret it as `input_validator_args` || `args`. We should make it clear.

I would prefer `args` || `input_validator_args`. Why is the opposite better?

niemela (changes requested):

spec/2023-07-draft.md
Matistjati: I have a slight preference for `input_validator_args` || `args`, because it feels more natural that the args intended for the input actual input validator come first.

How do we want to formally specify it? I suggested `arguments` = `input_validator_args` + ' ' + `args` earlier. What I dislike is that you lose a tiny bit of clarity that you get from natural language. IMO, || is too rare of an operator to use.

niemela: Yes, currying, you are absolutely right, I don't know what I was thinking?

RagnarGrootKoerkamp: Arguments is a list of strings from the programs point of view, not a string, so `+' '+` is weird. I guess that they *are* strings in the yaml files though... 

Please don't use || as a separator; that has other meanings in shell.

Matistjati: I changed it to "`arguments` are the `input_validator_args` followed by the `args` of the testcase being validated."

Is this acceptable?

niemela (approved):

PR #496, opened 2025-09-02
Specify output visualizers for multipass and interactive

### LINKED ISSUE #437: Output visualizer on interactive/multipass problem
What is the expected behavior when we run output visualizers on interactive or multipass problems? I can see two reasonable alternatives for interactive:
- Run them with an empty file as `.ans`. Write any info you need to the feedback directory
- For interactive, the ans file is the complete log of the contestant submissions' output during the interaction. I dislike this one, since it's a pain to implement. Maybe we could send the full communication log, but I would argue that that isn't the .ans file anymore.
- For multipass, the last .ans file produced
- Don't run them at all

I personally prefer option 1, especially given that there is no "obvious expected behavior" for multipass and interactive. I think there's a fair argument for option 3, but I'd much rather for it to blow up hard by feeding an empty `.ans` than the judge maybe, maybe not triggering a bug in their program.  Any other thoughts?

Matistjati:
Closes #437.

The rationale behind these choices are to keep it as simple as possible for the implementer: in many cases, it's infeasible for the judge to pass the required data without breaking backwards compatibility. So we might as well make it as pain-free as possible to increase adoption.

### DIFF
--- a/spec/2023-07-draft.md
+++ b/spec/2023-07-draft.md
@@ -1272,7 +1272,7 @@ Note that the visualizer would not typically be invoked by the judging system, t
 An output visualizer is an optional [program](#programs) that is run after every invocation of the output validator in order to generate images illustrating the submission output.
 An output visualizer program must be an application (executable or interpreted) capable of being invoked with a command line call. It is invoked using the same arguments as the output validator, except that `output_validator_args` is replaced with `output_visualizer_args`.
 It must be provided as a program (as specified [above](#programs)) in the directory `output_visualizer/`. 
-If the problem type is `interactive`, the command line argument `team_output` will be an empty file.
+If the problem type is `interactive`, the file `team_output` passed via stdin will be an empty file.
 If the problem type is `multi-pass` but not `interactive`, then `team_output` will contain the output of the submission from the last pass.
 Note that the `input_file` will not be adjusted for either `interactive` or `multi-pass` problems; it will be the `.in` file of the test case.
 


spec/2023-07-draft.md
```diff
@@ -1271,9 +1271,10 @@ Note that the visualizer would not typically be invoked by the judging system, t
 
 An output visualizer is an optional [program](#programs) that is run after every invocation of the output validator in order to generate images illustrating the submission output.
 An output visualizer program must be an application (executable or interpreted) capable of being invoked with a command line call. It is invoked using the same arguments as the output validator, except that `output_validator_args` is replaced with `output_visualizer_args`.
-It must be provided as a program (as specified [above](#programs)) in the directory `output_visualizer/`.
+It must be provided as a program (as specified [above](#programs)) in the directory `output_visualizer/`. If the problem type is `interactive`, the command line argument `team_output` will be an empty file. If the problem type is `multi-pass` but
```
niemela, line 1274:
```suggestion
It must be provided as a program (as specified [above](#programs)) in the directory `output_visualizer/`. 
If the problem type is `interactive`, the command line argument `team_output` will be an empty file. If the problem type is `multi-pass` but
```

spec/2023-07-draft.md
```diff
@@ -1271,9 +1271,10 @@ Note that the visualizer would not typically be invoked by the judging system, t
...
+not `interactive`, then `team_output` will contain the output of the submission from the last pass.
 
-All files written to the feedback directory by the output validator are accessible to the visualizer.
+All files written to the feedback directory by the output validator are accessible to the visualizer. If the output visualizer does not have sufficient data to generate an adequate visualization using only the command line arguments, it is recommended that the output validator writes the necessary data to the feedback directory.
```
niemela, line 1277:
```suggestion
All files written to the feedback directory by the output validator are accessible to the visualizer. 
If the output visualizer does not have sufficient data to generate an adequate visualization using only the command line arguments, it is recommended that the output validator writes the necessary data to the feedback directory.
```

niemela: Very nitpick: I feel that this should be stronger than "it is recommended". But, OTOH, it's clearly not a requirement from the spec, it's just.... that's what you'll have to to. Maybe it's fine?

niemela (approved):
Come minor style changes and a comment/question...

spec/2023-07-draft.md
Matistjati: I think I agree. My initial reasoning was that other solutions might exist, but honestly, I can't think of any reasonable ones.

niemela (approved):

spec/2023-07-draft.md
```diff
@@ -1271,9 +1271,12 @@ Note that the visualizer would not typically be invoked by the judging system, t
 
 An output visualizer is an optional [program](#programs) that is run after every invocation of the output validator in order to generate images illustrating the submission output.
 An output visualizer program must be an application (executable or interpreted) capable of being invoked with a command line call. It is invoked using the same arguments as the output validator, except that `output_validator_args` is replaced with `output_visualizer_args`.
-It must be provided as a program (as specified [above](#programs)) in the directory `output_visualizer/`.
+It must be provided as a program (as specified [above](#programs)) in the directory `output_visualizer/`. 
+If the problem type is `interactive`, the command line argument `team_output` will be an empty file. If the problem type is `multi-pass` but
+not `interactive`, then `team_output` will contain the output of the submission from the last pass.
```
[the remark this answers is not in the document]
Matistjati: Yes, that should definitely be clarified. I'm actually inclined to use the `.in` file from the first pass. Sending the last pass doesn't seem that much more valuable than the first, so we might as well reduce the amount of special cases (in this way, `input_file` will *always* be the `.in` file)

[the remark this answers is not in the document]
Matistjati: @niemela opinions? Thinking of adding
`The file `input_file` will *not* be adjusted for `multi-pass` problems, it will be the `.in` file of the test case.`

Matistjati:
@niemela merge pls 🥺

hairez (approved):

hairez (approved):

PR #504, opened 2025-09-04
Revert changes from #492 (naming restrictions for test data groups)

### LINKED ISSUE #480: Do we still need to disallow directories and test cases to have the same name?
In Test Data / Test Data Groups we say:
> A directory must not have the same name as a test case in the same directory.
For example, if the file `data/secret/group1/huge.in` exists then the directory `data/secret/group1/huge/` must not, and vice versa.

The was (IMO) definitely needed when we allowed multiple group levels and cases and groups on the same level, but now we don't so the only place where a directory could have the same name as a test case is when using directories to organize test cases inside a group. Thus the directory is not a "node" in the test data tree, and will not have a verdict or score.

I think we could remove this requirement.

(OTOH, I don't think it's ever *useful* to do, and it's probably not a great idea?)

hairez:
Revert changes from #492 regarding naming restrictions for test data groups and test cases.

Following the discussion in #480, we agreed to reinstate these restrictions.

This closes #480.

### DIFF
--- a/spec/2025-09.md
+++ b/spec/2025-09.md
@@ -599,6 +599,9 @@ For example, the test data group defined by the directory `data/secret/group1` h
 The test data groups themselves can contain directories, but not further groups.
 This means that there are no `test_group.yaml` further down in the directory hierarchy.
 
+A directory must not have the same name as a test case in the same directory.
+For example, if the file `data/secret/group1/huge.in` exists then the directory `data/secret/group1/huge/` must not, and vice versa.
+
 Each test data group must contain at least one test case, or a static validation test case.
 
 ### Input


spec/2025-09.md
```diff
@@ -599,6 +599,9 @@ For example, the test data group defined by the directory `data/secret/group1` h
 The test data groups themselves can contain directories, but not further groups.
 This means that there are no `test_group.yaml` further down in the directory hierarchy.
 
+A directory must not have the same name as a test case in the same directory.
+For example, if the file `data/secret/group1/huge.in` exists then the directory `data/secret/group1/huge/` must not, and vice versa.
```
Matistjati, line 603:
```suggestion
For example, if the file `data/secret/group1/huge.in` exists, then the directory `data/secret/group1/huge/` must not, and vice versa.
```

PR #512, opened 2025-09-06
Remove support for making any file downloadable by adding .download

### LINKED ISSUE #506: Remove the functionality to add `.download` to any kind of file and directory
Currently it says ```Additionally, any other file or directory with the extension `.download` is also available for download (without the `.download` extension).```

This should probably be changed, and made not available for directories? Feels difficult to support this feature. 

In some cases, `Attachments` would be used instead of `.download`. 

We should discuss what kind of files should be downloadable, if any.

niemela:
Fixes #506


hairez:
We should allow `.download` for test cases in `Secret/` as well. But that could maybe raised in a new issue.

hairez (approved):

niemela:
> We should allow `.download` for test cases in `Secret/` as well imo. But that could maybe raised in a new issue.

Yes, I agree, but I don't think that should be in this PR (or is part of the linked issue)

Issue #503, opened 2025-09-03
When is a testcase rejected?


mzuenni:
The spec currently specifies:
> Each `tc.in` under `invalid_input` must be rejected by at least one input validator.

There are two problems with this:
1. I think for *input validators* we do not specify what rejection means. It should be defined as "exits with exit code 43"
2. Is it ok for validators to crash if at least one rejected? If yes we may want to mention this.

I would argue in favor of 2. 

For example, consider a problem where the input is supposed to be a positive integer `x` with the extra property that `42 / x` is also an integer.
Now one checker can check that its a positive integer and the other can simply check that `42 & x == 0`.
However, the second checker will crash if `x` is zero. IMO this is fine since the other checker already ensures that `x` is not zero.

hairez:
I don’t fully understand the concern here.

Is it really a problem if the input validator “crashes” on invalid input? For example, many input validators use assert(), which technically causes a crash on failure. But I agree that this should be mentioned.

However, the new spec and legacy seem to cover this case:
```
The input validator must exit with code 42 on successful validation. Any other exit code means that the input file could not be confirmed as valid.
```

mzuenni:
> Is it really a problem if the input validator “crashes” on invalid input? For example, many input validators use assert(), which technically causes a crash on failure.

Yes, and I think this is bad. This makes it impossible to distinguish between the validator telling us that the file is invalid and it having a bug. For normal test cases this is not much of an issue because the validator should neither reject nor crash on those.
However, we introduced `invalid_outputs` to verify the behavior of the output checker. And IMO it would be nice to ensure here that the output validator does not crash but properly reject a test case (i.e. exits with 43).

niemela:
> Yes, and I think this is bad. This makes it impossible to distinguish between the validator telling us that the file is invalid and it having a bug. For normal test cases this is not much of an issue because the validator should neither reject nor crash on those.
> However, we introduced `invalid_outputs` to verify the behavior of the output checker. And IMO it would be nice to ensure here that the output validator does not crash but properly reject a test case (i.e. exits with 43).

Yes, exactly!

evouga:
So are we talking about input validators or output validators?

Imo:
- at least one *input* validator should cleanly exit with a non-42 error code on invalid input. It is ok for other input validators to crash;
- the *output* validator should cleanly exit with error code 43 on invalid output.

[...]

PR #514, opened 2025-09-07
Clarify how input validators report their results

evouga:
Attempt to fix https://github.com/Kattis/problem-package-format/issues/503

- Requires input validator programs to return 43 on validation failure
- Clearly distinguishes the input validator rejecting a test case from failing to operate properly 
- Clarifies that `invalid_input` must be cleanly rejected by at least one input validator (the others can do anything)

I've also moved the rationale for the strange exit codes to the Input Validator section (since it's earlier in the spec) and added a note that CTD and VIVA don't follow the exit code convention.

### DIFF
--- a/spec/2025-09.md
+++ b/spec/2025-09.md
@@ -670,7 +670,7 @@ Their goal is to ensure the integrity and quality of the test data and validatio
 
 The files under `invalid_input` are invalid inputs.
 Unlike in `sample` and `secret`, there are no `.ans` files.
-Each `tc.in` under `invalid_input` must be rejected by at least one input validator.
+Each `tc.in` under `invalid_input` must be rejected by at least one input validator. (Other input validators may reject the test case, accept it, or fail to operate properly.)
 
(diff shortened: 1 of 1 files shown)


spec/2025-09.md
```diff
@@ -985,12 +985,25 @@ first pass will be used.
...
 
-### Exit codes
+### Reporting Validation Results
 
-The input validator must exit with code 42 on successful validation.
-Any other exit code means that the input file could not be confirmed as valid.
+The input validator can accept a test case as valid, reject a test case, or fail to operate properly. Although validation fails in both of the latter cases, they should be clearly distinguished in any diagnostics reported to the user.
 
-#### Dependencies
+The Checktestdata and VIVA APIs define their mechanisms for reporting validation results. Input validator programs report a result via exit code.
```
```suggestion
Input validator programs report whether a test case is accepted or rejected via their exit code.
Exactly how VIVA and Checktestdata report results is outside the scope of this specification. 
```

PR #525, opened 2025-11-27
replace unavailable theme with JV-conseil/jekyll-theme-read-the-docs

godmar:


Matistjati:
This seems to switch to a CDN version that works. Maybe we should just merge this to get something that works, but IMHO, the real solution is to not rely on a CDN, but store the CSS file and other files we depend on in the repo. For reference, I'm pretty sure we use https://github.com/Kattis/jekyll-rtd-theme.

When I run this locally, it also switches the favicon to an IMO uglier one.
image

PR #533, opened 2025-12-07
Add more redirects to 2025-09

Matistjati:
Now that it's easy for us to add some redirects to the latest version, why not do it?


niemela (approved):

niemela:
It still *is* a draft though... we really should not have renamed if from `-draft` already. But that's my fault, and done is done....

Issue #523, opened 2025-11-24
Terminology: Test data groups


niemela:
First a very minor "issue", we use "test groups" and "test data groups" interchangeably. We should probably pick one.

Secondly as [mentioned elsewhere](https://github.com/Kattis/problem-package-format/issues/499#issuecomment-3556130481) we currently use the definition of "test group" to *exclude* `sample` and `secret`. We don't use this definition perfectly though, e.g. in the [group dependencies section](https://www.kattis.com/problem-package-format/spec/2025-09.html#group-dependencies) we currently say "A test data group, or `secret` may specify that it depends on some other test data groups.", but in fact one can also depend on `sample` so there should be a "... or `sample`" at the end of the sentence above. This example is not a huge issue of course, and is easy to fix, but it shows that we need to verify that we are using it correctly everywhere.

Also, as mentioned in the comment linked above, we could also choose the alternative definition where "test group" *includes* `sample` and `secret`. Which one we pick should IMO be heavily influenced by which choice makes the description easier to write (and read).

Thirdly, given the test groups can *only* exist as direct subdirectories of `data/secret` (pending of course a potential flip-flop on the semantics mentioned above), the current definition of the "name" of a group is more cumbersome than needed. We say "The name of a test data group is the path under `data/`. For example, the test data group defined by the directory `data/secret/group1` has the name `secret/group1`." This case could just have the name to be `group`, which would feel more natural. A side effect of this would be that a test group must not be called `sample` (and maybe, `secret`, but I'm not 100% sure that this is 100% needed... the fact that I don't really know might be an argument to include it...). Is this a problem?

Thoughts?

thorehusfeldt:
Thank you for taking this seriously. Some years ago, I tried to create consistent terminology and a glossary; you can find that (now outdated) appendix here: https://problem-package-format-copy.readthedocs.io/en/latest/format.html#glossary .

I would be super happy if we can find good names for this. Note also nodes like `invalid_inputs` (is that a test group) and its contents (are they test cases)?

niemela:
One implication of the current definition of "test data groups" is that the name for the configuration file `test_group.yaml` isn't so good after all, since it also configures non-"test groups". This is also somewhat indicated by the fact that this file is described in the section "Test data configuration", and it wouldn't quite work to rename the section to "Test data group configuration".

So, maybe we should rename the file back to `testdata.yaml` again?

thorehusfeldt:
[...]

Issue #499, opened 2025-09-03
Allowing `pass-fail` problems to have both test data groups and test cases in `secret/`


hairez:
To clarify, I would like to allow the following to be valid for `pass-fail` problems:
```
└── secret
    ├── dir1
    │   └──A.in
    └── B.in
```

Summarizing my understanding of discussion from #480 :

- The current formulation is `The secret directory must exist, and contain either some test cases, or some test data groups, but not both.`
- The PR #492 has allowed directories and test cases with the same name to coexist at any level but `secret`. 
- @RagnarGrootKoerkamp mentioned that it would be nice if this was allowed in `secret` as well for `pass-fail` problems.
- Some discussion regarding that there should not be any `test data groups` and only `test data directories` for `pass-fail` has been mentioned in #410.

- Argument against this is that one might want to give the submission feedback for a whole group of test cases (https://github.com/Kattis/problem-package-format/issues/410#issuecomment-2767485446). But I argue that submission feedback could be tied to a directory of test cases instead.

My suggestion is to allow test cases and directories to coexist in `secret/` for `pass-fail` problems.

 
Any other opinions, or points that I have missed?

Matistjati:
I don't really see the need to allow both test cases and directories? What do you lose by having either only testcases, or only test groups?

niemela:
> I was imagining that "groups" for `pass-fail` just gets treated as directories automatically, but maybe that is something to be discussed here as well.

That might solve @RagnarGrootKoerkamp's issue? Should we just say that there can't be any test data groups for `pass-fail` problems? I.e. any directories are just "organizational".

@Matistjati is there any (reasonable) reason we need a`test_data.yaml` and/or groups for `pass-fail`?

RagnarGrootKoerkamp:
one could still have `test_data.yaml` with different `output_validator_flags` for subsets of the data, right?

evouga:
Given that there is no good way to distinguish test groups from organizational directories, I am in favor of the status quo:
- subdirectories of `secret` are test groups; deeper subdirectories are organizational;
- `secret` can either have loose test cases, or test groups, but not both.

There are good reasons to forbid a mix of test groups and test cases in scoring problems (a mix significantly complicates aggregation), so it seems cleanest to me to consistently forbid a mix in pass-fail problems as well. The annoyance of putting loose test cases in an `other` directory is IMO quite minor.

niemela:
@Matistjati Am I missing something here?

[...]

PR #540, opened 2025-12-11
Clarify test data section

### LINKED ISSUE #505: Clarify compatibility and use cases for test data extensions
There are several supported file extensions: `.in`, `.ans`, `.out`, `.in.statement`, `.ans.statement`, `.interaction`, as well as `.download`, and maybe even `.yaml`.

We should make it clearer which test cases each of these extensions can be used for, and what they are compatible with.

There is already a table under the section for Test data, but I would like to extend it to include all supported file extensions.

Here are the file extensions that I have found and would like to be included:

Extension                       | Described In                                        | Summary                                        
------------------------------- | --------------------------------------------------- | ---------------------------------------------- 
`.in`                           | [Input](#input)                                     | Input piped to standard input                  
`.ans`                          | [Output Validator](#output-validator)               | Answer file given to the Output Validator      
`.out`                          | [Sample](#samples)                                  | Example team output, shown instead of `.ans`  
`.in.statement`                 | TODO                                                | Replaces the displayed `.in` file
`.ans.statement`                | TODO                                                | Replaces the displayed `.ans` file
`.interaction`                  | TODO                                                | Interaction shown for interactive problems
`.in.download`                  | TODO                                                | Replaces the downloadable `.in` file
`.ans.download`                 | TODO                                                | Replaces the downloadable `.ans` file
`.files`                        | [Input](#input)                                     | Input available via file I/O
`.yaml`                         | [Test Case Configuration](#test-case-configuration) | Additional configuration of the test case
`.png`, `.jpg`, `.jpeg`, `.svg` | [Illustrations](#illustrations)                     | Illustration of the test case

I also want to add another column called `Incompatible with` or something similar, since for example `.interaction` is not compatible with `.in.statement` or `.ans.statement`.  (see https://github.com/Kattis/problem-package-format/pull/516#discussion_r2545642340) 

Another column should maybe be `Allowed directories`, that clarifies where each file type could be used. For example, `.in.statement` should probably not be allowed in `Invalid Input` or `Invalid Output`. This is especially useful, since it is not obvious if `.statement` should be allowed in `Secret` if feedback for all test cases is allowed for a problem (mentioned to me by @niemela and @Matistjati).

### LINKED ISSUE #521: Terminology: Directory or folder
I think we are using the terms "directory" and "folder" interchangeably (which makes some sense, they mean the same thing), we should probably pick one and be consistent. 

We seem to use "directory" more often, so that choice seems reasonable to me?

Thoughts?

### LINKED ISSUE #522: Terminology: Test case or test data
We use both "test case" and "test data" somewhat interchangeably. (And arguably sometimes "input" is used to mean the same thing?).

This is not as clear cut, but maybe we should try to be more consistent?

Thoughts?

niemela:
General rewording and restructuring of test data section.

- Collect the "validation" test cases together.
- Move the test data group description earlier.
- Change all "test group" to "test data group" (Progress on #523)
- Change "folder" to "directory" (Closes #521)
- List all allowed files in test cases (Closes #505)
- Move description of statement and download overrides into test data section.
- Clarify that secret can have groups or test cases (and/or directories), but not both, and groups are defined by `test_group.yaml` (Progress on #499)
- Tried to use test case consistently (Closes #522)
   
Note that this ended up requiring a `.in.download` and a `.ans.download` file if you use a `.interaction` file. This is not what we discussed, and maybe not what we wanted, but it kinda logically follows from the rest.

### DIFF
--- a/spec/2025-09.md
+++ b/spec/2025-09.md
@@ -537,9 +537,9 @@ Directory              | Described In                      | Description
 ---------------------- | --------------------------------- | -----------
 `data/sample/`         | [Samples and full feedback](#samples-and-full-feedback) | Sample test data
 `data/secret/`         | [Test cases](#test-cases)         | Secret test data
-`data/invalid_input/`  | [Invalid input](#invalid-input)   | Inputs that must be rejected by input validator. 
-`data/invalid_output/` | [Invalid output](#invalid-output) | Outputs that must be rejected by output validator.
-`data/valid_output/`   | [Valid output](#valid-output)     | Outputs that must be accepted by output validator.
+`data/invalid_input/`  | [Invalid input](#invalid-input)   | Inputs that must be rejected by at least one input validator. 
+`data/invalid_output/` | [Invalid output](#invalid-output) | Outputs that must be rejected by the output validator.
+`data/valid_output/`   | [Valid output](#valid-output)     | Outputs that must be accepted by the output validator.
 
 The `secret` directory must exist, and contain either some test cases, or some [test data groups](#test-data-groups), but not both.
 All other directories are optional.


thorehusfeldt:
> Clarify that secret can have groups or directories, but not both,

Do you mean “test cases or test data groups, but not both”? (That was what #499) was about.)

spec/2025-09.md
```diff
@@ -531,43 +531,75 @@ Exactly how the solution description is used is up to the user or tooling.
...
+
+The secret data may be subdivided into test data groups.
+Every subdirectory of `data/secret/` that contains a `test_group.yaml` configuration file is a test data group.
+There can only be test data groups *or* test cases in `data/secret`, never both.
+That is, if there are any groups in `data/secret/` then there must not be any `.in` files directly in `data/secret/` nor any subdirectories that are not groups, and vice versa.
+The name of a test data group is the path under `data/`.
+For example, the test data group defined by the configuration file `data/secret/group1/test_group.yaml` has the name `secret/group1`.
+
+The test data groups themselves can contain directories, but not further groups.
+This means that there are no `test_group.yaml` further down in the directory hierarchy.
```
```suggestion
This means that there must not be any `test_group.yaml` further down in the directory hierarchy.
```

spec/2025-09.md
```diff
@@ -583,35 +615,19 @@ Key                       | Type                                | Default
...
+Note that since `.in` is required for all test cases, if there should be no input on standard input, an empty `.in` must still be provided.
+For a test case with a non-empty `.in`, it is piped to the submission as standard input.
 The submission will be run with the `args` sequence defined in the `test.yaml` file as command-line arguments.
 Note that usually the submission's entry point, whether it be a binary or an interpreted file, will be the absolute first command line argument.
 However, there exist languages, such as Java, where there is no initial command line argument representing the entry point.
 
-The directory `test.files`, if it exists, contains input files available to the submission via file I/O.
+The `.files` directory, if it exists, contains input files available to the submission via file I/O.
 All files in this directory must be copied into the submission's working directory after compiling, but before executing the submission,
 possibly overwriting the compiled submission file or included data in the case of name conflicts.
```
```suggestion
possibly overwriting the compiled submission file or [included](#included-files) files in the case of name conflicts.
```

spec/2025-09.md
```diff
@@ -531,43 +531,75 @@ Exactly how the solution description is used is up to the user or tooling.
...
-The sample data in `data/sample/` and the secret data in `data/secret/`.
+The test data is provided as separate [test cases](#test-cases) in subdirectories of `data/`:
+
+Directory              | Described In                      | Description
+---------------------- | --------------------------------- | -----------
+`data/sample/`         | [Samples and full feedback](#samples-and-full-feedback) | Sample test data
+`data/secret/`         | [Test cases](#test-cases)         | Secret test data
+`data/invalid_input/`  | [Invalid input](#invalid-input)   | Inputs that must be rejected by input validator. 
+`data/invalid_output/` | [Invalid output](#invalid-output) | Outputs that must be rejected by output validator.
+`data/valid_output/`   | [Valid output](#valid-output)     | Outputs that must be accepted by output validator.
```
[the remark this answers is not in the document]
niemela:
```suggestion
`data/invalid_input/`  | [Invalid input](#invalid-input)   | Inputs that must be rejected by at least one input validator. 
`data/invalid_output/` | [Invalid output](#invalid-output) | Outputs that must be rejected by the output validator.
`data/valid_output/`   | [Valid output](#valid-output)     | Outputs that must be accepted by the output validator.
```

niemela:
> Minor thoughts: maybe we should specify that require_pass, static_validation_score, score_aggregation and max_score are not inherited? Since this is a change from Legacy.

I feel that we explain this reasonably well. That said, I'm not against improving further :).

> Also, what are the semantics of `invalid_output` and `valid_output` for multipass problems? It makes sense for `invalid_output`: the output from the first pass must fail. But I don't see how a `valid_output` file could ever pass.

I think your observation is correct, but I don't think that's an issue. The semantics is the same for all problems, that just means that you probably will never use `valid_output` for multi-pass problems, and that's fine.

PR #559, opened 2025-12-16
Explain the limits specified in `problem.yaml`

### LINKED ISSUE #531: Should we specify some recommended limits for `output_visualizer`s?
The limits probably shouldn't be part of any config files, but providing some guidelines on what a "reasonable" limit on memory and execution time is would be nice. As a problem author, I have no idea whether the judges consider that to be 1 second 10 seconds. It's very feasible that that this actually matters, as it takes a while to generate larger images, *especially* if you're using Python.

### LINKED ISSUE #536: Is it a judge error for validators to exceed `validation_time`, `validation_memory` or `validation_output`?
"Obviously yes", but it currently isn't specified AFAICT. In addition, we should document that `validation_output` pertains to content written to stdout, not writing to files. This should probably be backported to legacy too AFAICT.

### LINKED ISSUE #538: Clarification on Counting Interaction Time Toward `validation_time`
For interactive and multi-pass problems, it should be clarified whether the time spent by user programs and validators during interactions is counted toward `validation_time` or `time_limit`.

This point is currently a bit unclear, and I couldn't find any explicit note about this in the specification. Although it seems somewhat obvious that user program time should count toward `time_limit`, while validator time should count toward `validation_time`. But it might still be worth clarifying?

The distinction would become more important if exceeding `validation_time` were considered a judge error (see #536).

### LINKED ISSUE #545: Explain the limits and their semantics
We could expand on the limits subsection in the problem metadata section. The only one that's explained at all is `time_*`, all the others are just assumed to be "obvious".

This is more general then, but would also solve #536.
I think it could also solve #538.

Specifically we should say (at least):
- explain what every one is about (even though it **is** fairly obvious)
- Say that `time_*`, `memory`, `output`, `code` need to be correct in both directions, i.e. the result will is incorrect if you allow submissions breaking the limits, as well as if you disallow submissions that don't. Failing these will cause a rejection (or possibly just be disallowed?)
- The `memory` one is a rabbit hole... we need to think what to say there that is helpful but not overly constraining on implementations.
- All the others are just lower bounds. I.e. `compilation_time` of X means that the system should wait *at least* X before giving up, but if it waits X+5, that's not *wrong* (it just takes longer).
- All the others cause Judge Error when failing.

niemela:
Explain all the `limits` in `problem.yaml`.

Closes #545 
Closes #536 
Closes #538 
Closes #531

### DIFF
--- a/spec/2025-09.md
+++ b/spec/2025-09.md
@@ -360,7 +360,8 @@ Systems are also permitted to enforce a reasonable wall-clock time limit to hand
 
 The `memory` limit specifies the maximum memory a submission may use.
 
-The `output` limit specifies the maximum total size of output a submission may produce, including standard output, standard error, and any files created (when file writing is allowed).
+The `output` limit specifies the maximum total size of output a submission may produce: 
+the sum of standard output, standard error, and (when file writing is allowed) the total size of all files that were created or modified by the submission.
 
 The `code` limit specifies the maximum total size of the submitted files of a submission.
 


spec/2025-09.md
```diff
@@ -337,27 +337,84 @@ The time of day defaults to the start of the day in UTC if not specified.
...
+`code`               | optional int > 0, in KiB                                 | system default | 128
+
+The `time_limit` specifies the maximum CPU time a submission may use per test case.
+Systems are also permitted to enforce a reasonable wall-clock time limit to handle submissions that consume little CPU time (e.g., by sleeping or blocking on I/O).
+
+The `memory` limit specifies the maximum memory a submission may use.
+
+The `output` limit specifies the maximum total size of output a submission may produce, including standard output, standard error, and any files created (when file writing is allowed).
+
+The `code` limit specifies the maximum total size of the source code of a submission.
```
```suggestion
The `code` limit specifies the maximum total size of the submitted files of a submission.
```

Matistjati (approved):
This is a very good addition to the spec!

spec/2025-09.md
```diff
@@ -337,27 +337,84 @@ The time of day defaults to the start of the day in UTC if not specified.
...
+`compilation_memory` | optional int > 0, in MiB       | system default | 2048
+`validation_time`    | optional int > 0, in seconds   | system default | 60
+`validation_memory`  | optional int > 0, in MiB       | system default | 2048
+`validation_output`  | optional int > 0, in MiB       | system default | 8
+
+The `compilation_time` and `compilation_memory` limits apply to compiling submissions.
+If a submission exceeds these limits during compilation, the verdict is Compile Error.
+
+The `validation_time`, `validation_memory`, and `validation_output` limits apply to running output validators.
+(Input validators are run at problem preparation time, not during judging, so these limits do not apply to them.)
```
Tagl: This is important, no parentheses please 😄 
```suggestion
Input validators are run at problem preparation time, not during judging, so these limits do not apply to them.
```

spec/2025-09.md
```diff
@@ -337,27 +337,84 @@ The time of day defaults to the start of the day in UTC if not specified.
...
+`memory`             | optional int > 0, in MiB                                 | system default | 2048
+`output`             | optional int > 0, in MiB                                 | system default | 8
+`code`               | optional int > 0, in KiB                                 | system default | 128
+
+The `time_limit` specifies the maximum CPU time a submission may use per test case.
+Systems are also permitted to enforce a reasonable wall-clock time limit to handle submissions that consume little CPU time (e.g., by sleeping or blocking on I/O).
+
+The `memory` limit specifies the maximum memory a submission may use.
+
+The `output` limit specifies the maximum total size of output a submission may produce, including standard output, standard error, and any files created (when file writing is allowed).
```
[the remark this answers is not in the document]
niemela: I fixed this by specifying that the entire size of any modified file is what counts. I think this makes the most sense.

Issue #587, opened 2026-07-15
Entrypoint specification problem for include files


gkreitz:
In the past (at least in problemtools and Kattis), the rule has basically been that something called `main.{suffix}` (typically case insensitive) is a driver, and that the entry point is derived from the file name in a language dependent way (e.g., `foo/Main.java` => `foo.Main`, and kotlin doing `Kt` things). In 2025-09, the rule seems to have changed, stating that:

> If any of the included files are supposed to be the main file (that is, a driver), that file must have the language-dependent name as given in the table referred to above.
- https://www.kattis.com/problem-package-format/spec/2025-09.html#included-files

In that table, most languages do not have a default entrypoint, meaning that it's impossible to add a driver in those languages. Furthermore, some of the entrypoints are not file names (e.g., `Main` and `MainKt`), so as the standard is currently written, it's also impossible to add a driver for those languages (or well, I guess you could argue that the tooling should somehow deal with a file called `MainKt`, but... no. :))

The legacy spec is even more broken here - it has the same wording copy-pasted, but the table no longer even exists in the spec (and digging back, the table that once existed did not match Kattis nor problemtools behavior).

I would like the rule to be that a file `[Mm][Aa][Ii][Nn].{suffix}` is the default entrypoint. If people care strongly about `__main__.py`, I think we should support that too (but I would prefer `main.py` to also work).

I would prefer for that entrypoint rule to also apply to submissions, even if the problem is smaller there as one can override `entrypoint` in the yaml for those (but that will get messy for a problem with many multi-file submissions)

Tagl:
I agree that the unofficial standard people follow is `main.py`, which I see much more of than the officially *correct* `__main__.py` for python.

PR #588, opened 2026-08-06
Fix removed entrypoint table in legacy

Matistjati:
"the table referred above." no longer exists. Instead link to appendix where it currently lives.


gkreitz:
This changes the semantics of legacy in a breaking way, requiring people to name python drivers as `__main__.py`, which is not consistent with legacy nor Kattis. We look for `main.{suffix}`, which I think should be codified in the standard (#587).

Matistjati:
Ok, fixing this seems more annoying than I initially thought.

gkreitz:
For the record, I think the correct fix is to apply this PR *and* fix #587 (i.e., make that table correspond to actual implementations).

